PR #13, opened 2020-02-21
Fix bugs

ChamseddineBhd:

**Other information**:

### DIFF
--- a/src/components/network/network.js
+++ b/src/components/network/network.js
@@ -37,8 +37,6 @@ export default class Network {
 
     lines = [];
 
-    voltageLevels = [];
-
     voltageLevelsByNominalVoltage = new Map();
 
     voltageLevelsById = new Map();
@@ -49,7 +47,6 @@ export default class Network {
 
     setSubstations(substations) {
         this.substations = substations;
-        console.log(substations)
 
         // add more infos
         substations.forEach(substation => {
@@ -57,8 +54,6 @@ export default class Network {
             substation.voltageLevels = substation.voltageLevels.sort((voltageLevel1, voltageLevel2) => voltageLevel1.nominalVoltage - voltageLevel2.nominalVoltage);
 
             substation.voltageLevels.forEach((voltageLevel, index) => {
-                 this.voltageLevels.push(voltageLevel)
-
                 // add substation id
                 voltageLevel.substationId = substation.id;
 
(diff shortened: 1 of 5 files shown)


src/components/network/network.js
```diff
@@ -37,6 +37,8 @@ export default class Network {
 
     lines = [];
 
+    voltageLevels = [];
```
geofjamg: It is redundant with voltageLevelsById

src/components/single-line-diagram.js
```diff
@@ -45,7 +45,7 @@ const SingleLineDiagram = (props) => {
             var bbox = svg.getBBox();
             svg.setAttribute("viewBox", 0 + " " + 0 + " " + (bbox.width + 20) + " " + (bbox.height + 20));
         }
-    }, []);
+    }, [props]);
```
geofjamg: You could be more precise by just adding properties (just property.svg?) that need to reapply the hack

src/redux/reducer.js
```diff
@@ -44,4 +47,6 @@ export const reducer = createReducer(initialState, {
     [REMOVE_VOLTAGE_LEVEL_DIAGRAM]: (state, action) => {
         state.diagram = null;
     }
-});
\ No newline at end of file
+});
+
+export const store = createStore(reducer);
```
geofjamg: Whyt have you move store creation here. This is strange to put store creation in reducer file.

src/components/single-line-diagram.js
ChamseddineBhd: ok

src/redux/reducer.js
ChamseddineBhd: ok i'll move it in a seprate file but it should be in the redux folder. what do you think ?

src/components/network/network.js
```diff
@@ -47,13 +49,16 @@ export default class Network {
 
     setSubstations(substations) {
         this.substations = substations;
+        console.log(substations)
```
geofjamg: this is for debugging => to remove

src/components/network/network.js
```diff
@@ -90,7 +95,7 @@ export default class Network {
     }
 
     getVoltageLevels() {
-        return Array.from(this.substations.values());
+        return Array.from(this.voltageLevels);
```
geofjamg: You can get this using voltageLevelsById values

src/redux/reducer.js
geofjamg: ok

PR #37, opened 2020-03-14
Add parameters component

geofjamg:
Signed-off-by: Geoffroy Jamgotchian <geoffroy.jamgotchian@gmail.com>

No

Feature

Parameters (theme and use name) are layed in the top bar.

Parameters have their own screen.

**Other information**:

### DIFF
--- a/src/components/parameters.js
+++ b/src/components/parameters.js
@@ -1,0 +1,104 @@
+/**
+ * Copyright (c) 2020, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import React from 'react';
+
+import {FormattedMessage} from "react-intl";
+
+import {useDispatch, useSelector} from "react-redux";
+
+import Box from "@material-ui/core/Box";
+import Container from "@material-ui/core/Container";
+import Divider from "@material-ui/core/Divider";
+import FormControlLabel from "@material-ui/core/FormControlLabel";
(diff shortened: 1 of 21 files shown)


src/components/parameters.js
```diff
@@ -0,0 +1,120 @@
...
+                    </Typography>
+                </Grid>
+                <Grid item xs={9}>
+                    <Switch
+                        checked={useName}
+                        onChange={handleToggleUseName}
+                        value={useName}
+                        color="primary"
+                        inputProps={{ 'aria-label': 'primary checkbox' }}
+                    />
```
mikatheo: Wouldn't it be better to give more space to the labels? Currently we have the useName label on 2 lines.

src/components/top-bar.js
```diff
@@ -40,28 +32,16 @@ const useStyles = makeStyles(theme => ({
...
-        dispatch(selectDarkTheme(!dark));
-    }
-
-    const handleToggleUseName = () => {
-        dispatch(toggleUseNameState());
+    const onParametersClick = () => {
+      if (props.onParametersClick) {
+          props.onParametersClick();
+      }
     };
```
mikatheo: When you click the button several times, it opens several windows on top of each other. Would it be possible to close on the second click the parameter window instead? (so same effect as the close button)

geofjamg: Thanks for the catch, it is a bug.
Close the parameters pane or disable the button in the top bar ?

geofjamg: Ok for closing the parameters pane.

src/components/parameters.js
geofjamg: Yes, I changed to code to equally split the table between the 2 columns.

PR #39, opened 2020-03-17
keep_the_same_zoom_level_for_substation_with_an_activable_scroll

FranckLecuyer:
Setting the svg height and width attribute from the svg bounding box, so this will dynamically adjust the size of the embedding window.
Defining a max width and max height for this window.
Adding scrollbars when necessary to scroll the svg inside the window.

feature

A substation is always drawn in a fixed size window and the svg is scaled to fit into this window

A substation is now drawn in a dynamically calculated size window (with a maximum fixed size) and the svg is not scaled anymore, so the zoom level is always kept.
Scrollbars are added when necessary to view all svg parts.

**Other information**:

### DIFF
--- a/src/components/single-line-diagram.js
+++ b/src/components/single-line-diagram.js
@@ -16,14 +16,20 @@ import {makeStyles} from "@material-ui/core/styles";
 import Typography from '@material-ui/core/Typography';
 
 const useStyles = makeStyles(theme => ({
+    div: {
+        overflowX: 'auto',
+        overflowY: 'auto'
+    },
     diagram: {
-        width: 800,
-        height: 600,
+        maxWidth: 800,
+        maxHeight: 700,
+        overflowX: 'auto',
+        overflowY: 'auto',
         "& .component-label": {
             fill: theme.palette.text.primary,
             "font-size": 12,
             "font-family": theme.typography.fontFamily
-        }
+        },
     },
     close: {
         padding: 0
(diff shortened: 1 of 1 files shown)


src/components/single-line-diagram.js
```diff
@@ -16,14 +16,20 @@ import {makeStyles} from "@material-ui/core/styles";
...
 
 const useStyles = makeStyles(theme => ({
+    div: {
+        overflowX: 'auto',
+        overflowY: 'auto'
+    },
     diagram: {
-        width: 800,
-        height: 600,
+        maxWidth: 800,
```
geofjamg: Ideally to be responsive it should depend on screen size. But probably we should do it in another issue.

src/components/single-line-diagram.js
```diff
@@ -33,13 +39,16 @@ const useStyles = makeStyles(theme => ({
 const SingleLineDiagram = (props) => {
 
     useEffect(() => {
-        var svg = document.getElementById("sld-svg").getElementsByTagName("svg")[0];
+        let svg = document.getElementById("sld-svg").getElementsByTagName("svg")[0];
```
geofjamg: could be a const

src/components/single-line-diagram.js
```diff
@@ -33,13 +39,16 @@ const useStyles = makeStyles(theme => ({
...
         if (svg) {
-            svg.style.height = "100%";
-            svg.style.width = "100%";
-            var bbox = svg.getBBox();
-            svg.setAttribute("viewBox", 0 + " " + 0 + " " + (bbox.width + 20) + " " + (bbox.height + 20));
-        }
+            let bbox = svg.getBBox();
```
geofjamg: Could be const

PR #54, opened 2020-03-25
Authentication

ChamseddineBhd:

**Other information**:


geofjamg:
Don't forget license headers!

public/signin-callback.html
```diff
@@ -0,0 +1,30 @@
...
+</head>
+
+<body>
+    <noscript>
+        You need to enable JavaScript to run this app.
+    </noscript>
+
+    <h1>Authentification callback processing...</h1>
+
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/oidc-client/1.8.2/oidc-client.js"></script>
```
jonenst: Don't use a cdn, we are hosting this ourselves

public/silent-renew.html
```diff
@@ -0,0 +1,7 @@
+<script src="https://cdnjs.cloudflare.com/ajax/libs/oidc-client/1.5.1/oidc-client.js"></script>
```
jonenst: Don't use a cdn, we are hosting this ourselves

src/components/top-bar.js
```diff
@@ -32,10 +41,52 @@ const useStyles = makeStyles(theme => ({
...
+            }}
+            {...props}
+        />
+));
+
+const StyledMenuItem = withStyles(theme => ({
+    root: {
+        '&:focus': {
+            backgroundColor: theme.palette.primary.main,
+            '& .MuiListItemIcon-root, & .MuiListItemText-primary': {
```
jonenst: not sure if we are supposed to use the Mui* classes ? Shouldn't we use our own classes ?

src/components/top-bar.js
```diff
@@ -32,10 +41,52 @@ const useStyles = makeStyles(theme => ({
...
+
 const TopBar = (props) => {
 
     const classes = useStyles();
 
+    const [anchorEl, setAnchorEl] = React.useState(null);
+    const open = Boolean(anchorEl);
+
+    const handleClick = event => {
+        console.log(event.currentTarget);
```
jonenst: remove debug log ?

src/services/AuthService.js
```diff
@@ -0,0 +1,39 @@
+import { Log, UserManager } from 'oidc-client';
+
+export class AuthService {
+  userManager;
+
+  constructor() {
+    console.log("hello");
```
jonenst: remove debug log

src/services/AuthService.js
```diff
@@ -0,0 +1,39 @@
...
+      redirect_uri: `${process.env.REACT_APP_CLIENT_ROOT}signin-callback.html`,
+      silent_redirect_uri: `${process.env.REACT_APP_CLIENT_ROOT}silent-renew.html`,
+      // tslint:disable-next-line:object-literal-sort-keys
+      post_logout_redirect_uri: `${process.env.REACT_APP_CLIENT_ROOT}`,
+      response_mode : 'fragment',
+      response_type: 'id_token token',
+      scope: process.env.REACT_APP_SCOPE,
+    };
+    this.userManager = new UserManager(settings);
+    Log.logger = console;
```
jonenst: remove log statements

jonenst: Or maybe not this one, not sure

src/services/AuthService.js
```diff
@@ -0,0 +1,39 @@
+import { Log, UserManager } from 'oidc-client';
+
+export class AuthService {
```
jonenst: I don't think we need this class, it's just a wrapper around UserManager ?

public/signin-callback.html
```diff
@@ -0,0 +1,30 @@
+<!DOCTYPE html>
```
jonenst: Don't add another html file in public folder, do this in our app using react router

public/silent-renew.html
jonenst: Don't add another html file in public folder, do this in our app using react router

.env
```diff
@@ -1,2 +1,7 @@
 REACT_APP_API_STUDY_SERVER=api/study-server
 REACT_APP_API_CASE_SERVER=api/case-server
+
+REACT_APP_CLIENT_ID=#
```
jonenst: It would be nice if our build produced html and javascript that work anywhere. The specific parts should come from kubernetes configmaps (or equivalently docker mounts "-v"). This means we can't have these new values here. Please try to implement it like that

.env.development
```diff
@@ -1,2 +1,7 @@
 REACT_APP_API_STUDY_SERVER=http://localhost:5001
 REACT_APP_API_CASE_SERVER=http://localhost:5000
+
```
jonenst: We have to implement a mode for the app where the authentication is mocked and we can login without an identityProvider. And this no-identityprovider mode It should be the default in development. Please try to implement it

src/services/AuthService.js
ChamseddineBhd: Yes, it initialize the userManger settings.

src/authentication/UserManagerMock.js
```diff
@@ -0,0 +1,30 @@
...
+        window.location = "/sign-in-callback";
+        return Promise.resolve(null);
+    }
+
+    signinSilent() {
+        return Promise.resolve(null);
+    }
+
+    signoutRedirect() {
+        sessionStorage.setItem("user", null);
```
jonenst: I think it's better to use a more specific test key. "powsybl-study-mock-user" ?

src/authentication/UserManagerMock.js
```diff
@@ -0,0 +1,30 @@
...
+    constructor(settings) {
+        this.settings = settings;
+    }
+
+    getUser() {
+        return Promise.resolve(JSON.parse(sessionStorage.getItem("user")));
+    }
+
+    signinRedirect() {
+        window.location = "/sign-in-callback";
```
jonenst: This will not work when deployed, we don't host the app at '/'
You have to use the same code as in index.js
```
const base = document.querySelector('base');
const basename = base && base.href && new URL(base.href).pathname || process.env.PUBLIC_URL && new URL(process.env.PUBLIC_URL).pathname || '/';
```
please refactor index.js so that the computed basename is stored in a global variable and available anywhere in the application

src/authentication/UserManagerMock.js
```diff
@@ -0,0 +1,30 @@
...
+        return Promise.resolve(null);
+    }
+
+    signinSilent() {
+        return Promise.resolve(null);
+    }
+
+    signoutRedirect() {
+        sessionStorage.setItem("user", null);
+        window.location = "/";
```
jonenst: same comment for basename

src/components/app.js
```diff
@@ -41,10 +42,51 @@ const getMuiTheme = (theme) => {
     }
 };
 
-const App = () => {
+const SignInCallback = () => {
+    const history = useHistory();
+
+    function handleCallback() {
+        new UserManagerHelper().getUserManagerInstance()
```
jonenst: Can you simplify the design to avoid having to "new" an object here ?

src/components/app.js
```diff
@@ -41,10 +42,51 @@ const getMuiTheme = (theme) => {
...
+    return (
+        <h1>Authentication callback processing...</h1>
+    )
+};
+
+const SilentRenew = () => {
+    function handleCallback() {
+        console.debug("SilentRenew");
+        new UserManagerHelper().getUserManagerInstance().signinSilentCallback().then(function () {
+            window.location = "/";
```
jonenst: basename

src/components/app.js
```diff
@@ -41,10 +42,51 @@ const getMuiTheme = (theme) => {
...
+    useEffect(() => {
+        handleCallback();
+    }, []);
+
+    return (
+        <h1>Authentication callback processing...</h1>
+    )
+};
+
+const SilentRenew = () => {
```
jonenst: Can you explain the silentrenew flow ? What pages are open, what urls, when do they refresh, etc.
I'm not sure this correct (after all, it's a silent renew, and you are making a component to show "Silent renew in progress" ...
Maybe we can leave out silent renew for now ?

src/components/study-manager.js
```diff
@@ -248,17 +250,60 @@ const StudyCard = ({study, onClick}) => {
 };
 
 const StudyManager = ({onStudyClick}) => {
-
+    const  userManagerHelper = new UserManagerHelper();
     const dispatch = useDispatch();
 
+    function login() {
```
jonenst: I think this whole block of code belongs somewhere else ? The studyt-manager shouldn't handle login/logout/token renew etc

src/components/study-manager.js
```diff
@@ -248,17 +250,60 @@ const StudyCard = ({study, onClick}) => {
...
     useEffect(() => {
+        getUser();
         fetchStudies()
             .then(studies => {
                 dispatch(loadStudiesSuccess(studies));
             });
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, []);
 
+    useEffect(() => {
```
jonenst: When you move the login/logout code elsewhere, you don't need this and the logout prop

src/components/top-bar.js
```diff
@@ -58,10 +110,44 @@ const TopBar = (props) => {
...
+                        aria-haspopup="true"
+                        onClick={handleClick}
+                    >
+                        <MenuIcon/>
+                    </Button>
+
+                    <StyledMenu
+                        id="customized-menu"
+                        anchorEl={anchorEl}
+                        keepMounted
```
jonenst: I don't see keepMounted in the Menu documentation, is this correct ?

src/redux/store.js
```diff
@@ -7,4 +7,5 @@
 import {createStore} from 'redux'
 import {reducer} from './reducer'
 
-export const store = createStore(reducer);
\ No newline at end of file
+export const store = createStore(reducer,
+    window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__());
```
jonenst: what is this ?

src/components/app.js
geofjamg: This class UserManagerHelper is not a helper. In addition to the bad naming i think this design is not best way to do it in javascript.

src/components/top-bar.js
jonenst: Ah I see it's passed down to the popover and then to the modal. Ok good

src/authentication/AuthService.js
```diff
@@ -0,0 +1,31 @@
+import { Log, UserManager } from 'oidc-client';
+import {UserManagerMock} from "./UserManagerMock";
+
+export class AuthService {
+  userManager;
```
geofjamg: I would declare userManager as a global variable (so out of AuthService)
const userManager = ...
And AuthService would be just a function to init userManager with the right instance depending of the context.

src/components/study-manager.js
```diff
@@ -247,18 +249,48 @@ const StudyCard = ({study, onClick}) => {
...
 };
 
-const StudyManager = ({onStudyClick}) => {
-
+const StudyManager = ({loggedOut, onStudyClick}) => {
+    const  userManagerHelper = new AuthService();
     const dispatch = useDispatch();
 
+    function login() {
+        userManagerHelper.getUserManagerInstance().signinRedirect();
```
geofjamg: So you cloud write just userManager.signinRedirect();

src/authentication/UserManagerMock.js
jonenst: now that we have the base tag even in developpment, you can use window.location = "." it should work everywhere

src/components/top-bar.js
jonenst: This is documented in the following example: https://material-ui.com/components/menus/#customized-menus "

jonenst: So it's ok

src/utils/authentication/AuthService.js
```diff
@@ -0,0 +1,15 @@
+import {UserManager} from 'oidc-client';
```
geofjamg: Copyright?

src/utils/authentication/UserManagerMock.js
```diff
@@ -0,0 +1,53 @@
+export class UserManagerMock {
```
geofjamg: Copyright?

PR #55, opened 2020-03-25
center map on substation on user clicks

jonenst:
Signed-off-by: Jon Harper <jon.harper87@gmail.com>

No

Feature

When clicking on a voltage level, only the svg is update

the map is centered on the substation of the voltage level

NO

**Other information**:
We can choose what to do when we close the svg, either nothing or zoom back to the whole network. For simplicity, I zoomed back to the whole network.
* We can also choose to not center on a substation during page load if we want to.
* We can also choose to no center on a substation when clicking on the map.
* We can also choose to not center on a substation on svg clicks.
* We can also choose to center on substation not when clicking on a vl in the list, but on a small button next to it, displayed only when selected (or hovered) ?

### DIFF
--- a/src/components/network/network-explorer.js
+++ b/src/components/network/network-explorer.js
@@ -22,6 +22,9 @@ import TextField from '@material-ui/core/TextField';
 import {FixedSizeList as List} from 'react-window';
 
 import Network from "./network";
+import ListItemText from "@material-ui/core/ListItemText";
+import IconButton from "@material-ui/core/IconButton";
+import GpsFixedIcon from '@material-ui/icons/GpsFixed';
 import Divider from "@material-ui/core/Divider";
 import {AutoSizer} from "react-virtualized";
 
@@ -34,7 +37,7 @@ const useStyles = makeStyles(theme => ({
     },
 }));
 
-const NetworkExplorer = ({network, onVoltageLevelClick}) => {
+const NetworkExplorer = ({network, onVoltageLevelDisplayClick, onVoltageLevelFocusClick}) => {
 
     const intl = useIntl();
 
@@ -59,22 +62,33 @@ const NetworkExplorer = ({network, onVoltageLevelClick}) => {
 
     useEffect(() => {
         if (currentVoltageLevel !== null) {
-            onVoltageLevelClick(currentVoltageLevel.id, currentVoltageLevel.name);
+            onVoltageLevelDisplayClick(currentVoltageLevel.id, currentVoltageLevel.name);
+            onVoltageLevelFocusClick(currentVoltageLevel.id, currentVoltageLevel.name);
         }
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [useName]);
 
-    function onClickHandler(index) {
-        if (onVoltageLevelClick !== null) {
+    function onDisplayClickHandler(index) {
+        if (onVoltageLevelDisplayClick !== null) {
             const vl = filteredVoltageLevels[index];
-            onVoltageLevelClick(vl.id);
+            onVoltageLevelDisplayClick(vl.id);
             setCurrentVoltageLevel(vl);
         }
     }
 
+    function onFocusClickHandler(index) {
+        if (onVoltageLevelFocusClick !== null) {
+            const vl = filteredVoltageLevels[index];
+            onVoltageLevelFocusClick(vl.id);
+        }
+    }
+
     const Row = ({ index, style }) => (
-        <ListItem button style={style} key={index} onClick={() => onClickHandler(index)}>
-            {useName ? filteredVoltageLevels[index].name : filteredVoltageLevels[index].id}
+        <ListItem button style={style} key={index} >
+            <ListItemText primary={useName ? filteredVoltageLevels[index].name : filteredVoltageLevels[index].id} onClick={() => onDisplayClickHandler(index)} />
+            <IconButton onClick={() => onFocusClickHandler(index)}>
+                <GpsFixedIcon />
+            </IconButton>
         </ListItem>
     );
 
--- a/src/components/network/network-explorer.js
+++ b/src/components/network/network-explorer.js
@@ -10,7 +10,7 @@ import PropTypes from "prop-types";
 
 import {useSelector} from "react-redux";
 
-import {useIntl} from "react-intl";
+import {FormattedMessage, useIntl} from "react-intl";
 
 import Grid from "@material-ui/core/Grid";
 import InputAdornment from '@material-ui/core/InputAdornment';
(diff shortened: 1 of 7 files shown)


geofjamg:
I would not do anything when closing the voltage level diagram, just let the center on last opened voltage level.

geofjamg:
Also would it be feasible to flyby to the voltage level instead of going directly to it?

jonenst:
> I would not do anything when closing the voltage level diagram, just let the center on last opened voltage level.

yeah you're right, done.

> Also would it be feasible to flyby to the voltage level instead of going directly to it?

Yes, I tried it and it just worked™.  You just have to add "transitionDuration: 1000" to the viewstate.


What do we want on page refresh with a voltagelevelid in the url ? Center on the whole network instantly like when there is no voltage level ? Center on the substation instantly ? Center on the substation with fly-by (this is what the code currently does)?

geofjamg:
> > I would not do anything when closing the voltage level diagram, just let the center on last opened voltage level.
> 
> yeah you're right, done.
> 
> > Also would it be feasible to flyby to the voltage level instead of going directly to it?
> 
> Yes, I tried it and it just worked™. You just have to add "transitionDuration: 1000" to the viewstate.
> 
> What do we want on page refresh with a voltagelevelid in the url ? Center on the whole network instantly like when there is no voltage level ? Center on the substation instantly ? Center on the substation with fly-by (this is what the code currently does)?

I would center on the substation without flyby.

jonenst:
> > What do we want on page refresh with a voltagelevelid in the url ? 

> I would center on the substation without flyby.

Done

jonenst:
I also added the FlyTo interpolator, how do you like it ? When you are zoomed in, it flies out to about the starting zoom level before flying back to the previous zoom level on the new substation

jonenst:
I changed the duration to 2s, it seems to be a good middle ground value. A bit slow when zoomed out, but not too fast when zoomed in

jonenst:
If that's ok, we can merge @geofjamg ?

geofjamg:
> If that's ok, we can merge @geofjamg ?

I will test again but I am not sure it was a good idea and a good feature.
I recommand you to test with a large network, it is slow and it also slow down diagram opening I suspect.

geofjamg:
> > If that's ok, we can merge @geofjamg ?
> 
> I will test again but I am not sure it was a good idea and a good feature.
> I recommand you to test with a large network, it is slow and it also slow down diagram opening I suspect.

Try to quickly click on various voltage levels and you will see that diagram opening is totally blocked by network map flyby.

geofjamg:
Maybe we should not track substation move automatically but on demand

geofjamg:
Or maybe we could let it as it is and add a parameter to activate it or not (not by default) and see if people like it even if it slow down substation browsing.

geofjamg:
Why did you commit this on jon PR?

mikatheo:
> Why did you commit this on jon PR?

Unless I'm mistaken, we are on the Jon PR.

geofjamg:
Also I would make the target appears only on list item hover, that is a common to propose additional functionalities without cluttering the UI with lot of useless buttons.

mikatheo:
> Also I would make the target appears only on list item hover, that is a common to propose additional functionalities without cluttering the UI with lot of useless buttons.

Okay, but we have to keep in mind that we will have to make this responsive later. This feature won't be usable on Mobile at the moment.

geofjamg:
> > Also I would make the target appears only on list item hover, that is a common to propose additional functionalities without cluttering the UI with lot of useless buttons.
> 
> Okay, but we have to keep in mind that we will have to make this responsive later. This feature won't be usable on Mobile at the moment.

hmmm right I didn't this in mind anymore about responsiveness. What is the right way to do it on mobile?

mikatheo:
> > > Also I would make the target appears only on list item hover, that is a common to propose additional functionalities without cluttering the UI with lot of useless buttons.
> > 
> > 
> > Okay, but we have to keep in mind that we will have to make this responsive later. This feature won't be usable on Mobile at the moment.
> 
> hmmm right I didn't this in mind anymore about responsiveness. What is the right way to do it on mobile?

On closer inspection, an action (or a secondary actions button) for all items of a list seems to be Material-design compliant. This is the only way it will always be available whatever the device used.  
image
Perhaps it will seem more pleasing to the eye if the density of substation occurrences is reduced? There could also be later subheading additional information such as country and voltage level?

src/components/study-pane.js
```diff
@@ -79,7 +79,9 @@ const StudyPane = () => {
 
     const [studyNotFound, setStudyNotFound] = useState(false);
 
-    const [voltageLevelId, setVoltageLevelId] = useState(null);
+    const [displayedVoltageLevelId, setDisplayedVoltageLevelId] = useState(null);
+
+    const [focusedVoltageLevelId, setFocusedVoltageLevelId] = useState(null);
```
geofjamg, line 84: What is the purpose of this variable?

mikatheo: "voltageLeveld" was initially used to open the diagram from the clicked voltage level in the explorer. Now I have also to share an another ID for the map from an action in the explorer. That's why there are now two variables in the study-pane common context. Do you have a suggestion to share this information ?

mikatheo:
Note for myself : internationalize the 2 labels

src/components/study-pane.js
geofjamg: ok I understand

src/components/study-pane.js
```diff
@@ -184,15 +186,21 @@ const StudyPane = () => {
...
     } else {
-        let voltageLevel = null;
-        if (voltageLevelId && network) {
-            voltageLevel = network.getVoltageLevel(voltageLevelId);
+        let displayedVoltageLevel, focusedVoltageLevel = null;
+        if (network) {
+            if (displayedVoltageLevelId) {
+                displayedVoltageLevel = network.getVoltageLevel(displayedVoltageLevelId);
+            }
+            if (focusedVoltageLevelId) {
```
geofjamg, line 194: You don't need this variable in the renderer?

src/components/network/network-map.js
```diff
@@ -62,39 +64,64 @@ const NetworkMap = (props) => {
     // we get the ref to the deck and it has not yet initialized..)
     function onAfterRender() {
             //use centered and deck to execute this block only once when the data is ready and deckgl is initialized
-            if (!centered && deck !== null && deck.viewManager != null && props.geoData !== null) {
+            //TODO, replace the next lines with setProps( { initialViewState } ) when we upgrade to 8.1.0
+            //see https://github.com/uber/deck.gl/pull/4038
+            //This is a hack because it accesses the properties of deck directly but for now it works
+            if ((!centered.centered || (props.centeredSubstationId && props.centeredSubstationId != centered.lastCenteredSubstation))
```
geofjamg, line 70: This prop should be added to propTypes and defaultProps

src/components/network/network-explorer.js
```diff
@@ -59,22 +71,61 @@ const NetworkExplorer = ({network, onVoltageLevelClick}) => {
...
+        if (onVoltageLevelFocusClick !== null) {
+            const vl = filteredVoltageLevels[voltageLevelMenuIndex];
+            onVoltageLevelFocusClick(vl.id);
+        }
+        handleClose();
+    }
+
+    const voltagelevelInfo = index => {
+        if (network.getSubstation(filteredVoltageLevels[index].substationId)) {
+            return network.getSubstation(filteredVoltageLevels[index].substationId).countryName + " - " + filteredVoltageLevels[index].nominalVoltage + " kV";
```
jonenst, line 106: maybe put the voltage first as it is always short. If we ellipsize it will still be visible

Issue #56, opened 2020-03-25
The name has to be truncated to preserve the card width


mikatheo:
Signed-off-by: Mika Théodore michael.theodore@rte-france.com

* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
The card can become very large if the name is.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**
Cards should be kept the same width, and their design should be upgraded.

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

PR #57, opened 2020-03-25
In the study list page: Truncate the card names, and add description/…

mikatheo:
…menu buttons (#56)

Signed-off-by: Mika Théodore michael.theodore@rte-france.com

Feature

The card can be very large if the name is long.

The name is truncated, and there are 2 new buttons to display the description and the menu

**Other information**:

### DIFF
--- a/src/redux/reducer.js
+++ b/src/redux/reducer.js
@@ -53,16 +53,10 @@ export const reducer = createReducer(initialState, {
 
     [OPEN_STUDY]: (state, action) => {
         state.studyName = action.studyName;
-        state.description = action.description;
-        state.caseFormat = action.caseFormat;
-        state.caseDate = action.caseDate;
     },
 
     [CLOSE_STUDY]: (state) => {
         state.studyName = null;
-        state.description = null;
-        state.caseFormat = null;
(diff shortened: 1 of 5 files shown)


geofjamg:
I didn't look at the code but just ran it. Here is my feeling.
 - this is very nice and looks far better than previous version.
 - I would try to compact the card action part by reducing the top and bottom margin.
 - When the study name is truncated, we have no way to see the full name. I would:
   - add a tooltip on the card with the full name
   - add the name inside the collapse section just before the description, without truncation this time.
 - I would remove move and export in the menu until it is really implemented. I think it is a bad practice to expose non implemented features.
 - right click on the card (whatever the place), open the menu like with the button. Is it on purpose?

src/redux/reducer.js
```diff
@@ -53,10 +53,16 @@ export const reducer = createReducer(initialState, {
 
     [OPEN_STUDY]: (state, action) => {
         state.studyName = action.studyName;
+        state.description = action.description;
```
geofjamg: You don't need these fields when the study is open?

mikatheo: Done

PR #63, opened 2020-03-30
Improve rendering performance

geofjamg:
Signed-off-by: Geoffroy Jamgotchian <geoffroy.jamgotchian@rte-france.com>

No

Performance improvement

Opening of single line diagram is very slow for large grids (1 second). The reason is when a single line diagram is open, network map and nominal voltage filtering components are always re-rendered even if their data have not changed.

After analysis it appears that re-rendering is caused by callback props (which are arrow functions) that are always differents at each render because callback functions are recreated at each render of parent component.

A quick reminder: React by default performs to shallow equality of component props to decide if render has to be called again.

With hooks, the common way to implement "shouldComponentUpdate" is using React.memo that allows to implement our own comparator and then exclude callbacks from comparison.

NetworkMap and NominalVoltageFilter rendering memoization has been implemented.
Even on large network, single line diagrams are displayed almost instantaneously.

**Other information**:

### DIFF
--- a/src/components/network/nominal-voltage-filter.js
+++ b/src/components/network/nominal-voltage-filter.js
@@ -92,11 +92,4 @@ NominalVoltageFilter.propTypes = {
     onNominalVoltageFilter: PropTypes.func
 };
 
-function areEqual(prevProps, nextProps) {
-    // dont't check for onNominalVoltageFilter equality as the function will change at every render
-    // of the parent
-    return prevProps.nominalVoltages === nextProps.nominalVoltages
-        && prevProps.filteredNominalVoltages === nextProps.filteredNominalVoltages;
-}
-
-export default React.memo(NominalVoltageFilter, areEqual);
+export default React.memo(NominalVoltageFilter);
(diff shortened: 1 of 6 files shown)


src/components/network/nominal-voltage-filter.js
```diff
@@ -37,18 +35,14 @@ const useStyles = makeStyles(theme => ({
...
 
     useEffect(() => {
-        if (props.network) {
-            setNominalVoltages(Array.from(props.network.voltageLevelsByNominalVoltage.keys()).sort((a, b) => b - a));
-        }
-        if (props.filteredVoltageLevels) {
-            setCheckedNominalVoltages(props.filteredVoltageLevels);
-        }
-    }, [props.network, props.filteredVoltageLevels]);
+        setCheckedNominalVoltages(props.filteredNominalVoltages);
```
jonenst: I think we can remove this as well ?

geofjamg: > I think we can remove this as well ?

We need to keep state of checkboxes?

PR #73, opened 2020-04-08
Add  extremities to line positions

ChamseddineBhd:


src/components/network/geo-data.js
```diff
@@ -15,6 +15,29 @@ const linePositionIndexer = (map, line) => {
     return map;
 };
 
+function distance(lat1, lon1, lat2, lon2, unit) {
```
geofjamg, line 18: Do you had a look at existing library for that? Maybe deckgl/mapbox alreadu have one.

ChamseddineBhd: Done

PR #77, opened 2020-04-08
Creation of the apps menu

niconoir:
Signed-off-by: NOIR Nicolas ext <nicolas.noir@rte-france.com>

- [X] The commit message follows our guidelines

This PR adds a new button on the top bar of the application to display available apps in GridSuite

**Other information**:


geofjamg:
image

When we click on the app button the (only) item appears to be blue because of focus. Is there a way to avoid this?

Also maybe the icon is a to big?

niconoir:
@mikatheo @geofjamg Do you think the button should appear even if the user is not connected ?

mikatheo:
> @mikatheo @geofjamg Do you think the button should appear even if the user is not connected ?

Good question. I’m for a pragmatic approach : no button if I’m not connected, because not usable.

niconoir:
> 
> When we click on the app button the (only) item appears to be blue because of focus. Is there a way to avoid this?
> 
> Also maybe the icon is a to big?

I agree the icon is maybe too big. I will try to reduce its size. I'll also take a look at the focus behaviour.

niconoir:
> > When we click on the app button the (only) item appears to be blue because of focus. Is there a way to avoid this?
> > Also maybe the icon is a to big?
> 
> I agree the icon is maybe too big. I will try to reduce its size. I'll also take a look at the focus behaviour.

If we remove focus from the Apps menu, the behaviour will be different from the one of the Preferences menu (which automatically gives focus to the first item of the menu). Do we want these two menus to behave differently ? I agree it seems strange to have blue focus on the single item available of the menu, but when there will be more items in the menu, it will make sense...

geofjamg:
> > > When we click on the app button the (only) item appears to be blue because of focus. Is there a way to avoid this?
> > > Also maybe the icon is a to big?
> > 
> > 
> > I agree the icon is maybe too big. I will try to reduce its size. I'll also take a look at the focus behaviour.
> 
> If we remove focus from the Apps menu, the behaviour will be different from the one of the Preferences menu (which automatically gives focus to the first item of the menu). Do we want these two menus to behave differently ? I agree it seems strange to have blue focus on the single item available of the menu, but when there will be more items in the menu, it will make sense...

Yes, agreed

PR #78, opened 2020-04-09
Add settings for labels display

JacquesBorsenberger:
svg bounding box change for negative point of origin

reorganise option in multiples tabs

### DIFF
--- a/src/components/parameters.js
+++ b/src/components/parameters.js
@@ -54,7 +54,7 @@ const Parameters = () => {
     const useName = useSelector(state => state.useName);
     const centerLabel = useSelector(state => state.centerLabel);
     const diagonalLabel = useSelector(state => state.diagonalLabel);
-    const [tabIndex, setTabIndex] = React.useState(useSelector(state => state.parametersTabIndex));
+    const [tabIndex, setTabIndex] = React.useState(0);
 
     const theme = useSelector(state => state.theme);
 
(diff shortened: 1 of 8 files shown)


src/components/parameters.js
```diff
@@ -48,13 +52,12 @@ const Parameters = () => {
     const classes = useStyles();
 
     const useName = useSelector(state => state.useName);
+    const centerLabel = useSelector(state => state.centerLabel);
+    const diagonalLabel = useSelector(state => state.diagonalLabel);
+    const [tabIndex, setTabIndex] = React.useState(useSelector(state => state.parametersTabIndex));
```
jonenst, line 57: Why does this come from the redux store ? To preserve between different router mount/unmounts ? But I don't see where the redux state is updated

src/redux/reducer.js
```diff
@@ -42,6 +44,9 @@ const initialState = {
     selectedFile : null,
     useName : getLocalStorageUseName(),
     user : null,
+    centerLabel : false,
+    diagonalLabel : false,
+    parametersTabIndex: 0,
```
jonenst, line 49: where is this updated ?

src/components/parameters.js
jonenst: If we use the redux store, we shouldn't copy the state in this component I think ?

jonenst:
> Fix 'Driagram' type

Funny to make a typo when asking to fix a typo :)

PR #89, opened 2020-04-15
Add check all uncheck all to Voltages Level menu

JacquesBorsenberger:

### DIFF
--- a/src/components/study-pane.js
+++ b/src/components/study-pane.js
@@ -187,6 +187,10 @@ const StudyPane = () => {
         setFilteredNominalVoltages(newFiltered);
     };
 
+    const toggleCheckAll = (toCheck) => {
+        setFilteredNominalVoltages(toCheck ? network.getNominalVoltages() : [] );
+    };
+
     if (studyNotFound) {
         return <StudyNotFound studyName={studyName}/>;
     } else {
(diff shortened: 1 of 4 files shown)


src/components/study-pane.js
```diff
@@ -230,7 +234,8 @@ const StudyPane = () => {
                             <div style={{position: "absolute", right: 10, bottom: 30, zIndex: 1}}>
                                 <NominalVoltageFilter nominalVoltages={network.getNominalVoltages()}
                                                       filteredNominalVoltages={filteredNominalVoltages}
-                                                      onNominalVoltageFilterChange={updateFilteredNominalVoltages} />
+                                                      onNominalVoltageFilterChange={updateFilteredNominalVoltages}
+                                                      onCheckAll={toggleCheckAll}/>
```
geofjamg, line 238: Why not just using the same callback onNominalVoltageFilterChange?

mikatheo:
Is it possible to add a hand-shaped cursor as for the checkboxes?

src/components/study-pane.js
JacquesBorsenberger: the parameter type is not the same (int / bool).

geofjamg: But you could with the other callback just notify with all or none nominal voltage in the list?

JacquesBorsenberger: the other callback toggle the state, if I notify all, it will invert the state, if I notify none, it will do nothing; 
I can test the parameter type in the callback.

geofjamg: I see the point: right now onNominalVoltageFilterChange can only take on one nominal voltage and that is used to either add or remove the nominal voltage if found or not. In order to have only one callback which seems to me a good thing we could refactor it to have 2 parameters, one with a list of nominal voltage to add, one with a list of nominal voltage to remove?

It is just a proposition. Do you think it would a better design?

JacquesBorsenberger: I agree that only one callback would be better. 
perhaps 1 list and a boolean  defining toggle or set behavior would be simpler, otherwise the handle toggle would have to check the state.

geofjamg: Agree, for a list + a boolean. That would be the best solution.

PR #92, opened 2020-04-15
Use gateway 

ChamseddineBhd:
Should be merged after merging : 
    - https://github.com/gridsuite/study-app/pull/97
    - https://github.com/gridsuite/deployment/pull/28
    - https://github.com/gridsuite/gateway/pull/1


jonenst:
I think we should design a more global system instead of having to manually pass the token at each fetch. Can you try to design that ?

jonenst:
Also refactor having to add the Authorization: Bearer manually for each fetch call

src/utils/rest-api.js
```diff
@@ -1,95 +1,134 @@
+import {store} from '../redux/store'
```
jonenst, line 1: put this line after the copyright notice

src/utils/rest-api.js
```diff
@@ -1,95 +1,134 @@
...
+}
+
 export function fetchStudies() {
     console.info("Fetching studies...");
-    const fetchStudiesUrl = process.env.REACT_APP_API_STUDY_SERVER + "/v1/studies";
+    const fetchStudiesUrl = PREFIX_STUDY_QUERIES + "/v1/studies";
     console.debug(fetchStudiesUrl);
-    return fetch(fetchStudiesUrl)
-        .then(response => response.json());
+    return fetch(fetchStudiesUrl, {
```
jonenst, line 31: create a method backendFetch(..) which delegates to fetchs and puts the auth header when USE_AUTHENTICATION is true

src/utils/rest-api.js
```diff
@@ -4,82 +4,118 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
+import {store} from '../redux/store'
```
jonenst, line 7: Ideally this would not be a global but let's save this for another PR as this is already existing

PR #97, opened 2020-04-17
Token silent renew 

ChamseddineBhd:
Should be merged after merging token-silent-renw branch in powsybl-deployment


src/utils/authentication/AuthService.js
```diff
@@ -94,4 +97,11 @@ function handleSigninCallback(dispatch, history, userManagerInstance) {
     });
 }
 
-export {userManagerPromise, login, logout, dispatchUser, handleSigninCallback, getPreLoginPath}
+function handleSilentRenewCallback(dispatch, userManagerInstance) {
+    userManagerInstance.signinSilentCallback().catch(function (e) {
+        dispatch(setLoggedUser(null));
```
jonenst, line 102: What does this do ?

ChamseddineBhd: If there's an issue with the silent renew we should force our user to login explicitly, to do that we empty the user in the store.

jonenst: but this is executed in the iframe no ?

ChamseddineBhd: Alright

PR #98, opened 2020-04-19
Create arrow layer

geofjamg:
Signed-off-by: Geoffroy Jamgotchian <geoffroy.jamgotchian@rte-france.com>


src/components/network/layers/arrow-layer.js
```diff
@@ -108,9 +108,17 @@ export default class ArrowLayer extends Layer {
 
     createTexture2D(gl, data, elementSize, format) {
         const elementCount = data.length / elementSize;
+        const width = Math.min(elementCount, this.state.maxTextureSize);
+        const height = Math.ceil(elementCount / this.state.maxTextureSize);
+        console.info(`Creating texture of ${elementCount} elements (${width} * ${height})`);
+        if (data.length < width * height * elementSize) {
+            const oldLength = data.length;
+            data.length += (width * height * elementSize - data.length);
```
jonenst, line 133: data.length = width * height * elementSize  ?

geofjamg: Fixed, shame on me this is the ugliest line of code ever...

src/components/network/layers/arrow-layer.js
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    animate(timestamp) {
+        if (this.state.stop) {
+            return;
+        }
+        this.setState({
+            timestamp: timestamp,
+        });
+        this.startAnimation();
```
jonenst, line 281: since this is running at 60 fps, maybe it's best to avoid recreating the lambda everytime and juste call requestAnimationFrame(animate) ?

src/components/network/layers/arrow-layer.js
```diff
@@ -0,0 +1,342 @@
...
+    };
+
+    startAnimation() {
+        window.requestAnimationFrame(timestamp => this.animate(timestamp));
```
jonenst, line 285: window.requestAnimationFrame(this.animate) ?

geofjamg: The problem is that if I do that I need to bind the animate method to this.

src/components/network/layers/arrow-layer-vertex.glsl.js
```diff
@@ -0,0 +1,134 @@
...
+  int flatIndex = instanceLineDistancesTextureOffset + point;
+  ivec2 textureIndex = calulateTextureIndex(flatIndex);
+  return texelFetch(lineDistancesTexture, textureIndex, 0).x;
+}
+
+/**
+ * Find the first point of the line that is after a given distance from the start (first line point).
+ * (implemented using a binary search)
+ */
+int findFirstLinePointAfterDistance(float distance) {
```
jonenst, line 58:
```
/**            
 * Find the first point of the line that is after a given distance from the start (first line point).
 * (implemented using a binary search)
 * The returned value is always between 1 and instanceLinePointCount - 1, even if the searched distance is out of bounds
 * Here are example returned values for a path having points at distance 0.0, 10.0, 20.0
 * -1 => 1
 *  0 => 1
 *  1 => 1
 *  9 => 1
 *  10 => 2
 *  11 => 2
 *  19 => 2
 *  20 => 2
 *  21 => 2
 */
int findFirstLinePointAfterDistance(float distance) {
  int firstPoint = 0;
  int lastPoint = instanceLinePointCount - 1;
  
  // variable length loops are not supported in GLSL, instanceLinePointCount is an upper bound that
  // will never be reached as binary search complexity is in O(log(instanceLinePointCount))
  for (int i = 0; i < instanceLinePointCount; i++) {
      if (firstPoint + 1 == lastPoint) {
          return lastPoint; 
      }   
      int middlePoint = (firstPoint + lastPoint) / 2;           
      float middlePointDistance = fetchLineDistance(middlePoint);      
      if (middlePointDistance <= distance) {
         firstPoint = middlePoint;
      } else {
         lastPoint = middlePoint;                            
      }  
  }   
} 
```

src/components/network/layers/arrow-layer-vertex.glsl.js
```diff
@@ -0,0 +1,134 @@
...
+  }
+}
+
+void main(void) {
+  if (instanceArrowDirection < 1.0) {
+      vFillColor = vec4(0, 0, 0, 0);
+      shouldDiscard = 1.0;
+  } else {
+      // arrow distance from the line start shifted with current timestamp
+      float arrowDistance = mod(instanceLineDistance * instanceArrowDistance + (instanceArrowDirection < 2.0 ? 1.0 : -1.0) * timestamp * instanceSpeedFactor, instanceLineDistance);
```
jonenst, line 91: float arrowDistance = mod((instanceArrowDirection < 2.0 ? 1.0 : -1.0) * timestamp * instanceSpeedFactor, instanceLineDistance);

mod(a*b+X, a) == mod(X, a) no ??

src/components/network/layers/arrow-layer-vertex.glsl.js
geofjamg: It works well thanks.

src/components/network/layers/arrow-layer-vertex.glsl.js
geofjamg: even if b < 1 ?

jonenst: Ah yeah only when b is an integer of course. So you're right, here it's absolutely needed other all arrows are at the same position

jonenst:
When I change the "full line path" setting it doesn't recompute the arrows paths

jonenst:
When the arrowmode prop is switched from STATIC to NONE or ANIMATED, it doesn't take it into account.

src/components/network/layers/arrow-layer-vertex.glsl.js
jonenst: somewhere we should add a comment :
instanceArrowDistance: a float in interval [0,1] describing the initial position of the arrow

src/components/parameters.js
```diff
@@ -152,20 +164,44 @@ const Parameters = ({showParameters, hideParameters}) => {
...
+                <Grid item xs={6}>
+                    <Typography component="span" variant="body1">
+                        <Box fontWeight="fontWeightBold" m={1}>
+                            <FormattedMessage id="LineFlowMode"/>:
+                        </Box>
+                    </Typography>
+                </Grid>
+                <Grid item container xs={6} className={classes.controlItem}>
+                    <Select labelId="line-flow-mode-select-label" value={lineFlowMode} onChange={handleLineFlowModeChange}>
+                        <MenuItem value={LineFlowMode.NONE}>None</MenuItem>
```
jonenst, line 181: i18n for these ?

geofjamg: Done

src/components/network/layers/arrow-layer-vertex.glsl.js
geofjamg: comment added

geofjamg:
> When I change the "full line path" setting it doesn't recompute the arrows paths

fixed

jonenst:
AWESOME ! :)

PR #99, opened 2020-04-20
Remove scollbars and change zoom speed on voltage level diagram

FranckLecuyer:
Change zoom factor of panZoom component to increase zoom speed

### DIFF
--- a/src/components/single-line-diagram.js
+++ b/src/components/single-line-diagram.js
@@ -29,8 +29,8 @@ const useStyles = makeStyles(theme => ({
     div: {
         maxWidth: maxWidth,
         maxHeight: maxHeight,
-        overflowX: 'auto',
-        overflowY: 'auto'
+        overflowX: 'hidden',
+        overflowY: 'hidden'
     },
     diagram: {
         "& .component-label": {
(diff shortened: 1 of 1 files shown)


src/components/single-line-diagram.js
```diff
@@ -98,7 +98,7 @@ const SingleLineDiagram = (props) => {
                 .addTo(divElt)
                 .size(svgWidth, svgHeight)
                 .viewbox(xOrigin, yOrigin, svgWidth, svgHeight)
-                .panZoom({panning: true, zoomMin: 0.5, zoomMax: 10, zoomFactor: 0.2, margins: {top: svgHeight/4, left: svgWidth/4, bottom: svgHeight/4, right: svgWidth/4}});
```
jonenst, line 101: On my mouse, to have a "normal" experience, I need to use zoomFactor = 10 or 20 !  What does it do on your mouse ??

jonenst: even 200 feels natural

FranckLecuyer: With my mouse, when i set the zoom factor to 10, it zooms a lot with only one wheel step ...

jonenst: Maybe we need to add a parameter next to the "center" and "diagonal labels" ?

FranckLecuyer: why not ...

PR #101, opened 2020-04-23
make parameter window modal

ChamseddineBhd:

### DIFF
--- a/src/components/parameters.js
+++ b/src/components/parameters.js
@@ -26,18 +26,11 @@ import Switch from "@material-ui/core/Switch";
 import Typography from "@material-ui/core/Typography";
 
 import {DARK_THEME, LIGHT_THEME, selectTheme, toggleUseNameState, toggleCenterLabelState, toggleDiagonalLabelState} from "../redux/actions";
-import AppBar from "@material-ui/core/AppBar";
 import Tabs from "@material-ui/core/Tabs";
 import Tab from "@material-ui/core/Tab";
 import Dialog from "@material-ui/core/Dialog";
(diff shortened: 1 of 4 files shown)


mikatheo:
> image
> 
> Could you try to increase the height of the dialog? It looks strange like this, less clear than previous version.
At least to add a margin between the button and the bottom of the window
image

src/components/parameters.js
```diff
@@ -146,30 +147,41 @@ const Parameters = () => {
             )
     }
 
+    function handleCloseDialog() {
+
+    }
```
mikatheo, line 152: Is it necessary to keep this empty function?

src/components/parameters.js
```diff
@@ -43,7 +44,7 @@ const useStyles = makeStyles(theme => ({
     }
 }));
 
-const Parameters = () => {
+const Parameters = (props) => {
```
mikatheo, line 47: Just a naive question here:
- I find it more judicious to add only the useful props rather all of them. Is there a good practice in ReactJS or is it up to the developper ?

src/components/parameters.js
```diff
@@ -64,7 +65,7 @@ const Parameters = () => {
     };
 
     const handleClose = () => {
-        history.goBack();
+
     };
```
mikatheo, line 69: Also empty function

src/components/parameters.js
ChamseddineBhd: I think that the way you recommend is best than passing just props because it lets us  declare explicitly our properties so we can know what our component is exactly waiting for.

PR #102, opened 2020-04-23
Fix warning

ChamseddineBhd:


geofjamg:
If you can describe a little what was the error and why.

ChamseddineBhd:
> If you can describe a little what was the error and why.

Failed prop type: The prop `justify` of `Grid` must be used on `container`.

PR #103, opened 2020-04-23
rename study

AbdelHedhili:


src/components/study-manager.js
```diff
@@ -253,6 +294,23 @@ const StudyCard = ({study, onClick}) => {
                     </Button>
                 </DialogActions>
             </Dialog>
+            <Dialog open={openRenameDialog} onClose={handleCloseRenameDialog} aria-labelledby="form-dialog-title">
```
jonenst, line 297: the aria-labelledby is the id of the DialogTitle and should be unique

src/components/study-manager.js
```diff
@@ -89,7 +92,9 @@ const useStyles = makeStyles((theme) => ({
 }));
 
 const StudyCard = ({study, onClick}) => {
-    const [open, setOpen] = React.useState(false);
+    const [openDeleteDialog, setOpenDelete] = React.useState(false);
+    const [openRenameDialog, setOpenRename]  = React.useState(false);
+    const [newStudyNameValue, setNewStudyName] = React.useState(study.studyName);
```
jonenst, line 97: I think it's weird to put this state in the studycard. Maybe extract it to another component ? The alternative is to use uncontrolled forms but that's a bit of an antipattern in react

jonenst: To be more precise, can we use more components to restrict the scope of relayouting (at each keystroke in input in the dialog) ?

ie move the studyName in something like:
```
const renameDialog = () => {
     const [newStudyNameValue, setNewStudyName] = React.useState(study.studyName); 
}
```

AbdelHedhili: @jonenst I made the change for renameDialog, is that what you expected ?

src/components/study-manager.js
```diff
@@ -253,10 +288,48 @@ const StudyCard = ({study, onClick}) => {
                     </Button>
                 </DialogActions>
             </Dialog>
+            <RenameDialog study={study}
```
jonenst, line 291: pass studyname directly to props

src/components/study-manager.js
```diff
@@ -253,10 +288,48 @@ const StudyCard = ({study, onClick}) => {
...
+        console.log(newStudyNameValue);
+        props.handleConfirm(newStudyNameValue);
+    };
+
+    return (
+        <Dialog open={props.openRenameDialog} onClose={props.handleCloseDialog} aria-labelledby="dialog-title-rename">
+            <DialogTitle id="dialog-title-rename"><FormattedMessage id="renameStudy"/></DialogTitle>
+            <DialogContent>
+                <InputLabel htmlFor="newStudyName"><FormattedMessage id="renameStudyMsg"/></InputLabel>
+                <TextField id="newStudyName" defaultValue={props.study.studyName} required={true} onChange={updateStudyNameValue}></TextField>
```
jonenst, line 319: We should either use value and a state (controlled component) or defautlValue and  a ref (uncontrolled component). Controlled components are the officially recommended solution

PR #104, opened 2020-04-23
Full screen mode

ChamseddineBhd:

### DIFF
--- a/src/components/top-bar.js
+++ b/src/components/top-bar.js
@@ -5,7 +5,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import React from "react";
+import React, {useState} from "react";
 
 import {FormattedMessage} from "react-intl";
 import {useHistory} from 'react-router-dom';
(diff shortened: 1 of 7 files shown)


src/components/top-bar.js
```diff
@@ -120,6 +120,38 @@ const TopBar = (props) => {
         history.replace("/");
     };
 
+    function requestFullscreen (ele) {
```
jonenst, line 123: Is it not possible to avoid this polyfilling ?

src/components/top-bar.js
```diff
@@ -90,6 +93,12 @@ const TopBar = (props) => {
 
     const [anchorElAppsMenu, setAnchorElAppsMenu] = React.useState(null);
 
+    const [ref, setRef] = useState(null);
```
jonenst, line 96: Use a more explicit variable name ?

jonenst: can we use "useRef" instead of a callback ref ?

PR #113, opened 2020-04-29
click on a switch to open/close it in svg

AbdelHedhili:


src/components/study-pane.js
```diff
@@ -175,6 +176,23 @@ const StudyPane = () => {
         history.replace("/studies/" + studyName)
     }
 
+    const handleUpdateSwitchState = useCallback( (studyName, breakerId, open, clickedDomElement, breakerMetadata) => {
+        updateSwitchState(studyName, breakerId, open).then( response => response.ok ? updateSvgSwitch(clickedDomElement, open, breakerMetadata) : console.debug(response));
+    }, []);
+
+    function updateSvgSwitch(clickedDomElement, open, breakerMetadata) {
```
geofjamg, line 183: This is not the right way to do it as for instance when topological coloring will be activated, update can only be done on server side. So we need to reload again from the server the SVG generated with the new topology.

src/components/single-line-diagram.js
```diff
@@ -77,9 +81,11 @@ const SingleLineDiagram = (props) => {
                     setSvg({svg: null, metadata: null, error, svgUrl: props.svgUrl});
                 });
         } else {
-            setSvg(nosvg);
+            setSvg(noSvg);
         }
-    }, [props.svgUrl]);
+    }, [props.svgUrl, useSelector(state => state.currentSvg)]);
```
jonenst, line 86: Hmm a dependency that is not used ? Is there no other way ?

src/components/single-line-diagram.js
```diff
@@ -91,16 +97,20 @@ const SingleLineDiagram = (props) => {
...
             const svgHeight = bbox.height + 40;
-
             // using svgdotjs panzoom component to pan and zoom inside the svg, using svg width and height previously calculated for size and viewbox
             divElt.innerHTML = ""; // clear the previous svg in div element before replacing
             const draw = SVG()
                 .addTo(divElt)
                 .size(svgWidth, svgHeight)
                 .viewbox(xOrigin, yOrigin, svgWidth, svgHeight)
                 .panZoom({panning: true, zoomMin: 0.5, zoomMax: 10, zoomFactor: 0.3, margins: {top: svgHeight/4, left: svgWidth/4, bottom: svgHeight/4, right: svgWidth/4}});
+            console.log(svgDisplayInfo);
```
jonenst, line 107: remove debug log

jonenst:
I don't think the state should be in the redux store, it's very local to the study-pane

src/components/single-line-diagram.js
jonenst: Ah I see, then It's better to use a imperative handle to force the refresh of the svg I think

src/components/single-line-diagram.js
```diff
@@ -77,9 +81,11 @@ const SingleLineDiagram = (props) => {
                     setSvg({svg: null, metadata: null, error, svgUrl: props.svgUrl});
                 });
         } else {
-            setSvg(nosvg);
+            setSvg(noSvg);
         }
-    }, [props.svgUrl]);
+    }, [props.svgUrl, useSelector(state => state.currentSvg)]);
+
+    const svgDisplayInfo = useSelector(state => state.svgDisplayInfo);
```
jonenst, line 88: should be a regular prop

src/components/study-pane.js
```diff
@@ -230,6 +236,8 @@ const StudyPane = () => {
                             <div style={{ position: "absolute", left: 10, top: 10, zIndex: 1 }}>
                                 <SingleLineDiagram onClose={() => closeVoltageLevelDiagram()}
                                                    onNextVoltageLevelClick={showVoltageLevelDiagram}
+                                                   onBreakerClick={handleUpdateSwitchState}
+                                                   studyName={studyName}
```
jonenst, line 240: should not need to pass the studyname by props

src/components/single-line-diagram.js
```diff
@@ -60,11 +61,13 @@ const SvgNotFound = (props) => {
...
 
-const nosvg = {svg: null, metadata: null, error: null, svgUrl: null};
+const noSvg = {svg: null, metadata: null, error: null, svgUrl: null};
 
 const SingleLineDiagram = (props) => {
 
-    const [svg, setSvg] = useState(nosvg);
+    const [svg, setSvg] = useState(noSvg);
+
+    const studyName = useSelector(state => state.studyName);
```
jonenst, line 70: I don't think this component should know the study name. It already gets the full svg url containing the studyname from the parent component. Instead, it should expose a method in its ref (this is done with useImperativeHandle) to refresh the svg

src/components/single-line-diagram.js
```diff
@@ -77,9 +80,11 @@ const SingleLineDiagram = (props) => {
                     setSvg({svg: null, metadata: null, error, svgUrl: props.svgUrl});
                 });
         } else {
-            setSvg(nosvg);
+            setSvg(noSvg);
         }
-    }, [props.svgUrl]);
```
jonenst, line 82: We shouldn't need to store the full svg just to test if it's the same as the previous one

src/components/study-pane.js
```diff
@@ -175,6 +180,22 @@ const StudyPane = () => {
...
+                setCurrentSvg(fetchSvg(svgUrl));
+                setSvgDisplayInfo(svgDisplayInfo);
+            }
+            else {
+                console.error(response);
+            }
+        });
+    }, []);
+
+    const resetSvgDisplayInfo = useCallback(() => {
```
jonenst, line 195: This should be replaced with the imperative ref

src/components/study-pane.js
```diff
@@ -175,6 +180,22 @@ const StudyPane = () => {
         history.replace("/studies/" + studyName)
     }
 
+    const handleUpdateSwitchState = useCallback( (studyName, breakerId, open, svgUrl, svgDisplayInfo) => {
+        updateSwitchState(studyName, breakerId, open).then( response => {
+            if (response.ok) {
+                setCurrentSvg(fetchSvg(svgUrl));
+                setSvgDisplayInfo(svgDisplayInfo);
```
jonenst, line 187: svgDisplayInfo should be local to the function that forces the refresh of an svg in the sld

src/components/single-line-diagram.js
```diff
@@ -118,6 +130,21 @@ const SingleLineDiagram = (props) => {
                     const meta = svg.metadata.nodes.find( other => other.id === id );
                     props.onNextVoltageLevelClick(meta.nextVId);
                 })});
+
+            // handling the click on a breaker
+            const breakers = svg.metadata.nodes.filter(element => element.componentType === "BREAKER");
+            breakers.forEach(breaker => {
+                console.log(breaker);
```
jonenst, line 137: remove debug log

src/utils/rest-api.js
```diff
@@ -117,6 +117,13 @@ export function deleteStudy(studyName) {
     });
 }
 
+export function updateSwitchState(studyName, switchId, open) {
+    console.info("updating switch " + switchId + " ...");
+    const updateSwitchUrl = process.env.REACT_APP_API_STUDY_SERVER + "/v1/studies/" + studyName + "/network-modification/switches/" + switchId + "?open=" + open;
+    console.debug(updateSwitchUrl);
+    return fetch(updateSwitchUrl, {method : 'put'});
```
jonenst, line 124: use (newly merged) backendFetch when merging master into this

src/components/study-pane.js
```diff
@@ -89,6 +90,10 @@ const StudyPane = () => {
 
     const [filteredNominalVoltages, setFilteredNominalVoltages] = useState([]);
 
+    const [currentSvg, setCurrentSvg] = useState(null);
```
jonenst, line 93: remove this ?

src/components/single-line-diagram.js
```diff
@@ -60,11 +61,19 @@ const SvgNotFound = (props) => {
...
+const SingleLineDiagram = forwardRef((props, ref)  => {
 
-    const [svg, setSvg] = useState(nosvg);
+    const [svg, setSvg] = useState(noSvg);
+    const [svgDisplayInfo, setSvgDisplayInfo] = useState();
+
+    useImperativeHandle(ref, () => ({
+        reloadSvg: (svgUrl, svgDisplayInfo) => {
+            fetchSvg(svgUrl).then(svg => setSvg(svg));
+            setSvgDisplayInfo(svgDisplayInfo);
```
jonenst, line 74: i don't think svgDisplayInfo should be a state. instead it should be a ref that doesn't trigger relayouting when it is changed.

src/components/single-line-diagram.js
```diff
@@ -60,11 +61,19 @@ const SvgNotFound = (props) => {
...
-const SingleLineDiagram = (props) => {
+const SingleLineDiagram = forwardRef((props, ref)  => {
 
-    const [svg, setSvg] = useState(nosvg);
+    const [svg, setSvg] = useState(noSvg);
+    const [svgDisplayInfo, setSvgDisplayInfo] = useState();
+
+    useImperativeHandle(ref, () => ({
+        reloadSvg: (svgUrl, svgDisplayInfo) => {
+            fetchSvg(svgUrl).then(svg => setSvg(svg));
```
jonenst, line 73: I don't think we should duplicate the fetchSvg call here. Instead we should just force a relayout of component:
```
  const [, updateState] = React.useState();
  const forceUpdate = React.useCallback(() => updateState({}), []);
```

src/components/single-line-diagram.js
```diff
@@ -118,6 +130,21 @@ const SingleLineDiagram = (props) => {
...
+            breakers.forEach(breaker => {
+                console.log(breaker);
+                const domEl = document.getElementById(breaker.id);
+                domEl.style.cursor = "pointer";
+                domEl.addEventListener("click", function(event) {
+                    const clickedElementId = event.currentTarget.id;
+                    const breakerMetadata = svg.metadata.nodes.find(value => value.id === clickedElementId);
+                    const breakerId = breakerMetadata.equipmentId;
+                    const open = breakerMetadata.open;
+                    props.onBreakerClick(breakerId, !open, props.svgUrl, {viewBox :draw.viewbox(), zoom: draw.zoom()});
```
jonenst, line 145: shouldn't pass the viewbox and zoom to parent. Instead store them in a ref

jonenst: shouldn't pass the svgUrl to the parent

src/components/study-pane.js
```diff
@@ -89,6 +90,10 @@ const StudyPane = () => {
...
+
+    const [svgDisplayInfo, setSvgDisplayInfo] = useState(null);
```
jonenst, line 95: remove this ?

src/components/study-pane.js
```diff
@@ -175,6 +180,17 @@ const StudyPane = () => {
...
+    const handleUpdateSwitchState = useCallback( (breakerId, open, svgUrl, svgDisplayInfo) => {
+        updateSwitchState(studyName, breakerId, open).then( response => {
+            if (response.ok) {
+                reloadSvg(svgUrl, svgDisplayInfo);
+            }
+            else {
+                console.error(response);
+            }
+        });
+    }, []);
```
jonenst, line 192: missing studyName from the dependencies

src/components/study-pane.js
```diff
@@ -199,6 +215,11 @@ const StudyPane = () => {
         mapRef.current.centerSubstation(network.getVoltageLevel(id).substationId);
     }, [mapRef, network]);
 
+    const sldRef = useRef();
+    const reloadSvg = useCallback((svgUrl, svgDisplayInfo)=> {
+        sldRef.current.reloadSvg(svgUrl, svgDisplayInfo);
+    }, [sldRef]);
```
jonenst, line 221: I don't think you need sldRef in the deps, it doesn't change

jonenst:
@geofjamg can you review again and mark your change requested as resolved if okay ?

PR #116, opened 2020-05-04
minor fix + changed the rename icon

AbdelHedhili:

When renaming a study if we pressed cancel after typing a new name and then reopened the rename popup, the name printed was the one we typed before canceling. 
Now if we cancel and reopen the study name is displayed.

also changed the icon for the rename option.


jonenst:
>  There is a problem from the first dev in the lines 186/187, an inconsistency between rename and delete option.

good catch, I fixed it and now clicking outside of the rename popup does close the rename popup

jonenst:
when we click outside the rename dialog to close it, do we want the modified text to be reset like when we click cancel or not (if reclicking rename)?

PR #117, opened 2020-05-06
Refactor the study-app manager

mikatheo:
- [x] Reorganization of the 'handle' functions
- [x] PropTypes validation for each component
- [x] Enter to close/validate the actions
- [x] Handle the warnings
- [x] Extract the 2 dialogs to generalise them

### DIFF
--- a/src/components/study-manager.js
+++ b/src/components/study-manager.js
@@ -10,20 +10,11 @@ import PropTypes from 'prop-types';
 import {useDispatch, useSelector} from "react-redux";
 
 import {FormattedMessage} from "react-intl";
-
-import Grid from '@material-ui/core/Grid';
 import {makeStyles} from "@material-ui/core/styles";
 import Typography from '@material-ui/core/Typography';
 import Card from '@material-ui/core/Card';
 import CardActionArea from '@material-ui/core/CardActionArea';
 import CardContent from '@material-ui/core/CardContent';
-import Container from '@material-ui/core/Container';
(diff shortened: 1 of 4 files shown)


src/components/study-manager.js
```diff
@@ -133,43 +133,55 @@ const StudyCard = ({study, onClick}) => {
         />
     ));
 
+    // ########################################################################
```
jonenst, line 136: We don't have things like in our source code.. Not sure about adding them. @geofjamg ?

geofjamg: Personnally I hate this kind of ascii art separator. It doesn't clarify anything just:

// action menu

is enough.

mikatheo: I agree it was unconventional and ugly here, but could we consider documenting the Front by using a standard like this?
- https://react-styleguidist.js.org/docs/documenting.html

Personally I regret the lack of comment, even though I haven't put much into it myself so far. It's superfluous for some, but more readable for others. This will be essential at least for the common-ui repo, so I could start here.

geofjamg: Everybody is welcome to add as much comment as possible as far as it is done in a coventional and normalized way.

mikatheo:
My proposal for the dialogs (focus on the textField at opening, and circled button for the default one)
image
image

mikatheo:
@jonenst New commit with management of unique identifiers

PR #123, opened 2020-05-13
Port to WebGL v1

geofjamg:
For Edge, Safari compatibility which does not support WebGL v2

### DIFF
--- a/src/components/network/layers/arrow-layer-vertex.glsl.js
+++ b/src/components/network/layers/arrow-layer-vertex.glsl.js
@@ -37,7 +37,7 @@ vec4 texelFetch(sampler2D sampler, ivec2 index, ivec2 size) {
  * Calculate 2 dimensions texture index from flat index. 
  */
 ivec2 calculateTextureIndex(int flatIndex, ivec2 textureSize) {
-  int x = int(mod(float(flatIndex), float(textureSize.x)));
+  int x = flatIndex - flatIndex / textureSize.x * textureSize.x;
   int y = flatIndex / textureSize.y;
   return ivec2(x, y);
 }
--- a/src/components/network/layers/arrow-layer-vertex.glsl.js
+++ b/src/components/network/layers/arrow-layer-vertex.glsl.js
@@ -33,16 +33,12 @@ vec4 texelFetch(sampler2D sampler, ivec2 index, ivec2 size) {
   return texture2D(sampler, vec2(x, y));
 }
 
-float trunc(float v) {
-  return float(int(v));
-}
-
 /**
  * Calculate 2 dimensions texture index from flat index. 
  */
 ivec2 calulateTextureIndex(int flatIndex, ivec2 textureSize) {
   int x = int(mod(float(flatIndex), float(textureSize.x)));
-  int y = int(trunc(float(flatIndex) / float(textureSize.y)));
+  int y = int(float(flatIndex) / float(textureSize.y));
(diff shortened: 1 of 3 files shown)


src/components/network/layers/arrow-layer-vertex.glsl.js
```diff
@@ -68,23 +83,29 @@ float fetchLineDistance(int point) {
...
+          float middlePointDistance = fetchLineDistance(middlePoint);      
+          if (middlePointDistance <= distance) {
+             firstPoint = middlePoint;
+          } else {
+             lastPoint = middlePoint;                            
+          }
+      }
+  }
+  return lastPoint; 
+}
```
jonenst, line 108: Can you test this loop, is it better ?
```
  const int log2MaxPointCount = 15;
  for (int i = 0; i < log2MaxPointCount; i++) {
      int middlePoint = (firstPoint + lastPoint + 1) / 2;           
      float middlePointDistance = fetchLineDistance(middlePoint);      
      int a = int(step(middlePointDistance, distance));
      int b = 1 - a;
      firstPoint = b * firstPoint + a * middlePoint;
      lastPoint = a * lastPoint + b * middlePoint;                            
  }
```
could also be
```
firstPoint = xx ? firstPoint: middlePoint;
lastPoint = !xx ? lastPoint : middlePoint;
`


Trying to make the code fully branchless as per popular wisdom but I don't know if it's useful

Comment (jonenst) [reply src/components/network/layers/arrow-layer-vertex.glsl.js:108]:
```
      int a = int(distance < middlePointDistance);
      int b = 1 - a;
      firstPoint = a * firstPoint + b * middlePoint;
      lastPoint = b * lastPoint + a * middlePoint;                            
```
works too

jonenst: I use
     function onAfterRender() {
+        if (Math.random() > 0.99) {
+            console.log(deck.metrics);
+        }
for quick and dirty fps monitoring the browser console.
Using this, I get about 25fps on a big french network with the real lines. Actually depending on the number of iterations, I'm getting
1 iterations: 30fps
10 iterations: 28fps
100 iterations: 25fps
1000 iterations: 8fps

geofjamg: It works but it is less optimized as first if avoid many texture fetch

geofjamg: for (int i = 0; i < log2MaxPointCount; i++) {
      if (firstPoint + 1 != lastPoint) {
          int middlePoint = (firstPoint + lastPoint + 1) / 2;           
          float middlePointDistance = fetchLineDistance(middlePoint);      
          int a = int(step(middlePointDistance, distance));
          int b = 1 - a;
          firstPoint = b * firstPoint + a * middlePoint;
          lastPoint = a * lastPoint + b * middlePoint;  
      }
  }

geofjamg: Still optimized but with one branch less

jonenst: Keep your "if" gives actually the best performance. I suspect this is because it removes expensive texelFetch calls. Let's keep it as is !

geofjamg: So I just keep the fist if or all ?

PR #124, opened 2020-05-14
Addition of loadflow button

niconoir:


niconoir:
> Could you change the button color to green and add margins ?

Color changed and margins added

PR #125, opened 2020-05-15
Page not found

ChamseddineBhd:
Add Not found page to the app


src/components/app.js
```diff
@@ -29,6 +29,7 @@ import {
 } from '../utils/authentication/AuthService';
 
 import Authentication from "./authentication";
+import PageNotFound from "./PageNotFound";
```
mikatheo, line 32: You have choosen here the nomenclature **OneTwoThree.js** rather than **one-two-three.js**.
It seems we use the second one in our code instead.

mikatheo: In the end, I would like to have an organization in this form: components/PageNotFound/index.js in the commons-ui [seems to be recommended](https://medium.com/@alexmngn/how-to-better-organize-your-react-applications-2fd3ea1920f1). But for the moment, for the sake of uniformity in this repo :)...

geofjamg: > In the end, I would like to have an organization in this form: components/PageNotFound/index.js in the commons-ui [seems to be recommended](https://medium.com/@alexmngn/how-to-better-organize-your-react-applications-2fd3ea1920f1). But for the moment, for the sake of uniformity in this repo :)...

I hate this layout. This is very unusual and a terrible idea. Everything is called index.js. What a mess in your ide. Nobody do that.

PR #129, opened 2020-05-19
improve study card layout

FranckLecuyer:


FranckLecuyer:
> I think it would be better to only display the tooltip in case of truncated name.

enterDelay and enterNextDelay properties added

PR #132, opened 2020-05-25
active power label layout

### LINKED ISSUE #133: Substation label unreadable
When using the "light" theme the substation label on the map becomes unreadable (white font color on white map)

AbdelHedhili:

Add a label next to the lines to indicate the active power. 
Also fix #133


src/components/network/geo-data.js
```diff
@@ -88,4 +91,45 @@ export default class GeoData {
 
         return [ substationPosition1, substationPosition2 ];
     };
+
+    getCoordinateInLine(positions, lineDistance, percent) {
+        if (percent > 100 || percent < 0) {
+            throw Error("percent value incorrect: " + percent);
```
jonenst, line 97: It seems equivalent, but please use the more common ```new Error("...")```

src/components/network/geo-data.js
```diff
@@ -88,4 +91,45 @@ export default class GeoData {
...
+        let reducedAngle = angle;
+        if(angle > 180) {
+            reducedAngle = angle - 180;
+        }
+        const neededOffset = this.getNeededOffset(reducedAngle);
+        return {distance :computeDestinationPoint(goodPolyline[0], leftDistance, angle), angle: angle, offset: neededOffset};
+    }
+
+    getNeededOffset(angle) {
+        if (angle <= 180 && angle > 168) {
```
jonenst, line 123: Should use trigonometry instead to always get the normal ?

src/components/network/geo-data.js
```diff
@@ -88,4 +91,45 @@ export default class GeoData {
...
+
+    getCoordinateInLine(positions, lineDistance, percent) {
+        if (percent > 100 || percent < 0) {
+            throw Error("percent value incorrect: " + percent);
+        }
+        if (lineDistance <= 0) {
+            return null
+        }
+
+        let wantedDistance = lineDistance * percent / 100;
```
jonenst, line 103: I think these distances are already computed somewhere for the arrow layers, maybe we can share the code ?

src/components/network/network-map.js
```diff
@@ -164,6 +164,24 @@ const NetworkMap = forwardRef((props, ref) => {
         }
     }
 
+    function getLabelPosition(network, line, percent, fullPath) {
```
jonenst, line 167: is this used ?

src/components/network/geo-data.js
```diff
@@ -88,4 +90,67 @@ export default class GeoData {
...
+            return null;
+        }
+        let lineDistance = cumulativeDistances[cumulativeDistances.length-1];
+        let wantedDistance = lineDistance * percent / 100;
+        let lowerBound = 0;
+        let upperBound = cumulativeDistances.length - 1;
+        let i = 0;
+
+        //binary search
+        while (true) {
```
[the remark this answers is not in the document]
jonenst: Also please switch to the less surprising "standard" binary search with only one comparision at each step

src/components/network/geo-data.js
```diff
@@ -88,4 +90,67 @@ export default class GeoData {
...
+                segmentDistance = ruler.lineDistance(positions.slice(i, i + 2));
+                cumulativeDistance = cumulativeDistance + segmentDistance;
+                cumulativeDistanceArray[i+1] = cumulativeDistance;
+            }
+            return cumulativeDistanceArray;
+        }
+        return null;
+    }
+
+    getCoordinateInLine(positions, cumulativeDistances, percent) {
```
jonenst, line 111: rename to "labelDisplayPosition" or something like this (it only works for the label because it has the offset)

src/components/network/layers/arrow-layer.js
```diff
@@ -206,43 +206,27 @@ export default class ArrowLayer extends Layer {
...
-                    lineDistancesTextureData.push(lineDistance);
                 });
+                lineDistancesTextureData.push(...line.cumulativeDistances);
             }
             if (linePointCount > MAX_LINE_POINT_COUNT) {
                 throw new Error(`Too many line point count (${linePointCount}), maximum is ${MAX_LINE_POINT_COUNT}`);
             }
             lineAttributes.set(line, {
-                distance: lineDistance,
+                distance: lineDistancesTextureData[lineDistancesTextureData.length - 1],
```
jonenst, line 222: Bug: if the line has 0 points, this will set the distance of the previous line

src/components/network/line-layer.js
```diff
@@ -5,10 +5,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import {CompositeLayer, PathLayer} from 'deck.gl';
+import {CompositeLayer, PathLayer, TextLayer} from 'deck.gl';
 
 import ArrowLayer, {ArrowDirection} from "./layers/arrow-layer";
 import getDistance from "geolib/es/getDistance";
+import cheapRuler from 'cheap-ruler';
```
jonenst, line 12: unused ?

src/components/network/line-layer.js
```diff
@@ -53,18 +54,51 @@ class LineLayer extends CompositeLayer {
...
-                                                         {latitude: positions[1][1], longitude: positions[1][0]});
-
-                        const arrowCount = Math.ceil(lineDistance / DISTANCE_BETWEEN_ARROWS);
-
+                        const directLinePositions = props.geoData.getLinePositions(props.network, line, false);
+                        const directLineDistance = getDistance({latitude: directLinePositions[0][1], longitude: directLinePositions[0][0]},
+                                                         {latitude: directLinePositions[1][1], longitude: directLinePositions[1][0]});
+                        const arrowCount = Math.ceil(directLineDistance / DISTANCE_BETWEEN_ARROWS);
+
+                        ///const positions = props.geoData.getLinePositions(props.network, line, props.lineFullPath);
```
jonenst, line 79: remove

src/components/network/geo-data.js
```diff
@@ -88,4 +90,67 @@ export default class GeoData {
...
+        if (percent > 100 || percent < 0) {
+            throw new Error("percent value incorrect: " + percent);
+        }
+        if (cumulativeDistances === null || cumulativeDistances.length < 2 || cumulativeDistances[1] === 0) {
```
jonenst, line 115: Why the ``` cumulativeDistances[1] === 0``` check ?

src/components/network/geo-data.js
```diff
@@ -88,4 +90,67 @@ export default class GeoData {
...
+        let reducedAngle = angle;
+        if (angle > 180) {
+            reducedAngle = angle - 180;
+        }
+        const neededOffset = this.getNeededOffset(reducedAngle, 10);
+        return {distance :computeDestinationPoint(goodPolyline[0], leftDistance, angle), angle: angle, offset: neededOffset};
+
+    }
+
+    getNeededOffset(angle, offsetDistance) {
```
jonenst, line 152: rename getLabelOffset

src/components/network/line-layer.js
```diff
@@ -53,18 +54,51 @@ class LineLayer extends CompositeLayer {
...
+                    const lineMap = compositeData.lineMap;
                     compositeData.arrows = compositeData.lines.flatMap(line => {
                         // calculate distance between 2 substations as a raw estimate of line size
-                        const positions = props.geoData.getLinePositions(props.network, line, false);
-                        const lineDistance = getDistance({latitude: positions[0][1], longitude: positions[0][0]},
-                                                         {latitude: positions[1][1], longitude: positions[1][0]});
-
-                        const arrowCount = Math.ceil(lineDistance / DISTANCE_BETWEEN_ARROWS);
-
+                        const directLinePositions = props.geoData.getLinePositions(props.network, line, false);
```
jonenst, line 74: Use the real distance now that it's available ?

src/components/network/layers/arrow-layer.js
```diff
@@ -189,10 +189,10 @@ export default class ArrowLayer extends Layer {
     }
 
     createTexturesStructure(props) {
-        const start = performance.now()
+        const start = performance.now();
 
         const linePositionsTextureData = [];
-        const lineDistancesTextureData = [];
+        let lineDistancesTextureData = [];
```
jonenst, line 195: revert

src/components/network/line-layer.js
```diff
@@ -77,11 +111,39 @@ class LineLayer extends CompositeLayer {
 
             this.setState({compositeData: compositeData});
         }
+        //update the label position when toggling fullPath
+        if (changeFlags.propsChanged) {
```
jonenst, line 115: Please change to have only one code that initializes the positions/distances and active power labels. Maybe use (if changeFlags.propsOrDataChanged) or something else ?

src/components/network/network-map.js
AbdelHedhili: It is.

jonenst: where ?

src/components/network/line-layer.js
AbdelHedhili: My dev didn't add access to the real distance, it was obtainable before too. There is no point in recalculating the numbers of arrow in each line every time we change the fullLinePath parameter. This approximation is more than  enough to print a good result.

src/components/network/network-map.js
AbdelHedhili: didn't look at the right function, good remark, deleted.

src/components/network/line-layer.js
AbdelHedhili: I don't see how this would work, I don't need to do the same processing. When the page first load we need to make several initialization, basically populate the compositeData array. However, when the prop fullPath changes i need to update only the label position and offset. The propsOrDataChanged wouldn't help as I would just do inside a if (propsChanged) inside as i need to do that only when one specific prop changed.

src/components/network/geo-data.js
AbdelHedhili: added a comment to explain it : "checks if the line is composed of only one segment with a distance = 0 (a few line are like that in some files)" which kind of means nothing and breaks the algo

jonenst: what if the line has more than 1 segment, but the first 2 points have the same coordinates ?
what if cumulativeDistance has length 1 ?

maybe instead check for cumulativeDistance[length -1] == 0 ?

src/components/network/line-layer.js
jonenst: let's do this in another PR

src/components/network/line-layer.js
jonenst: let's do this in another PR

src/components/network/line-layer.js
jonenst: This code is almost duplicated and could be unified :
```
                compositeData.forEach(compositeData => {
                    let lineMap = new Map();
                    compositeData.lines.forEach(line => {
                        const positions = props.geoData.getLinePositions(props.network, line, props.lineFullPath);
                        const cumulativeDistances = props.geoData.getLineDistances(positions);
                        lineMap.set(line.id, {positions: positions, cumulativeDistances: cumulativeDistances, line: line});
                    });
                    compositeData.lineMap = lineMap;
                });
``` 
and
```
                    compositeData.lineMap.forEach(lineData => {
                        lineData.positions = props.geoData.getLinePositions(props.network, lineData.line, props.lineFullPath);
                        lineData.cumulativeDistances = props.geoData.getLineDistances(lineData.positions);
                    });
```

same for
```
                        let coordinates1 = props.geoData.getCoordinateInLine(lineData.positions, lineData.cumulativeDistances, 15);
                        let coordinates2 = props.geoData.getCoordinateInLine(lineData.positions, lineData.cumulativeDistances, 85);
                        if (coordinates1 !== null && coordinates2 !== null) {
                            compositeData.activePower.push({
                                line: line,
                                p: line.p1,
                                percent: 15,
                                printPosition: [coordinates1.distance.longitude, coordinates1.distance.latitude],
                                offset: coordinates1.offset,
                                });
                            compositeData.activePower.push({
                                line: line,
                                p: line.p2,
                                percent: 85,
                                printPosition: [coordinates2.distance.longitude, coordinates2.distance.latitude],
                                offset: coordinates2.offset});
                        }
                        line.cumulativeDistances = lineData.cumulativeDistances;
                        line.positions = lineData.positions;
```
and
```
                    compositeData.activePower.forEach(activePower => {
                        let lineData = compositeData.lineMap.get(activePower.line.id);
                        const positions = lineData.positions;
                        const cumulativeDistances = lineData.cumulativeDistances;
                        let coordinates = props.geoData.getCoordinateInLine(positions, cumulativeDistances, activePower.percent);
                        if (coordinates) {
                            activePower.printPosition = [coordinates.distance.longitude, coordinates.distance.latitude];
                            activePower.offset = coordinates.offset;
                        }
                    });
                    compositeData.arrows.forEach(arrows => {
                        let lineData = compositeData.lineMap.get(arrows.line.id);
                        arrows.line.positions = lineData.positions;
                        arrows.line.cumulativeDistances = lineData.cumulativeDistances;
                    });
```

it should all go in something like this:
```
if (propsOrDataChanged) {
  if (dataChanged || (propsChanged && oldProps.lineFullPath !== undefined && props.lineFullPath) {
     //the refactored code here
  }
}
```
no ?

src/components/network/geo-data.js
AbdelHedhili: Indeed, changed.

src/components/network/line-layer.js
AbdelHedhili: `compositeData.forEach(compositeData => {
                    let lineMap = new Map();
                    compositeData.lines.forEach(line => {
                        const positions = props.geoData.getLinePositions(props.network, line, props.lineFullPath);
                        const cumulativeDistances = props.geoData.getLineDistances(positions);
                        lineMap.set(line.id, {positions: positions, cumulativeDistances: cumulativeDistances, line: line});
                    });
                    compositeData.lineMap = lineMap;
                });`
here I just initialize the structure

`compositeData.lineMap.forEach(lineData => {
                        lineData.positions = props.geoData.getLinePositions(props.network, lineData.line, props.lineFullPath);
                        lineData.cumulativeDistances = props.geoData.getLineDistances(lineData.positions);
                    });`
and there I update some fields of it.

and same for the 2 others. I don't understand how or why we should try to merge them into 1 code, they don't serve the same purpose and are not called at the same time. I even find it better to have distinct part with clear goal.

AbdelHedhili: Ok so we accept to recreate everything when the prop fullLine changes to have only one code

src/components/network/geo-data.js
```diff
@@ -81,11 +83,77 @@ export default class GeoData {
...
+        }
+        return null;
+    }
+
+    /**
+     * Find the polyline in which we reach the wanted distance and return the polyline
+     * along with the distance left to travel on this polyline to be at the exact wanted distance
+     * (implemented using a binary search)
+     */
+    findPolyline(positions, cumulativeDistances, wantedDistance) {
```
jonenst, line 115: Rename to findSegment (a polyline is typically more than 1 segment) ?
Also rename "leftDistance" to "remainingDistance" ?

src/components/network/layers/arrow-layer.js
```diff
@@ -206,43 +206,29 @@ export default class ArrowLayer extends Layer {
...
             if (linePointCount > MAX_LINE_POINT_COUNT) {
                 throw new Error(`Too many line point count (${linePointCount}), maximum is ${MAX_LINE_POINT_COUNT}`);
             }
-            lineAttributes.set(line, {
-                distance: lineDistance,
-                positionsTextureOffset: linePositionsTextureOffset,
-                distancesTextureOffset: lineDistancesTextureOffset,
-                pointCount: linePointCount
-            });
+            if (positions.length > 0) {
```
jonenst, line 221: Is it correct to not set the attributes at all ? Before this change it was set. I think It's easier to get right if we keep the previous behavior.

jonenst:
Merged, thanks !

PR #134, opened 2020-06-03
switch update error feedback

FranckLecuyer:


src/components/single-line-diagram.js
```diff
@@ -71,17 +78,19 @@ const SingleLineDiagram = forwardRef((props, ref)  => {
...
                 svgPrevViewbox.current = svgDraw.current.viewbox();
             }
             updateState(s=>!s)
     }, []);
 
-
     useImperativeHandle(ref, () => ({
-        reloadSvg: forceUpdate
+        reloadSvg: forceUpdate,
+        updateSwitchMsg: (m) => setUpdateSwitchMsg(m)
```
jonenst, line 93: unecessary use of lambda ?

jonenst: Wouldn't it be better to pass the message in a prop ? We are using an imperativeHandle for the refresh because we want it to redownload the exact same url and props don't allow that. But for the message, we do things only when the message changes, so props are more straightforward

src/components/single-line-diagram.js
```diff
@@ -155,7 +164,7 @@ const SingleLineDiagram = forwardRef((props, ref)  => {
             });
             svgDraw.current = draw;
         }
-    }, [svg]);
+    }, [svg, updateSwitchMsg]);
```
jonenst, line 167: I don't understand why we need to put updateSwitchMsg in the dependencies here

src/components/study-pane.js
```diff
@@ -178,15 +181,22 @@ const StudyPane = () => {
...
+        eltClose.style.visibility = open ? "hidden" : "visible";
 
         updateSwitchState(studyName, breakerId, open).then( response => {
             if (response.ok) {
                 sldRef.current.reloadSvg();
+                sldRef.current.updateSwitchMsg("");
             }
             else {
                 console.error(response);
+                eltOpen.style.visibility = open ? "hidden" : "visible";
```
jonenst, line 197: Is this really necessary ? or shouldn't we just redownload the svg ?

src/components/single-line-diagram.js
FranckLecuyer: At first, i didn't add this dependancy, but i got strange behaviour when navigating between voltage level after an error occurred : svg not properly zoomed for example.
So i added this dependancy and it works well now ...

src/components/study-pane.js
FranckLecuyer: I preferred reverting the switch modification this way : it's easy and it doesn't add time overhead by reloading the entiere svg

src/components/single-line-diagram.js
FranckLecuyer: You're right : using a prop here makes sense. The modification is done

src/components/single-line-diagram.js
FranckLecuyer: Modification done

jonenst:
Maybe we can rework the system to preserve the viewbox in another PR, make save the current viewbox not before starting the download of the svg, but after the new svg has been downloaded and right before replacing the new svg. In this case, setting the prevViewBox to null is maybe not necessary anymore ?

PR #136, opened 2020-06-08
represent disconnected line by dash line on map

JacquesBorsenberger:
if line is disconnected at both end, display line in black or white (depending on theme)


package.json
```diff
@@ -32,7 +32,8 @@
     "react-virtualized": "^9.21.2",
     "react-window": "^1.8.5",
     "redux": "^4.0.5",
-    "typeface-roboto": "0.0.75"
+    "typeface-roboto": "0.0.75",
+    "@deck.gl/extensions": "latest"
```
jonenst, line 36: please use a fixed version, and also commit the resulting package-lock.json

jonenst: Looks like we already have @deck.gl/extensions@8.0.17 in master, do you need this ?

src/components/network/line-layer.js
```diff
@@ -81,24 +81,32 @@ class LineLayer extends CompositeLayer {
...
+                },
                 getWidth: 2,
                 visible: this.props.filteredNominalVoltages.includes(compositeData.nominalVoltage),
                 updateTriggers: {
-                    getPath: [this.props.lineFullPath]
-                }
+                    getPath: [this.props.lineFullPath],
+                    getColor: [this.props.disconnectedLineColor]
+                },
+                getDashArray: line => doDash(line) ? [15, 10] : [0,0],
```
jonenst, line 108: use the same 2 array objects instead of a new one for each line

src/components/network/line-layer.js
```diff
@@ -81,24 +81,32 @@ class LineLayer extends CompositeLayer {
 
     renderLayers() {
         const layers = [];
-
         // lines : create one layer per nominal voltage, starting from higher to lower nominal voltage
         this.state.compositeData.forEach(compositeData => {
             const color = this.props.getNominalVoltageColor(compositeData.nominalVoltage);
 
+            const doDash = line => {
```
jonenst, line 88: can you make it symmetric between the "line.p1 == null || line.p2 == null" and the "line.p1==null && line.p2 == null" ?
Either make 2 functions doDash and isDisconnected or inline both in their respective callsite.
Just for symmetry and aesthetics

src/components/network/network-map.js
```diff
@@ -187,6 +187,7 @@ const NetworkMap = forwardRef((props, ref) => {
             network: props.network,
             geoData: props.geoData,
             getNominalVoltageColor: getNominalVoltageColor,
+            disconnectedLineColor: labelColor,
```
jonenst, line 190: can you either rename the labelColor variable ( "foregroundNeutralColor" ? )
or create a separate variable disconnectedLineColor (with a separate computation, even if it's the same value as the the labelColor ) ?

src/components/network/line-layer.js
```diff
@@ -81,24 +81,37 @@ class LineLayer extends CompositeLayer {
...
+                return line.p1 == null || line.p2 == null;
+            };
+            const isDisconnected = line => {
+                return line.p1 == null && line.p2 == null
+            };
+            const noDashArray=[0,0];
```
geofjamg, line 94: You should declare it outside of render method

jonenst: Javascript why u so tricky >_>

PR #137, opened 2020-06-08
Notifs

jonenst:


jonenst:
first version without reconnects and without filters: always reload the map and the svg.

PR #145, opened 2020-06-18
Parallel path

JacquesBorsenberger:


jonenst:
The asymmetry of the fork and the arrows flying next to the lines are fixed by computing the angles with the same hack as for the power labels (using mercator projection angle computations instead of doing atan2(delta_lat, delta_lon)
Capture d’écran de 2020-07-29 19-19-50
Capture d’écran de 2020-07-29 19-18-35

PR #148, opened 2020-06-22
export network

AbdelHedhili:


src/utils/dialogs.js
```diff
@@ -119,6 +126,81 @@ const RenameDialog = ({ open, onClose, onClick, title, message, currentName }) =
...
 
+/**
+ * Dialog to export the network case #TODO To be moved in the common-ui repository once it has been created
+ * @param {Boolean} open Is the dialog open ?
+ * @param {EventListener} onClose Event to close the dialog
+ * @param {EventListener} onClick Event to submit the export
+ * @param {String} title Title of the dialog
+ * @param {String} message Message of the dialog
+ */
+const ExportDialog = ({ open, onClose, onClick, title, message, availableFormat}) => {
```
geofjamg, line 137: I would move it to its own file

AbdelHedhili: why this one ? it's not bigger than the others dialog and the file is still small.

jonenst:
doesn't work in firefox 68 ?
I need to add the <a> to the body for it to work:
```diff
+        document.body.appendChild(a)
         a.click();
+        document.body.removeChild(a)
```

According to https://stackoverflow.com/a/1667520/2748664

Another problem is that it closes the dev websocket so hot reloading doesn't work anymore after exporting a file


Maybe it's best to just do:
```
     const handleClickExport = (format, url) => {
-        let a = document.createElement('a');
-        a.href = url;
-        a.click();
+        window.open(url, '_blank');
         handleCloseExport();
     };
```
It's a bit ugly but at least it works

PR #150, opened 2020-06-23
Prepare authentication mutialization

ChamseddineBhd:


jonenst:
One problem with this is that UserManagerMock is no longer pruned by webpack when building for production. We should try to go back to how it was before and not have test code deployed to production.

PR #160, opened 2020-07-08
Automatic code formatting

ChamseddineBhd:


geofjamg:
Formatter seems to remove end semi colons. Is it a general accepted good practice? Intellij always tells me the contrary.

ChamseddineBhd:
> Formatter seems to remove end semi colons. Is it a general accepted good practice? Intellij always tells me the contrary.

it's better to keep semicolons.
i have fixed the code style.

PR #161, opened 2020-07-15
Escape query string parameters in rest-api.js

etiennehomer:
Signed-off-by: HOMER Etienne <etiennehomer@gmail.com>


etiennehomer:
I couldn't  construct URLs object and set their fields. We can have a look to this option when you're back.

jonenst:
Does it work with a study named "A/B" or "foo?" ? I think this can be simplified and improved

etiennehomer:
No it's not working... I'll have a look to this

PR #164, opened 2020-07-27
Encode and decode parameters to allow string escaping

etiennehomer:
Signed-off-by: HOMER Etienne <etiennehomer@gmail.com>

### DIFF
--- a/src/components/study-pane.js
+++ b/src/components/study-pane.js
@@ -147,7 +147,7 @@ const StudyPane = () => {
     }, [network]);
 
     function loadNetwork(studyName) {
-        console.info(`Loading network of study '${studyName}'...`);
+        console.info(`Loading network of study '${decodeURIComponent(studyName)}'...`);
 
         const substations = fetchSubstations(studyName);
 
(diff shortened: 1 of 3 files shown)


src/components/study-pane.js
```diff
@@ -149,9 +149,9 @@ const StudyPane = () => {
     function loadNetwork(studyName) {
         console.info(`Loading network of study '${studyName}'...`);
 
-        const substations = fetchSubstations(studyName);
+        const substations = fetchSubstations(decodeURIComponent(studyName));
```
jonenst, line 152: Should not use decodeURIComponent here and other places in this file

src/components/study-pane.js
```diff
@@ -70,7 +70,7 @@ const useStyles = makeStyles((theme) => ({
 const INITIAL_POSITION = [0, 0];
 
 const StudyPane = () => {
-    const { studyName } = useParams();
+    const studyName = decodeURIComponent(useParams()['studyName']);
```
jonenst, line 73: use .studyName instead of ['studyName'] for coding style

PR #172, opened 2020-08-19
Change cursor style ('pointer') when mouse is on lines or substations

FranckLecuyer:


src/components/network/network-map.js
```diff
@@ -227,6 +229,10 @@ const NetworkMap = forwardRef((props, ref) => {
         }
     }
 
+    function cursorHandler({}) {
```
jonenst, line 232: remove '{}' ?

FranckLecuyer: Corrected

AbdelHedhili (approved):

jonenst (changes requested):
With this change we lose the change of mouse pointer when panning the map
grabbing

Need to do something like this (from the source code of deckgl):

```javascript
//default prop getCursor
const getCursor = ({isDragging}) => (isDragging ? 'grabbing' : 'grab');
```

jonenst:
Also we should make the clicking on the substations and on the lines work before merging this

PR #179, opened 2020-08-25
add private/public option in study creation

AbdelHedhili:


package.json
```diff
@@ -32,7 +32,8 @@
     "react-window": "^1.8.5",
     "reconnecting-websocket": "^4.4.0",
     "redux": "^4.0.5",
-    "typeface-roboto": "0.0.75"
+    "typeface-roboto": "0.0.75",
+    "jwt-decode": "latest"
```
jonenst, line 36: unused ?

src/utils/rest-api.js
```diff
@@ -33,7 +33,7 @@ export function fetchStudies() {
     console.info('Fetching studies...');
     const fetchStudiesUrl = PREFIX_STUDY_QUERIES + '/v1/studies';
     console.debug(fetchStudiesUrl);
-    return backendFetch(fetchStudiesUrl).then((response) => response.json());
+    return backendFetch(fetchStudiesUrl).then((response) => response.json()).then(value => value.filter((current,index,array)=>array.findIndex(t=>(t.studyName === current.studyName))===index));
```
jonenst, line 36: this should be dedupped in the backend

jonenst (approved):

src/components/app.js
```diff
@@ -111,8 +111,8 @@ const App = () => {
             });
     }, []);
 
-    function studyClickHandler(studyName) {
-        history.push('/studies/' + encodeURIComponent(studyName));
+    function studyClickHandler(studyName, userId) {
+        history.push('/studies/' + encodeURIComponent(studyName) + '/' + encodeURIComponent(userId.subject) + '/' + encodeURIComponent(userId.issuer));
```
jonenst, line 115: Not sure what to do here:
- keep a short form url with only the studyname in urls and assume the currently logged in user/issuer; in addition to having a longer form with the user/issuer that's shareable with other users
- always use the longer form

Anyway I think the url should be /studies/issuer/user/studyname

geofjamg: This seems to me an incorrect REST api design: studies/{studyName} is unseccable. 
I would use query parameters as there is no hierarchy beetwen issuer, user and studies
/studies/{studyName}?issuer=,user=

AbdelHedhili: I don't feel like it's a good idea either to use query parameters as the base url should refer to a specific resource and it's not the case if we do : studies/{studyName}

geofjamg: Why? 
studies/studyName is a resource

AbdelHedhili: but you cannot make a link between a studyName and a study anymore. The issuer and subject are mandatory to get a study and therefore should be a path variable IMO.

geofjamg:
Issuers and user are meta infos even if mandatory but not a resource. What you want to get is a study not a user or anything else.

src/components/create-study-form.js
```diff
@@ -161,6 +163,7 @@ export const CreateStudyForm = () => {
 
     const [studyName, setStudyName] = React.useState('');
     const [studyDescription, setStudyDescription] = React.useState('');
+    const [studyPrivacy, setStudyPrivacy] = React.useState('public');
```
jonenst, line 166: private by default ?

src/components/study-manager.js
```diff
@@ -152,22 +152,24 @@ const StudyCard = ({ study, onClick }) => {
...
         setOpenDelete(true);
     };
 
     const handleClickDelete = () => {
-        deleteStudy(study.studyName).then(() => {
+        deleteStudy(study.studyName, study.userId).then((response) => { response.ok ?
             fetchStudies().then((studies) => {
                 dispatch(loadStudiesSuccess(studies));
-            });
+            }) : setDeleteError(true);
```
jonenst, line 166: pass the error message here

src/redux/reducer.js
```diff
@@ -77,11 +78,13 @@ export const reducer = createReducer(initialState, {
     },
 
     [OPEN_STUDY]: (state, action) => {
-        state.studyName = action.studyName;
+        state.studyName = action[0];
```
jonenst, line 81: this works ? shouldn't it be action.studyname[0]?

src/redux/actions.js
```diff
@@ -39,8 +39,8 @@ export function loadCasesSuccess(cases) {
 
 export const OPEN_STUDY = 'OPEN_STUDY';
 
-export function openStudy(studyName) {
-    return { type: OPEN_STUDY, studyName: studyName };
+export function openStudy(studyName, userId) {
+    return { type: OPEN_STUDY, studyName: [studyName, userId] };
```
jonenst, line 43: rename "studyName:"  in the action to studyRef or something ?

src/translations/en.json
```diff
@@ -18,6 +20,7 @@
   "deleteStudyMsg": "Are you sure you want to delete the study?",
   "studyNameDidNotMatchMsg": "The given study name did not match with ",
   "studyDeletedSuccessMsg": "Study deleted",
+  "deleteStudyError": "Cannot remove the study if you are not the owner",
```
jonenst, line 23: There can be other sources of errors when deleting a study, but maybe we can do this in a later PR

src/components/create-study-form.js
```diff
@@ -206,13 +209,13 @@ export const CreateStudyForm = () => {
 
         clearTimeout(timer.current);
         timer.current = setTimeout(() => {
-            updateStudyFormState(name);
+            updateStudyFormState(name, store.getState().user.profile.sub);
```
jonenst, line 212: use useSelector instead of accessing the store globally like this. Actually this export should be removed maybe??

jonenst (approved):

PR #181, opened 2020-08-31
add an iframe to download grid

JacquesBorsenberger:
window.location.href = url;
 works but break if an error occurs

### DIFF
--- a/src/components/study-manager.js
+++ b/src/components/study-manager.js
@@ -88,6 +88,7 @@ const useStyles = makeStyles((theme) => ({
     },
 }));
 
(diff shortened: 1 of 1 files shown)


jonenst:
The problem we are also trying to solve is that window.open("...", "_blank") opens new tab and immediately closes it, causing some visual glitch. Is this fixed ?
But yes as you say, having it resistant to server errors (a crash, or even just a missing "Content-disposition: attachment" header) is also a good thing.

JacquesBorsenberger:
yes the visual glitch is fixed

jonenst (approved):

PR #193, opened 2020-09-15
show overloaded lines on network map

FranckLecuyer:


geofjamg (approved):

src/redux/local-storage.js
```diff
@@ -24,6 +24,11 @@ export const getLocalStorageBoolean = (key, defValue) => {
     return value === 'true';
 };
 
+export const getLocalStorageInteger = (key, defValue) => {
```
jonenst, line 27: remove this function ?
- It's called getInteger but doesn't do any integer handling
- no need to define a function doing just a || b

FranckLecuyer: Done

PR #196, opened 2020-09-17
Choose voltage level to display on substation click on the map

FranckLecuyer:


src/components/voltage_level_choice.js
```diff
@@ -0,0 +1,122 @@
+import React from 'react';
```
geofjamg, line 1: Could yoi add license header?

FranckLecuyer: Done

geofjamg (approved):

PR #205, opened 2020-09-22
Add line overload view

JacquesBorsenberger:


geofjamg (changes requested):
You should virtualize your table. On a large network, and using a very low threshold you have many constraints and switch from nominal voltage mode to overload mode take several seconds because of table rendering.
Also there is a strange glitch, when you scroll the table with the mouse, the table seems to blink.

geofjamg:
image
Maybe when there is few constraints, we should avoid displaying the scrollbar?
The first column with equipment ID is not readable enough, maybe we can have a text background (just behind the text) a little bit light (or dark depending on the theme).
@mikatheo what do you think about this?

JacquesBorsenberger:
table is virtualized and scrollbar is only present if needed.

src/redux/local-storage.js
```diff
@@ -24,6 +24,11 @@ export const getLocalStorageBoolean = (key, defValue) => {
     return value === 'true';
 };
 
+export const getLocalStorageInteger = (key, defValue) => {
```
geofjamg, line 27: See jon comment on usefulness of this function 
https://github.com/gridsuite/gridstudy-app/pull/193#discussion_r490432841

src/components/network/overloadedLinesView.js
```diff
@@ -0,0 +1,178 @@
...
+                        width: 70,
+                    },
+                    {
+                        label: intl.formatMessage({ id: 'Intensity' }),
+                        dataKey: 'intensity',
+                        numeric: true,
+                        width: 70,
+                    },
+                    {
+                        label: intl.formatMessage({ id: 'Limit' }),
```
geofjamg, line 147: There is no translation for Limit with an upper case

geofjamg (changes requested):

src/components/network/overloadedLinesView.js
JacquesBorsenberger: translation added

geofjamg (approved):

PR #211, opened 2020-10-01
Add loadflow parameters 

JacquesBorsenberger:
need some review for the parameters labels


geofjamg:
> need some review for the parameters labels

Done

geofjamg (changes requested):
Reset parameters to default does not work as it is supposed to reinit all values fro mserver using values of LoadFlowParameters default constructor.

geofjamg (approved):

PR #213, opened 2020-10-02
lockout lines

AbdelHedhili:


geofjamg (changes requested):
Could you move out in another PR the navigation part (the arrow instead of label new navigation style). Also could you see with @mikatheo about the arrow design. UI designers have worked on it and there is a sketch for that.

package.json
```diff
@@ -31,7 +31,8 @@
         "react-window": "^1.8.5",
         "reconnecting-websocket": "^4.4.0",
         "redux": "^4.0.5",
-        "typeface-roboto": "0.0.75"
+        "typeface-roboto": "0.0.75",
+        "color-name": "latest"
```
jonenst, line 35: I don't think this is used. And even if you do use it, I don't think we need this dependency at all

jonenst (changes requested):
At least remove color-name and merge with master (addNavigationArrow already merged)

Haven't looked in details but I think the code can be simplified, to investigate when we resume work on this

PR #225, opened 2020-10-16
Sort and filter tables

JacquesBorsenberger:


geofjamg (changes requested):
When we first display a table, it is not yet sorted, we have to click to either sort ascending ot descending, it is disturbing. 
For me we should have one of the following behaviour:
  - either we have 3 sorting states (not sorted, ascending, and descending) and by default it is not sorted.
 - either we only have 2 sorting states (ascending and descending) and it has to be sorted by default (ascending)

geofjamg (approved):

PR #226, opened 2020-10-16
edit access rights of study after creation

AbdelHedhili:


src/utils/rest-api.js
```diff
@@ -323,6 +323,27 @@ export function renameStudy(studyName, userId, newStudyName) {
     }).then((response) => response.json());
 }
 
+export function changeStudyAccessRights(studyName, userId, toPrivate) {
+    console.info('Change access rights of study ' + studyName);
+    const changeStudyAccessRightsUrl =
+        PREFIX_STUDY_QUERIES +
```
geofjamg, line 329: Thre is a new getStudyUrl() to share this little piece of code. Could you use it?

src/utils/dialogs.js
```diff
@@ -297,4 +307,136 @@ ExportDialog.propTypes = {
...
+/**
+ * Dialog to change the access rights of a study #TODO To be moved in the common-ui repository once it has been created
+ * @param {Boolean} open Is the dialog open ?
+ * @param {EventListener} onClose Event to close the dialog
+ * @param studyName the name of the study to export
+ * @param userId the name of the logged in user
+ * @param {String} title Title of the dialog
+ * @param {String} isPrivate tells if the study is private or not
+ */
+const AccessRightsDialog = ({
```
geofjamg, line 319: dialogs.js is not dedicated to receive all dialogs of the app. Is was just supposed to get very generic and reusable dialogs like "rename". Could you create a access-rights-dialog.js for that?

geofjamg (changes requested):
See comments

geofjamg (changes requested):
- public to private works
- private to public doesn't work => to fix
- are we allowed to switch public to private when we don't own the study? I think we should not.
- when we switch private to study we should fire an study list update event sot that all other navigators update (here you do it synchronously which was the old way to do it)

AbdelHedhili:
>     * are we allowed to switch public to private when we don't own the study? I think we should not.
> 
>     * when we switch private to study we should fire an study list update event sot that all other navigators update (here you do it synchronously which was the old way to do it)
 
I confirm we can't change the rights on a study we don't own.

src/components/access-rights-dialog.js
```diff
@@ -0,0 +1,153 @@
...
+};
+
+AccessRightsDialog.propTypes = {
+    open: PropTypes.bool.isRequired,
+    onClose: PropTypes.func.isRequired,
+    studyName: PropTypes.string.isRequired,
+    userId: PropTypes.string.isRequired,
+    title: PropTypes.string.isRequired,
+    isPrivate: PropTypes.string.isRequired,
+};
```
SlimaneAmar, line 151: Why don't we use the plain object study as a property ?

SlimaneAmar (approved):

geofjamg (approved):

PR #227, opened 2020-10-19
[WIP] Delete a study being created

SlimaneAmar:
Signed-off-by: Slimane AMAR <slimane.amar@rte-france.com>


src/components/loader-with-overlay.js
```diff
@@ -33,14 +31,7 @@ const LoaderWithOverlay = ({
...
             <CircularProgress color={color} size={loaderSize} />
-            <div
-                className={classes.message}
-                style={{
-                    fontSize: loadingMessageSize,
-                }}
-            >
-                <FormattedMessage id={loadingMessageText} />
-            </div>
+            <FormattedMessage id={loadingMessageText} />
```
med-hedi, line 34: Why did you delete all this? You are sure that this modification will not impact the loader displayed before the map is loaded?

src/components/study-manager.js
```diff
@@ -474,23 +486,24 @@ const StudyManager = ({ onClick }) => {
...
-                        <StudyCard
-                            studyCreationLoader={false}
-                            study={study}
-                            onClick={() =>
-                                onClick(study.studyName, study.userId)
-                            }
-                        />
-                    </Grid>
-                ))}
+                {studies instanceof Array &&
```
geofjamg, line 489: In which case studies is not an array even if empty?

SlimaneAmar: I will do that

src/components/loader-with-overlay.js
SlimaneAmar: I just remove the div element to have a column type alignment for the circular progress element and the message element

src/components/study-manager.js
SlimaneAmar: Done

geofjamg (approved):

PR #231, opened 2020-10-20
Fix bug rendering network with voltage filtering after AS

SlimaneAmar:


geofjamg (changes requested):
I would prefer NetworkMap component to remain as much independent as possible from redux store so that one day we can move it to commons and share it with other fronts/projects.
Could you fix the issue but keep network, geodata and nominal voltage filter passed as a props?

SlimaneAmar:
useName is also in the redux store

geofjamg (approved):

PR #235, opened 2020-10-22
Move the calculation model button to the bottom right and with a new design

FranckLecuyer:


geofjamg:
I would move it at the really bottom of the map under the "Open street map ..."
@mikatheo I don't kown if it is important but in the picture there is an alternate change of background in the model list popup (2 types of grey)
In the spec there is an inconsistency between picture 1 and 2. When loadflow in not yet run, is it supposed to be light grey with dark text (picture 1) or the inverse (picture 2)?

mikatheo:
> I would move it at the really bottom of the map under the "Open street map ..."
> @mikatheo I don't kown if it is important but in the picture there is an alternate change of background in the model list popup (2 types of grey)
> In the spec there is an inconsistency between picture 1 and 2. When loadflow in not yet run, is it supposed to be light grey with dark text (picture 1) or the inverse (picture 2)?



> I would move it at the really bottom of the map under the "Open street map ..."

This is unfortunately not allowed:
https://docs.mapbox.com/help/how-mapbox-works/attribution/

> @mikatheo I don't kown if it is important but in the picture there is an alternate change of background in the model list popup (2 types of grey)

Not require

> In the spec there is an inconsistency between picture 1 and 2. When loadflow in not yet run, is it supposed to be light grey with dark text (picture 1) or the inverse (picture 2)?

The 2nd image is to be taken into account, with a black with an Alpha = 50% as mentioned in the acceptance criteria.

mikatheo:
It seems to be possible to move the allocations to another corner (bottom left?). Then you can move closer to the bottom edge (not glued), but of course you will have to do the same for the list of filters:

https://docs.mapbox.com/help/how-mapbox-works/attribution/ =>
> By default, the Mapbox wordmark and information button are located on the bottom left of the map. You may move these elements to a different position, but they must stay on the map view. For display options, see the API documentation for the UiSettings class. You may also reposition these elements within the activity’s XML layout.

mikatheo (approved):

geofjamg (approved):

PR #237, opened 2020-10-23
Add substation to voltagelevel list

JacquesBorsenberger:


FranckLecuyer (commented):
We can't just center a substation on a map anymore, without opening it at the same time
Is this what we want ?

geofjamg (changes requested):
We lost the voltage level menu which was not the intend of this feature:
image

JacquesBorsenberger:
> We can't just center a substation on a map anymore, without opening it at the same time
> Is this what we want ?

no, fixed it ;)

JacquesBorsenberger:
> We lost the voltage level menu which was not the intend of this feature:

Requested by @mikatheo in US
> [Michael - 23/10] Suppression des trois points qui n'ont plus de sens pour le moment, et alourdisse l'IHM

geofjamg:
> > We lost the voltage level menu which was not the intend of this feature:
> 
> Requested by @mikatheo in US
> 
> > [Michael - 23/10] Suppression des trois points qui n'ont plus de sens pour le moment, et alourdisse l'IHM

Ok

geofjamg (dismissed):
The app has now a scroll bar. It was and is supposed to just fit the screen:
image
Could you fix it?

mikatheo:
> The app has now a scroll bar. It was and is supposed to just fit the screen:
> image
> Could you fix it?

Fixed!

mikatheo:
> The app has now a scroll bar. It was and is supposed to just fit the screen:
> image
> Could you fix it?

Fixed (adaptation of the auto-sizer height following the removal of the Model button)

jonenst (approved):

PR #246, opened 2020-10-28
Add full screen single line diagram

med-hedi:


FranckLecuyer (changes requested):
See comments sent in Slack ...

geofjamg (changes requested):
Once maximized the visible part of the diagram is not the one of the window, see cut dagram:
image
- once maximized the app is not full screen anymore and a scroll bar is displayed:
image

med-hedi:
@geofjamg for the scroll it would be fixed in the US didier to correct the problem with scroll

geofjamg:
> @geofjamg for the scroll it would be fixed in the US didier to correct the problem with scroll

That would be a regression as there is no problem in the actual code. I do not any reason to fix it with a correct hard-coded offset as it is done today.
Anyway as you wish but this PR won't be merge with such a bug.

med-hedi:
@geofjamg @FranckLecuyer all your comments have been fixed

geofjamg (changes requested):
- Is it on purpose that you added a white border:
image
- scroll bar issue is not fixed
- panning limits are not correct when maximized

geofjamg (changes requested):
There is a scroll issue when the diagram is maximized.

med-hedi:
> There is a scroll issue when the diagram is maximized.

I fixed the scroll problem, it is only reproducible on firefox which is why I did not see it in the commit before because I tested on chrome

geofjamg (approved):

geofjamg (changes requested):
There is a regression: we loose the pan position when running a loadflow.

geofjamg:
> There is a regression: we loose the pan position when running a loadflow.

Ok this is not a regression because of this PR

geofjamg (approved):

PR #248, opened 2020-10-30
Add highlight voltage level

med-hedi:

### DIFF
--- a/src/components/network/network-explorer.js
+++ b/src/components/network/network-explorer.js
@@ -82,6 +82,8 @@ const NetworkExplorer = ({
 
     const classes = useStyles();
 
+    const [selectedIndex, setSelectedIndex] = React.useState(0);
+
     const [filteredVoltageLevels, setFilteredVoltageLevels] = React.useState(
         []
     );
--- a/src/components/network/network-explorer.js
+++ b/src/components/network/network-explorer.js
@@ -8,7 +8,7 @@
 import React, { useCallback, useEffect } from 'react';
 import PropTypes from 'prop-types';
 
-import { useSelector } from 'react-redux';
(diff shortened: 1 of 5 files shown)


geofjamg (changes requested):
- the chosen selection color is not different enough and so on not visible.
image
  Could you use a color close to the sketch?
- In light this is even worst and there is also a glitch see the small white band on top of the selection:
image
- the substation selection is not done.
- do we try the triangle like in the sketch. it would probably help to improve the readability of the selection (in the current state I'am not totally convinced)? 
image

geofjamg:
@mikatheo could you look at this?

med-hedi:
@mikatheo before starting the US we agreed not to do the triangle, finally we do it?

med-hedi:
@geofjamg ok, i will correct that

mikatheo:
> @mikatheo before starting the US we agreed not to do the triangle, finally we do it?

We did indeed delay for the triangle last week, which may be the subject of another PR after discussion with the designers.
However, I agree with the other remarks, I just submitted them to Eline for more details. I will transfer the more precise elements to you very soon.

geofjamg:
> > @mikatheo before starting the US we agreed not to do the triangle, finally we do it?
> 
> We did indeed delay for the triangle last week, which may be the subject of another PR after discussion with the designers.
> However, I agree with the other remarks, I just submitted them to Eline for more details. I will transfer the more precise elements to you very soon.

Ok

med-hedi:
@geofjamg I made the necessary modifications and I validated the color codes with @ChristineHesse before pushing, for this PR we still have the triangle to do that will be the objective of another US.

geofjamg (approved):

PR #253, opened 2020-11-03
move_mapbox_attributions_to_bottom_left

FranckLecuyer:


geofjamg (changes requested):
This looks strange:
image
Maybe just at the right of mapbox logo ?

jonenst:
This is not compatible with the new drawer which would hide the license

PR #254, opened 2020-11-03
add navigation arrows

AbdelHedhili:


geofjamg (changes requested):
- Arrows seem to be a little bit too large compare to original sketch
image
 - It should all be aligned. It looks like very strange in the previous picture.
 - I don't think the way it is implemented is correct (see in code comments)

src/components/single-line-diagram.js
```diff
@@ -183,7 +187,117 @@ const SingleLineDiagram = forwardRef((props, ref) => {
...
     useLayoutEffect(() => {
+        function addNavigationArrowDef() {
+            let svg = document.getElementById('sld-svg').children[0]; //Get svg element
+            let defElement = document.createElementNS(
+                'http://www.w3.org/2000/svg',
+                'defs'
+            );
+
+            let html;
+            if (theme === LIGHT_THEME) {
```
geofjamg, line 198: This is not a correct way to use theme. Everything should be contained in the theme to be able to add another one without changing the code. 
Why did you duplicate the html code in the 2 parts of the if? Why just not using a color from the theme to fill the jtml code?

src/components/single-line-diagram.js
```diff
@@ -183,7 +187,117 @@ const SingleLineDiagram = forwardRef((props, ref) => {
...
+                    '<polygon style="stroke: #212121;fill: #212121" points="0 0, 50 80, 100 0"/>',
+                    '</marker>',
+                ].join('');
+            }
+
+            defElement.innerHTML = html;
+            svg.appendChild(defElement);
+        }
+
+        function createSvgArrow(element, transform, position) {
```
geofjamg, line 216: Why do you generate the svg in js? Why not an external SVG resource that will be added to decorate the original one? This is a lot of boilerplate code that we don't want to see.

geofjamg (changes requested):

geofjamg (approved):

PR #258, opened 2020-11-12
Add offset to fork start

JacquesBorsenberger:

# Conflicts:
#	src/components/network/network-explorer.js
#	src/components/study-pane.js


jonenst:
Capture d’écran de 2020-11-12 15-22-22
on microbe I get this for the substation that has only 1 vl

src/components/network/layers/fork-line-layer.js
```diff
@@ -32,20 +34,32 @@ export default class ForkLineLayer extends LineLayer {
...
 target = source ;
 if( abs(instanceLineParallelIndex) != 9999. ) {
     float offsetPixels = clamp(project_size_to_pixel( distanceBetweenLines), minParallelOffset, maxParallelOffset );
     float offsetCommonSpace = project_pixel_size(offsetPixels);
-    vec4 trans = vec4(cos(instanceLineAngle), -sin(instanceLineAngle ), 0, 0.) * instanceLineParallelIndex ;
+
+    float offsetSubstation = clamp(project_size_to_pixel(substationRadius*instanceOffsetStart ), 
+                                    1., 
+                                    substationMaxPixel * instanceOffsetStart );
+    float offsetSubstationCS = project_pixel_size(offsetSubstation) ;
```
jonenst, line 53: It took me time to understand that CS means CommonSpace, I think it's best to use the explicit form ? At lease the variables OffsetCommonSpace and OffsetSubstationsCS should use the same naming convention

src/components/network/layers/fork-line-layer.js
```diff
@@ -32,20 +34,32 @@ export default class ForkLineLayer extends LineLayer {
...
             `,
             'float segmentIndex = positions.x': `;
 target = source ;
 if( abs(instanceLineParallelIndex) != 9999. ) {
     float offsetPixels = clamp(project_size_to_pixel( distanceBetweenLines), minParallelOffset, maxParallelOffset );
     float offsetCommonSpace = project_pixel_size(offsetPixels);
-    vec4 trans = vec4(cos(instanceLineAngle), -sin(instanceLineAngle ), 0, 0.) * instanceLineParallelIndex ;
+
+    float offsetSubstation = clamp(project_size_to_pixel(substationRadius*instanceOffsetStart ), 
+                                    1.,
```
jonenst, line 51: why a min of 1. instead of 0. ?

src/components/network/line-layer.js
```diff
@@ -171,6 +172,23 @@ class LineLayer extends CompositeLayer {
...
+        return (
+            [
+                ...new Set(
+                    substation.voltageLevels.map((vl) => vl.nominalVoltage) // only one voltage level
+                ),
+            ]
+                .sort((a, b) => {
+                    return a - b; // force numerical sort
+                })
+                .indexOf(vl.nominalVoltage) + 0.1
```
jonenst, line 188: Add a comment to say that this means 10% of the radius. and why not 50% to be in the middle of the width of the ring of the voltage level ?

src/components/network/line-layer.js
```diff
@@ -428,6 +406,54 @@ class LineLayer extends CompositeLayer {
...
     }
 
+    recomputeParallelLinesIndex(compositeData, props) {
+        compositeData.forEach((compositeData) => {
+            const mapOriginDestination = compositeData.mapOriginDestination;
+            // calculate index for line with same subsation set
+            // The index is a real number in a normalized unit.
+            // +1 => distanceBetweenLines on side
+            // -1 => distanceBetweenLines on the other side
+            // 0.5 => half of distanceBetweenLines
```
jonenst, line 416: Why did you remove the 9999. special case ?

src/components/network/line-layer.js
jonenst: Do we really need the +x% ? I don't see any problem without it.
If we don't have it, then it means that for substations with 1 voltage level we don't have to translate the source at all

src/components/network/layers/fork-line-layer.js
JacquesBorsenberger: to have the same minRadiusMinPixels from the substation

src/components/network/line-layer.js
JacquesBorsenberger: we always want a forkline to connect to the correct substation so it is a 0 offset when we do not want parallel line

src/components/network/layers/fork-line-layer.js
```diff
@@ -32,20 +34,32 @@ export default class ForkLineLayer extends LineLayer {
             'vs:#decl': `
 attribute float instanceLineParallelIndex;
 attribute float instanceLineAngle;
+attribute float instanceOffsetStart;
 uniform float distanceBetweenLines;
 uniform float maxParallelOffset;
 uniform float minParallelOffset;
+uniform float substationRadius;
+uniform float substationMaxPixel;
```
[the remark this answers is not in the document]
JacquesBorsenberger: dont know why but it was already bugged : here it is without this modification
image

src/components/network/line-layer.js
jonenst: Then we should remove the 9999. handling in the shaders as well.

We had the 9999. special value because we don't want the arrow to stop before reaching the substation in dynamic mode. So we lose this in favor of having the line connect to the correct ring of the voltage level

PR #265, opened 2020-11-19
Fix scroll

med-hedi:

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -5,7 +5,7 @@
     file, You can obtain one at http://mozilla.org/MPL/2.0/.
 -->
 <!DOCTYPE html>
-<html lang="en">
+<html class="singlestretch-parent" lang="en">
     <head>
         <base href="%PUBLIC_URL%/" />
         <meta charset="utf-8" />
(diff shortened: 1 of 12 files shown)


jonenst (changes requested):
this breaks the results and table tabs:
before: Capture d’écran de 2020-11-23 16-25-29

after:
Capture d’écran de 2020-11-23 16-25-48

jonenst:
Looking into this, we have a problem throughout the app. I think we need to change globaly how we handle elements like lists and table that need to grow to the bottom of the page:
Right now, we use
- height: calc ( 100vh - 65px) (note the off by one error here, our topbar is 64 px not 65px..) for the map and the table tabs.
- autosizer with a somehow random value ( height - 48 ) for the child in the table and results tabs
- height: calc ( 100vh - 65px) in the results page although there are extra tabs (loadflow/security analysis) before the main content so the 65px is wrong and should be 64 (topbar) + 61 (tabs)  = 125px (hence the scrollbar)


Instead, I think we should always use autosize as a single child of a parent which has correct size. The parent gets the correct size by using flex-grow: 1. The correct height is computed up until the root ```<html>``` tag.

med-hedi:
Ok i'm going to do that follow what you said on slack

geofjamg (approved):

jonenst (approved):
this one brings joy

PR #268, opened 2020-11-26
switch config ui to backend and add notifs

AbdelHedhili:


ChamseddineBhd (approved):

geofjamg (changes requested):
As discused last week in standup meeting: theme is a very special case because we need to get the theme before being logged in and we cannot get parameter frol server before being logged. So for the theme we are going to use both server side saving and local storage so that when we load the application, we first use local storage param to get the theme, then once we are logged we get one from server.
When we change the theme in the UI OR when the theme is changed by a notification, we also update the local storage.

src/redux/reducer.js
```diff
@@ -75,21 +51,21 @@ const initialState = {
...
+    useName: true,
     user: null,
-    centerLabel: getLocalStorageCenterLabel(),
-    diagonalLabel: getLocalStorageDiagonalLabel(),
-    lineFullPath: getLocalStorageLineFullPath(),
-    lineParallelPath: getLocalStorageLineParallelPath(),
-    lineFlowMode: getLocalStorageLineFlowMode(),
-    lineFlowColorMode: getLocalStorageLineFlowColorMode(),
-    lineFlowAlertThreshold: getLocalStorageLineFlowAlertThreshold(),
+    centerLabel: false,
```
SlimaneAmar, line 56: It is not possible to use a dictionary ?

AbdelHedhili: what do you mean ?

src/components/parameters.js
```diff
@@ -125,36 +123,59 @@ const Parameters = ({ showParameters, hideParameters }) => {
 
     const handleChangeTheme = (event) => {
         const theme = event.target.value;
-        dispatch(selectTheme(theme));
+        let json = JSON.stringify([
```
geofjamg, line 126: Why do you strinify the object here ? This is really ugly in the code and cumbersome. If you need to stringify just to put it in the body (not sure this is necessaey, to check) why not doing it in the  updateConfigParameters function?

src/components/app.js
```diff
@@ -206,6 +309,25 @@ const App = () => {
         }
     }, [user]);
 
+    useEffect(() => {
+        if (user !== null) {
+            fetchConfigParameters().then((params) => {
+                console.debug('received UI parameters :');
+                console.debug(params);
+                //if it's the user first connexion we want to set the default parameters in the database
```
geofjamg, line 317: Why ?

geofjamg (changes requested):

geofjamg (approved):

PR #274, opened 2020-12-07
Display voltage level on click on security analysis constraint

FranckLecuyer:


med-hedi (approved):

geofjamg (changes requested):
There is a refresh issue with the diagram, when a substation has already been open and you click on an equipment on security analysis result. The old one diseapear, it loads then the old one go back and then new appears.

geofjamg (approved):

PR #276, opened 2020-12-10
Fix pan reset after switch state change

JacquesBorsenberger:


geofjamg (changes requested):
There is a glitch when we switch to another voltage level if we have zoomed in. We first see the voltage level without the zoom and then the new voltage level.

jonenst:
I think we said we should not merge this because the fix comes from another PR where the svg is not downloaded 3 times ?

PR #277, opened 2020-12-11
notify on study creation error

FranckLecuyer:


src/components/study-manager.js
```diff
@@ -494,7 +499,25 @@ const StudyManager = ({ onClick }) => {
         const ws = connectNotificationsWsUpdateStudies();
 
         ws.onmessage = function (event) {
-            dispatchStudies();
+            let eventData = JSON.parse(event.data);
+            if (eventData.headers) {
```
SlimaneAmar, line 503: Perhaps a function `displayErrorIfExist`

src/components/study-manager.js
```diff
@@ -494,7 +499,25 @@ const StudyManager = ({ onClick }) => {
...
+                const error = eventData.headers['error'];
+                if (error !== undefined) {
+                    const studyName = eventData.headers['studyName'];
+                    const errorMessage =
```
SlimaneAmar, line 507: Perhaps a const for hearder name `error` or const enum for all header names of te web socket (study)

src/components/study-manager.js
```diff
@@ -480,6 +483,8 @@ const StudyManager = ({ onClick }) => {
 
     const classes = useStyles();
 
+    const { enqueueSnackbar } = useSnackbar();
+
```
SlimaneAmar, line 487: Why we not use commons-ui snackbar for this ?

src/components/study-manager.js
FranckLecuyer: Done

src/components/study-manager.js
FranckLecuyer: I'm using snackbar from commons-ui : see app.js where SnackbarProvider is imported and used

See demo/src/app.js in commons-ui, where there is an exemple of snackbar use

SlimaneAmar: ok

SlimaneAmar (approved):

PR #278, opened 2020-12-15
Security analysis results minor changes

FranckLecuyer:


JacquesBorsenberger (approved):

geofjamg (changes requested):
There is a regression: empty contingencies are not filtered anymore so results are totally unreadable.

geofjamg (approved):

PR #289, opened 2021-01-21
Add tooltip for substation name

med-hedi:

### DIFF
--- a/src/components/network/network-explorer.js
+++ b/src/components/network/network-explorer.js
@@ -18,6 +18,7 @@ import { darken, lighten, makeStyles } from '@material-ui/core/styles';
 import ListItem from '@material-ui/core/ListItem';
 import SearchIcon from '@material-ui/icons/Search';
 import TextField from '@material-ui/core/TextField';
+import Tooltip from '@material-ui/core/Tooltip';
 
 import Network from './network';
 import Divider from '@material-ui/core/Divider';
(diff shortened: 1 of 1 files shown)


src/components/network/network-explorer.js
```diff
@@ -80,8 +81,39 @@ const useStyles = makeStyles((theme) => ({
...
+            overflow: 'hidden',
+            whiteSpace: 'nowrap',
+            textOverflow: 'ellipsis',
+        },
+    },
+}));
+
+const useStylesCustomTooltip = makeStyles((theme) => ({
+    arrow: {
+        color: theme.palette.type === 'dark' ? '#404040' : '#f0f0f0',
```
geofjamg, line 100: We should never do that as if we add a new theme we will need to change this code. 
Instead enrich themes with new variable and use it here

geofjamg (changes requested):

src/components/network/network-explorer.js
```diff
@@ -247,68 +287,98 @@ const NetworkExplorer = ({
...
-                                                    : substation.id}
-                                            </Typography>
+                        {substationAndCountry.length >
+                        maxLengthSubstationName ? (
+                            <CustomTooltip title={substationAndCountry}>
+                                <ListItem
+                                    component={'li'}
+                                    button
+                                    key={substation.id}
+                                    className={classes.selectedTooltipName}
```
jonenst, line 297: missing classes when selectItem === substation.id to highlight selection

src/components/network/network-explorer.js
```diff
@@ -247,68 +287,98 @@ const NetworkExplorer = ({
...
-                                                noWrap
-                                            >
-                                                {useName
-                                                    ? substation.name
-                                                    : substation.id}
-                                            </Typography>
+                        {substationAndCountry.length >
+                        maxLengthSubstationName ? (
+                            <CustomTooltip title={substationAndCountry}>
+                                <ListItem
```
jonenst, line 293: We should refactor to avoid duplicating this `<ListItem>..<ListItem>` code

src/components/network/network-explorer.js
```diff
@@ -125,6 +151,8 @@ const NetworkExplorer = ({
 
     const listeRef = useRef(null);
 
+    const maxLengthSubstationName = 21; // Max length of substation name
```
jonenst, line 154: Not sure if it's better to use css text ellipsis ? https://material-ui.com/system/display/#text-overflow
It makes it harder to detect which substations needs a tooltip. But then again, tooltips are not very userfriendly on touch devices

jonenst (changes requested):
refactor to not duplicate the code, and always pass the full text to the component which then decides to truncate or use css ellipsis based on the implementation

src/components/network/network-explorer.js
```diff
@@ -80,8 +81,33 @@ const useStyles = makeStyles((theme) => ({
             color: theme.palette.type === 'dark' ? '#000' : '#FFF',
         },
     },
+    selectedTooltipName: {
+        height: (itemSize * 3) / 4,
```
jonenst, line 85: Do we really need this ad hoc css code ?
I guess the 230 is based on the drawer width (300), at least it should use the variable to compute the 230 from the 300

src/components/network/network-explorer.js
jonenst: Ha my bad, it already uses text ellipsis, and the limit of 21 chars is only used to decide to put the tooltip or not. it's ok then. We could use code like https://stackoverflow.com/questions/56588625/react-show-material-ui-tooltip-only-for-text-that-has-ellipsis to dynamically see if there is an ellipsis but since the drawer has a fixed width, the 21 chars is a good solution for now.

(until we decide that on 2000px large screen, we can afford to have a bigger drawer...need a UI expert for this)

PR #291, opened 2021-01-27
Fix voltage level / substation highlighting

JacquesBorsenberger:
scroll to substation and highlight voltageLevel in network explorer when navigating or opening voltage level


geofjamg (changes requested):
Something does not work anymore around loading feedback. We feel like navigating is lagging because progress bar is starting display at the very end of the navigation.

src/components/study-pane.js
```diff
@@ -777,6 +777,19 @@ const StudyPane = (props) => {
         }
     }
 
+    function scrollAndOpenVoltageLevel(vlId) {
+        if (!network) return;
+        setDisplayedVoltageLevelId(vlId);
+        window.setTimeout(
```
jonenst, line 783: why do we need setTimeout here to call setState ?

src/components/network/network-explorer.js
```diff
@@ -174,10 +174,12 @@ const NetworkExplorer = ({
                     index++;
                 }
             }
-            listeRef.current.scrollToRow(index);
-            // Workaround, remove when https://github.com/bvaughn/react-virtualized/issues/995 is resolved
             setTimeout(() => {
                 listeRef.current.scrollToRow(index);
+                // Workaround, remove when https://github.com/bvaughn/react-virtualized/issues/995 is resolved
```
jonenst, line 179: why do you need to change this and nest everything in another setTimeout ?

src/components/study-pane.js
```diff
@@ -777,6 +777,13 @@ const StudyPane = (props) => {
         }
     }
 
+    function scrollAndOpenVoltageLevel(vlId) {
```
jonenst, line 780: since showVoltageLevelDiagram, this breaks changing the url to reflect the currently shown sld

geofjamg:
To not decrease the sld browsing experience, it had been decided with @mikatheo to not scroll in the list to show the selected voltage level but just to fix the selection issue.

jonenst (changes requested):
fix urls not updating

geofjamg:
so @JacquesBorsenberger can you remove the scrolling part to just keep the fix so we can merge it?

JacquesBorsenberger:
do we also remove the scrolling when coming from result table ?

jonenst:
@JacquesBorsenberger no keep the scrolling coming from the table, only remove it when clicking on the sld navigation arrows

src/components/study-pane.js
```diff
@@ -777,6 +777,13 @@ const StudyPane = (props) => {
...
+        if (!network) return;
+        setDisplayedVoltageLevelId(vlId);
```
jonenst, line 782: don't call this, showVOltageLevelDiagram will update the url and this will trigger an useEffect to update displayedVoltageLevelId . (the same bug is present in onClickNmKConstraint, we should clean it later)

src/components/study-pane.js
```diff
@@ -777,6 +777,13 @@ const StudyPane = (props) => {
...
+        showVoltageLevelDiagram(vlId)
+        dispatch(selectItemNetwork(vlId));
```
jonenst, line 784: please investigate if we can remove selectItemNetwork and just base highlighting on displayedVoltageLevelId and displayedSubstationId

jonenst: maybe in another PR though

JacquesBorsenberger: setDisplayedVoltageLevelId(vlId)  :  sld displayed
showVoltageLevelDiagram(vlId) : url
dispatch(selectItemNetwork(vlId)) : highlight in network-explorer

geofjamg (approved):

PR #293, opened 2021-02-01
Lazy loading gps data

JacquesBorsenberger:


src/components/network/network.js
```diff
@@ -5,37 +5,39 @@
...
+import { RemoteRessourceHandler } from '../util/remote-ressource-handler';
+
 const elementIdIndexer = (map, element) => {
     map.set(element.id, element);
     return map;
 };
 
 export default class Network {
-    substations = [];
+    substations;
```
jonenst, line 16: why not `= undefined` like this others ?

src/components/network/network.test.js
```diff
@@ -35,21 +35,22 @@ const substations = [
 
 test('network', () => {
     // Data initialisation
-    const network = new Network();
-    network.setSubstations(substations);
+    const network = new Network(() => substations);
+    //network.setSubstations(substations);
```
jonenst, line 39: don't comment old code, just delete it

src/components/security-analysis-result.js
```diff
@@ -479,34 +482,56 @@ const SecurityAnalysisResult = ({ result, onClickNmKConstraint }) => {
                     )}
                 </div>
                 <div style={{ flexGrow: 1 }}>
+                    {' '}
```
jonenst, line 485: remove this ?

src/components/security-analysis-result.js
```diff
@@ -479,34 +482,56 @@ const SecurityAnalysisResult = ({ result, onClickNmKConstraint }) => {
...
+                if (resultFetched) {
+                    setResultFetched(false);
+                }
+                resultFetcher.get(() => setResultFetched(true));
+                return (
+                    <LoaderWithOverlay
+                        color="inherit"
+                        loaderSize={70}
+                        isFixed={true}
+                        loadingMessageText={'Loading'}
```
jonenst, line 517: i18n

src/components/network/network-table.js
```diff
@@ -1234,6 +1241,26 @@ const NetworkTable = (props) => {
...
+                setDataFetched({ ...dataFetched, ...{ [index]: false } });
+            resource.get(() =>
+                setDataFetched({ ...dataFetched, ...{ [index]: true } })
+            );
+            return (
+                <LoaderWithOverlay
+                    color="inherit"
+                    loaderSize={70}
+                    isFixed={true}
+                    loadingMessageText={'Loading'}
```
jonenst, line 1257: i18n

src/components/study-pane.js
```diff
@@ -356,87 +354,44 @@ const StudyPane = (props) => {
...
+                    () => fetchLccConverterStations(studyName, userId),
+                    () => fetchVscConverterStations(studyName, userId),
+                    () => fetchShuntCompensators(studyName, userId),
+                    () => fetchStaticVarCompensators(studyName, userId),
+                    (error) => {
+                        console.error(error.message);
+                        setStudyNotFound(true);
+                    }
+                );
                 dispatch(loadNetworkSuccess(network));
```
jonenst, line 383: can we remove this, since all things in the network are now lazy loaded ?

src/components/util/remote-ressource-handler.js
```diff
@@ -0,0 +1,54 @@
...
+    updating = undefined;
+
+    cbUpdateDone = new Set();
+
+    get(cbUpdateDone) {
+        if (this.values === undefined) {
+            if (cbUpdateDone) this.cbUpdateDone.add(cbUpdateDone);
+            if (!this.updating) {
+                this.updating = true;
+                Promise.all([this.fetcher()])
```
jonenst, line 29: why Promise.all of a single element array ?

src/components/util/remote-ressource-handler.js
```diff
@@ -0,0 +1,54 @@
...
+                    .then((val) => {
+                        this.values = val[0];
+                        if (this.postUpdate) this.postUpdate(this.values);
+                        this.updating = false;
+                        this.cbUpdateDone.forEach((cb) => cb());
+                        this.cbUpdateDone.clear();
+                    })
+                    .catch((error) => {
+                        if (this.errorHandler) this.errorHandler(error);
+                        this.values = {};
```
jonenst, line 39: in case of error, why use {}. maybe find a better api ?

src/components/util/remote-ressource-handler.js
```diff
@@ -0,0 +1,54 @@
...
+            return undefined;
+        }
+        if (cbUpdateDone) cbUpdateDone();
+        return this.values;
+    }
+
+    length(cbUpdateDone) {
+        return (
+            (this.values !== undefined && this.values.length) ||
+            this.get(cbUpdateDone)
```
jonenst, line 51: this returns the values, not the length ?

src/components/network/line-layer.js
```diff
@@ -168,13 +168,24 @@ function getArrowSpeedFactor(speed) {
     }
 }
 
+// a frame is 16 ms at 60fps (target render for most browser)
+const tsThreshold = 16;
+const PATH_UPDATED = 'lastPathUpdate';
+const LINE_UPDATED = 'lineUpdated';
+const PROXIMITY_UPDATED = 'lastPathUpdate';
```
jonenst, line 175: why same constant as PATH_UPDATED ?

src/components/network/network-explorer.js
```diff
@@ -107,6 +107,16 @@ const NetworkExplorer = ({
     const [filteredVoltageLevels, setFilteredVoltageLevels] = React.useState(
         []
     );
+    const [substationsLoaded, setSubstationsLoaded] = React.useState(false);
+
+    useEffect(() => {
```
jonenst, line 112: I think we should simplify the API. Maybe create our own hook useNetwork or something ? I'm not sure. It's a pain to have to deal with the loading everywhere.

src/components/util/remote-ressource-handler.js
```diff
@@ -0,0 +1,54 @@
...
+        this.errorHandler = errorHandler;
+        this.postUpdate = postUpdate;
+    }
+
+    values = undefined;
+    updating = undefined;
+
+    cbUpdateDone = new Set();
+
+    get(cbUpdateDone) {
```
jonenst, line 24: I think this API is a bit weird.. always return a promise, instead of returning undefined or the value, and calling the callback ? Or if we stick with the current API, rename get to getOrFetch ? getOtherwiseFetch ?

src/components/util/remote-ressource-handler.js
```diff
@@ -0,0 +1,54 @@
...
+                        this.values = {};
+                    });
+            }
+            return undefined;
+        }
+        if (cbUpdateDone) cbUpdateDone();
+        return this.values;
+    }
+
+    length(cbUpdateDone) {
```
jonenst, line 48: rename to lengthOrFetch? lengthOtherwiseFetch ?

src/components/network/network.js
```diff
@@ -271,4 +234,71 @@ export default class Network {
             this.getThreeWindingsTransformer(id)
         );
     }
+
+    makeEqHandler(cb, errHandler, postUpdateCB) {
```
jonenst, line 238: please rename to makeEquipmentHandler, longer names are preferable, and eq is often used to mean "equals"

src/components/network/network.js
```diff
@@ -5,37 +5,39 @@
...
     map.set(element.id, element);
     return map;
 };
 
 export default class Network {
-    substations = [];
+    substations;
 
-    lines = [];
+    lines = undefined;
```
jonenst, line 18: why
```lines = undefined;```
instead of just
```lines;```
(same for all others)

File: src/components/network/network.js [hunk 16]
```diff
@@ -73,9 +75,9 @@ export default class Network {
             });
         });
 
-        this.voltageLevels = this.substations.flatMap(
-            (substation) => substation.voltageLevels
-        );
+        this.voltageLevels = this.substations
+            .get()
```
jonenst, line 79: why .get() here and .values above ?

src/components/util/remote-ressource-handler.js
```diff
@@ -0,0 +1,54 @@
...
+    cbUpdateDone = new Set();
+
+    get(cbUpdateDone) {
+        if (this.values === undefined) {
+            if (cbUpdateDone) this.cbUpdateDone.add(cbUpdateDone);
+            if (!this.updating) {
+                this.updating = true;
+                Promise.all([this.fetcher()])
+                    .then((val) => {
+                        this.values = val[0];
```
jonenst, line 31: when Promise.all is removed, remove val[0]

src/components/network/line-layer.js
```diff
@@ -544,267 +1027,36 @@ class LineLayer extends CompositeLayer {
...
-        });
+    genLineKey(line) {
+        return line.voltageLevelId1 > line.voltageLevelId2
+            ? line.voltageLevelId1 + '##' + line.voltageLevelId2
+            : line.voltageLevelId2 + '##' + line.voltageLevelId1;
+    }
 
-        return layers;
+    /*
+     *  function used to call onNext in the same 'frame' on the next one if we're short on time
```
jonenst, line 1043: "or" the next one

src/components/network/line-layer.js
```diff
@@ -544,267 +1027,36 @@ class LineLayer extends CompositeLayer {
...
+     *   timestamp : when the 'frame' started
+     *   onNext : callback to call, if undefined does nothing
+     * */
+    scheduleNext(timestamp, onNext) {
```
jonenst, line 1047: timestamp is never used ?

jonenst:
Did you see https://developer.mozilla.org/en-US/docs/Tools/Performance/Scenarios/Intensive_JavaScript ? It recomends using RequestAnimationFrame instead of window.setTimeout

src/components/network/line-layer.js
```diff
@@ -195,336 +206,808 @@ class LineLayer extends CompositeLayer {
...
-    genLineKey(line) {
-        return line.voltageLevelId1 > line.voltageLevelId2
-            ? line.voltageLevelId1 + '##' + line.voltageLevelId2
-            : line.voltageLevelId2 + '##' + line.voltageLevelId1;
+    updateLines(ts, updatedLines) {
+        const linesByNominalVoltage = this.state.linesByNominalVoltage;
+        const nominalVoltageUpdated = new Set();
+        this.workArray(updatedLines, (i) => {
+            const line = updatedLines[i];
+            //nominalVoltageIndex ti plop
```
jonenst, line 561: remove this comment

src/components/network/line-layer.js
```diff
@@ -195,336 +206,808 @@ class LineLayer extends CompositeLayer {
...
 
-                line.angle = this.computeAngle(
-                    props,
-                    positions[0],
-                    positions[positions.length - 1]
+        nvUpdated.forEach((nv) =>
+            nvIndexes.push(this.props.network.nominalVoltages.indexOf(nv))
+        );
+        this.workArray(nvIndexes, (i) =>
+            this.setState({ [LINE_UPDATED + nvIndexes[i]]: performance.now() })
```
jonenst, line 599: why not ts ?

src/components/network/line-layer.js
```diff
@@ -195,336 +206,808 @@ class LineLayer extends CompositeLayer {
...
+                                props.disconnectedLineColor,
+                                props.lineFlowColorMode,
+                                props.lineFlowAlertThreshold,
+                                props.updatedLines,
+                            ],
+                        },
+                    })
+                )
+        );
+        this.setState({ layers: layers }, () => this.scheduleNext(next));
```
jonenst, line 825: when is the lambda executed ? What is the value of next ?

src/components/network/line-layer.js
```diff
@@ -195,336 +206,808 @@ class LineLayer extends CompositeLayer {
...
+                                props.lineFlowColorMode,
+                                props.lineFlowAlertThreshold,
+                                props.updatedLines,
+                            ],
+                        },
+                    })
+                )
+        );
+        this.setState({ layers: layers });
+        let next = undefined;
```
jonenst, line 928: why `= undefined`

src/components/network/line-layer.js
```diff
@@ -195,336 +206,808 @@ class LineLayer extends CompositeLayer {
...
+                    nominalVoltageIndex + 1,
+                    doLabel,
+                    doNextNominalVoltage
                 );
-                line.proximityFactorEnd = this.getProximityFactor(
-                    positions[positions.length - 2],
-                    positions[positions.length - 1]
+        else if (doLabel)
+            next = (ts) => this.computeLabels(ts, 0, doNextNominalVoltage);
+        if (next !== undefined) this.scheduleNext(timestamp, next);
```
jonenst, line 942: scheduleNext already does the if undefined, remove it ?

jonenst: Or does scheduleNext really need to do the check ? In this case, it's better to remove it in scheduleNext ?

src/components/network/line-layer.js
```diff
@@ -195,336 +206,808 @@ class LineLayer extends CompositeLayer {
...
+                            ],
+                            getText: [
+                                state[ACTIVE_POWER + nominalVoltageIndex],
+                            ],
+                            data: [state[ACTIVE_POWER + nominalVoltageIndex]],
+                        },
+                    })
+                )
+        );
+        this.setState({ layers: layers }, () => this.scheduleNext(next));
```
jonenst, line 988: same question, when is the lambda executed and what is the value of next?

src/components/network/line-layer.js
```diff
@@ -195,336 +206,808 @@ class LineLayer extends CompositeLayer {
...
+
+        let next = undefined;
```
jonenst, line 990: why `= undefined`

src/components/network/line-layer.js
```diff
@@ -195,336 +206,808 @@ class LineLayer extends CompositeLayer {
...
+        if (
+            doNextNominalVoltage &&
+            this.hasNextNominalVoltage(nominalVoltageIndex)
+        )
+            next = (ts) =>
```
jonenst, line 995: with the current scheduleNext that checks its input, you could write:
const next =  doNextNominalVoltage && this.hasNextNominalVoltage && (ts) => this.computeLabels(
                    ts,
                    nominalVoltageIndex + 1,
                    doNextNominalVoltage
                );

src/components/network/line-layer.js
```diff
@@ -544,267 +1027,36 @@ class LineLayer extends CompositeLayer {
...
+     *   timestamp : when the 'frame' started
+     *   onNext : callback to call, if undefined does nothing
+     * */
+    scheduleNext(timestamp, onNext) {
+        if (onNext !== undefined)
+            window.setTimeout(() => onNext(performance.now()), 0);
+    }
+
+    workArray(array, cb, currentIndex) {
+        let start = performance.now();
```
jonenst, line 1053: pass the array length directly

src/components/network/line-layer.js
```diff
@@ -544,267 +1027,36 @@ class LineLayer extends CompositeLayer {
...
+     *  function used to call onNext in the same 'frame' on the next one if we're short on time
+     *   timestamp : when the 'frame' started
+     *   onNext : callback to call, if undefined does nothing
+     * */
+    scheduleNext(timestamp, onNext) {
+        if (onNext !== undefined)
+            window.setTimeout(() => onNext(performance.now()), 0);
+    }
+
+    workArray(array, cb, currentIndex) {
```
jonenst, line 1052: I don't think workArray in its current form does what you think it does. To me, it does respect tsThreshold because after the while loop, it just reexcutes itself with a new start.. but it doesn't yield execution at all.

What was workArray supposed to do ?

jonenst:
I think this is a very complex change, we need to decide carefully if we want to merge

src/components/network/network-explorer.js
```diff
@@ -167,6 +177,7 @@ const NetworkExplorer = ({
         network,
         identifiedElementComparator,
         generateFilteredSubstation,
+        substationsLoaded,
```
jonenst, line 180: add && substationsLoaded to the if above ?

src/components/network/network-explorer.js
```diff
@@ -107,6 +107,16 @@ const NetworkExplorer = ({
...
+    const [substationsLoaded, setSubstationsLoaded] = React.useState(false);
+
+    useEffect(() => {
+        if (
+            network &&
+            network.substations.get(() => setSubstationsLoaded(true)) ===
+                undefined
+        )
+            setSubstationsLoaded(false);
+    }, [substationsLoaded, network]);
```
jonenst, line 119: why depend on substationLoaded ?

src/components/network/network-explorer.js
```diff
@@ -187,7 +198,7 @@ const NetworkExplorer = ({
                 listeRef.current.scrollToRow(index);
             }, 0);
         }
-    }, [visibleSubstation, filteredVoltageLevels]);
+    }, [visibleSubstation, filteredVoltageLevels, substationsLoaded]);
```
jonenst, line 201: Add substationsLoaded to the if above ?

src/components/network/network.js
JacquesBorsenberger: this one is computed from voltages levels

src/components/network/network.test.js
JacquesBorsenberger: done

src/components/security-analysis-result.js
JacquesBorsenberger: done

src/components/util/remote-ressource-handler.js
JacquesBorsenberger: corrected

src/components/network/line-layer.js
JacquesBorsenberger: shouldn't, modified

src/components/util/remote-ressource-handler.js
JacquesBorsenberger: done

src/components/network/network.js
JacquesBorsenberger: done

src/components/network/network.js
JacquesBorsenberger: done

src/components/network/network.js
JacquesBorsenberger: use values

src/components/util/remote-ressource-handler.js
JacquesBorsenberger: done

src/components/network/line-layer.js
JacquesBorsenberger: done

src/components/network/line-layer.js
JacquesBorsenberger: was used in debug for searching long loop, removed

src/components/network/line-layer.js
JacquesBorsenberger: done

src/components/network/line-layer.js
JacquesBorsenberger: done

src/components/network/line-layer.js
JacquesBorsenberger: should not be there.

src/components/network/line-layer.js
JacquesBorsenberger: long habit of never declare uninitialized variables

src/components/network/line-layer.js
JacquesBorsenberger: removed

src/components/network/line-layer.js
JacquesBorsenberger: ok

src/components/network/line-layer.js
JacquesBorsenberger: missing a scheduleNext... It was supposed to cut long loops between frames.

src/components/network/network-explorer.js
JacquesBorsenberger: done

jonenst:
Also, storing the timestamp of the computation in the state and using as an updateTrigger is weird, maybe we can do better ?

src/components/network/line-layer.js
JacquesBorsenberger: removed

src/components/network/line-layer.js
```diff
@@ -195,220 +206,283 @@ class LineLayer extends CompositeLayer {
...
-                    new Map()
-                );
+    /*
+     *   first things to do when new data, index line by nomimalVoltage
+     *   timestamp : time when the batch was called used to evaluate when we need the pass to the next 'frame'
+     *   nominalVoltageIndex : nominal voltage index (in network.nominalVoltages
+     *   linesByNominalVoltage : map [nominalVoltageIndex: [voltagesLevels] ]
+     *   next : computeLinesMetadata
+     * */
+    indexLinesByNominalVoltage(timestamp) {
```
jonenst, line 224: unused param timestamp

jonenst:
discussed with geoffroy,
- we think that the RemoteRessourceHandler API is too weird to be merged as is. We need to find a better way to deal with this. I think it should be possible to keep network.lines as the real line array, and store the fact that this has been loaded elsewhere in the network object (network.loaded[lines] or something). And use the natural react mecanism to rerender components when the needed data has been loaded and the object network.xxx is present.
- the state machine of the line layer is complex and hard to maintain. it uses uncommon practices (storing timestamps in the state). It should at least be simplified before we merge. But we would like to see the code comparison with a solution based on a worker thread sending messages to the main thread at each step of the computation to decide which is best to merge.

PR #299, opened 2021-02-09
fix rename study

ChamseddineBhd:


jonenst (changes requested):
don't fetchStudies manually, it's already done by the notifications

jonenst (approved):

PR #301, opened 2021-02-09
fields could be null

ChamseddineBhd:


jonenst (changes requested):
Need to decide if these are really required and force the caller to not use the component if the data is not ready, or if these are optional.

PR #305, opened 2021-02-15
Take into account the latest i18n/L10n correction report

niconoir:


src/translations/en.json
```diff
@@ -3,8 +3,10 @@
     "addNewStudyDescription": "To add a new study, you should enter the following information:",
     "caseExist": "Does the case already exist?",
     "studyName": "study name",
+    "studyNameProperty": "study name: ",
     "studyDescription": "study description",
-    "owner": "Owner",
+    "studyDescriptionProperty": "study description: ",
```
SlimaneAmar, line 8: case-sensitive is is taken into account

src/translations/en.json
```diff
@@ -124,7 +126,7 @@
     "RatioTap3": "Ratio tap 3",
     "PhaseTap3": "Phase tap 3",
 
-    "viewOverloadsTable": "View overloads table",
+    "viewOverloadsTable": "View overload table: ",
```
SlimaneAmar, line 129: There are several overloads

src/translations/en.json
```diff
@@ -69,18 +71,18 @@
...
     "StaticArrows": "Static arrows",
     "AnimatedArrows": "Animated arrows",
     "Feeders": "Feeder",
-    "LineFlowColorMode": "Line color mode",
+    "LineFlowColorMode": "Line color mode: ",
     "NominalVoltage": "Nominal voltage",
     "Overloads": "Overloads",
     "Overload": "Overload",
-    "AlertThresholdLabel": "Alert threshold",
+    "AlertThresholdLabel": "Threshold for displaying lines\n before overload (%): ",
```
SlimaneAmar, line 85: It seems that the newline is not taken into account ?

src/translations/en.json
niconoir: This change was made on purpose, it was required.

src/translations/en.json
niconoir: Yes, it was added for test purpose and I forgot to remove it. Removed.

src/translations/en.json
niconoir: I agree, added uppercases where necessary.

SlimaneAmar (approved):

FranckLecuyer (approved):

PR #307, opened 2021-02-18
Stop sa computation fixes

ChamseddineBhd:


jonenst:
I like that there is a runnable with a text and its actions together, instead of having just text and the action be hardcoded to "onStopComputationClick".

But is it ok to have only one runnable, whereas before we could have several ?

FranckLecuyer:
> I like that there is a runnable with a text and its actions together, instead of having just text and the action be hardcoded to "onStopComputationClick".
> 
> But is it ok to have only one runnable, whereas before we could have several ?

There are still several runnables, but only one action (stop) on these runnables, for the moment

jonenst (approved):

PR #311, opened 2021-03-05
Allow the slider to move with the mouse

niconoir:


AbdelHedhili (approved):

geofjamg (changes requested):
When we open a study with overload table already checked, the slider is disable we have to uncheck and recheck.

image

geofjamg:
> When we open a study with overload table already checked, the slider is disable we have to uncheck and recheck.
> 
> image

Seem to be already present in master

geofjamg (approved):

PR #316, opened 2021-03-15
Make displayed columns parameter persistent

SlimaneAmar:


src/components/network/network-table.js
```diff
@@ -514,53 +563,51 @@ const NetworkTable = (props) => {
...
+        }
+        setBackupSelectedColumnsNames(null);
         setPopupSelectColumnNames(false);
-    };
+    }, [tabIndex, selectedColumnsNames, backupSelectedColumnsNames]);
 
     const handleToggle = (value) => () => {
-        const newChecked = new Set(checkedColumnsNames[tabIndex]);
-        if (checkedColumnsNames[tabIndex].has(value)) {
+        const newChecked = new Set([...selectedColumnsNames]); // copy
```
jonenst, line 590: why do you need a copy to create a new set anyway ?

src/components/network/network-table.js
```diff
@@ -94,6 +112,37 @@ const NetworkTable = (props) => {
         [lineEdit, tabIndex]
     );
 
+    useEffect(() => {
+        fetchConfigParameter(
```
jonenst, line 116: why fetch the parameters when changing tab ? Everything has already been fetched at page load and stored in redux, and the websocket connection updates redux when there is a change, so you can just use them direclty no?

src/components/network/network-table.js
SlimaneAmar: Because you can't modify directly the state variable `selectedColumnsNames` (immutable)
Is it true ?

src/components/network/network-table.js
```diff
@@ -69,21 +80,28 @@ const useStyles = makeStyles((theme) => ({
...
+
     const [tabIndex, setTabIndex] = useState(0);
     const [lineEdit, setLineEdit] = useState({});
     const [rowFilter, setRowFilter] = useState(undefined);
     const [popupSelectColumnNames, setPopupSelectColumnNames] = useState(false);
+    const [selectedColumnsNames, setSelectedColumnsNames] = useState(null);
+    const [
+        backupSelectedColumnsNames,
+        setBackupSelectedColumnsNames,
+    ] = useState(null); // Optimization to reduce database access
```
jonenst, line 100: it's better to use a ref if the component doesn't need to be rendered when the value changes

SlimaneAmar: yes

src/components/network/network-table.js
```diff
@@ -69,21 +80,28 @@ const useStyles = makeStyles((theme) => ({
...
 
+    const notifiedDisplayedColumns = useSelector(
+        (state) => state.displayedColumns
+    );
+
     const [tabIndex, setTabIndex] = useState(0);
     const [lineEdit, setLineEdit] = useState({});
     const [rowFilter, setRowFilter] = useState(undefined);
     const [popupSelectColumnNames, setPopupSelectColumnNames] = useState(false);
+    const [selectedColumnsNames, setSelectedColumnsNames] = useState(null);
```
jonenst, line 96: not sure about duplicating the values in redux here in the state of the component..

src/components/network/network-table.js
```diff
@@ -94,6 +112,37 @@ const NetworkTable = (props) => {
...
+            if (
+                TABLES_NAMES.indexOf(notifiedTableName) === tabIndex &&
+                JSON.stringify([...selectedColumnsNames]) !==
+                    notifiedDisplayedColumns.value
+            ) {
+                setSelectedColumnsNames(
+                    new Set(JSON.parse(notifiedDisplayedColumns.value))
+                );
+            }
+            dispatch(selectDisplayedColumns(null));
```
jonenst, line 142: it's weird to reset the redux state to null here. If redux states becomes the source of truth instead of the local copied state, this dispatch must be removed

src/components/network/network-table.js
```diff
@@ -724,4 +769,4 @@ NetworkTable.propTypes = {
     studyName: PropTypes.string,
 };
 
-export default NetworkTable;
+export default React.memo(NetworkTable);
```
jonenst, line 772: why React.memo ? props have not changed

src/components/network/network-table.js
SlimaneAmar: We need a local state which immediately reflects the choices made in the selection dialog (popuup)

src/components/network/network-table.js
SlimaneAmar: It is a hybrid (redux/local) solution to have a local state.
I will try to implement a solution full redux.

src/components/network/network-table.js
SlimaneAmar: See comment below

src/components/network/network-table.js
SlimaneAmar: new Set(selectedColumnsNames.values())

jonenst (approved):

PR #317, opened 2021-03-16
Tabular equipment view: automatically adapt column width to its content

JacquesBorsenberger:
make width optional, add Max/Min width,
calculate width based on content


src/components/util/virtualized-table.js
```diff
@@ -247,12 +301,16 @@ class MuiVirtualizedTable extends React.PureComponent {
...
         };
 
+        const sizes = this.sizes(this.props.columns);
         return (
             <AutoSizer>
                 {({ height, width }) => (
                     <Table
                         height={height}
-                        width={width}
+                        width={Math.max(
```
jonenst, line 310: This is bad because it overflows the parent:
Capture d’écran de 2021-03-31 16-46-27

src/components/util/virtualized-table.js
```diff
@@ -283,6 +342,7 @@ class MuiVirtualizedTable extends React.PureComponent {
                                     cellRenderer={this.cellRenderer}
                                     dataKey={dataKey}
                                     {...other}
+                                    width={sizes[dataKey]}
```
jonenst, line 345: also add `flexGrow={1}` to have a fullwidth table

src/components/util/virtualized-table.js
```diff
@@ -10,6 +10,33 @@ import { AutoSizer, Column, Table } from 'react-virtualized';
 import TableSortLabel from '@material-ui/core/TableSortLabel';
 import memoize from 'memoize-one';
 
+function getCSS(cssName) {
```
jonenst, line 13: this is too much of a hack, we can't merge it I think

jonenst (approved):

PR #321, opened 2021-04-01
Lazy data loading (network tables)

JacquesBorsenberger:


src/components/network/overloaded-lines-view.js
```diff
@@ -134,8 +133,9 @@ const OverloadedLinesView = (props) => {
                 <VirtualizedTable
                     height={Math.min(lines.length * rowHeight + 50, 400)}
                     className={classes.table}
-                    rowCount={lines.length}
-                    rowGetter={({ index }) => lines[index]}
+                    /*rowCount={lines.length}
+                    rowGetter={({ index }) => lines[index]}*/
```
jonenst, line 137: why not delete the code ?

src/components/security-analysis-result.js
```diff
@@ -479,15 +482,19 @@ const SecurityAnalysisResult = ({ result, onClickNmKConstraint }) => {
                     )}
                 </div>
                 <div style={{ flexGrow: 1 }}>
-                    {tabIndex === 0 &&
+                    {' '}
```
jonenst, line 485: remove this whitespace ?

src/components/network/network.js
```diff
@@ -271,4 +251,93 @@ export default class Network {
...
+        errHandler,
+        dispatch
+    ) {
+        this.generateEquipementHandler({
+            substations,
+            errHandler,
+            postUpdate: (net) => net.completeSubstationsInfos(),
+        });
+        this.lazyLoaders.set(
+            equipements.vol,
```
jonenst, line 301: is this undefined ?

src/components/network/config-tables.js
```diff
@@ -31,6 +34,8 @@ export const TABLES_DEFINITIONS = {
     VOLTAGE_LEVELS: {
         index: 1,
         name: 'VoltageLevels',
+        resource: equipements.voltageLevels,
```
jonenst, line 37: is this undefined (and it works because the other is equipements.vol which is also undefined) ?

JacquesBorsenberger: yeah there is something missing... like an entry in equipements, should be the key in the previous comment

src/components/network/equipment-table.js
```diff
@@ -0,0 +1,266 @@
...
+            cellRenderer: (cellData) =>
+                createEditableRow(cellData, equipmentType),
+        };
+    }
+
+    return fetching ? (
+        <LoaderWithOverlay
+            color="inherit"
+            loaderSize={70}
+            isFixed={true}
```
jonenst, line 249: dont put the loader on all the screen, instead put in only on the table

src/redux/reducer.js
```diff
@@ -100,6 +101,16 @@ export const reducer = createReducer(initialState, {
         state.network = action.network;
     },
 
+    [UPDATE_NETWORK]: (state, action) => {
+        let newNetwork = Object.assign(
```
jonenst, line 105: why not use new Network() ? Maybe add a comment to explain

src/components/network/network.js
```diff
@@ -271,4 +252,93 @@ export default class Network {
...
+                    value,
+                    (value) =>
+                        this.dispatch(updateNetwork(key, value, postUpdate)),
+                    (key) => errHandler(key)
+                )
+            );
+        }
+    }
+
+    fetchEquipment(equipement, cb) {
```
jonenst, line 271: why is the parameter called "cb" (I guess for callback?), but it's passed to fetch() which takes a boolean forceUpdate ?

src/components/util/remote-resource-handler.js
```diff
@@ -0,0 +1,35 @@
...
+
+    constructor(fetcher, setter, errorHandler) {
+        this.fetcher = fetcher;
+        this.setter = setter;
+        this.errorHandler = errorHandler;
+    }
+
+    fetched = undefined;
+
+    fetch(forceUpdate) {
```
jonenst, line 21: why add forceUpdate, it does not seem to be used

src/components/network/equipment-table.js
```diff
@@ -0,0 +1,266 @@
...
+        width: '100%',
+        height: '100%',
+        cursor: 'initial',
+    },
+}));
+
+const ROW_HEIGHT = 48;
+
+export const EquipmentTable = ({
+    props,
```
jonenst, line 30: don't name this just 'props', it's very surprising

src/components/network/equipment-table.js
```diff
@@ -0,0 +1,266 @@
...
+        flex: 1,
+        width: '100%',
+        height: '100%',
+        cursor: 'initial',
+    },
+}));
+
+const ROW_HEIGHT = 48;
+
+export const EquipmentTable = ({
```
jonenst, line 29: Is it better to only pass the array of the ressource instead of the full network ?

src/components/network/equipment-table.js
```diff
@@ -0,0 +1,266 @@
...
+        let groovyCr =
+            'equipment = network.get' +
+            lineEdit.equipmentType +
+            "('" +
+            lineEdit.id.replace(/'/g, "\\'") +
+            "')\n";
+        Object.values(lineEdit.newValues).forEach((cr) => {
+            groovyCr += cr.changeCmd.replace(/\{\}/g, cr.value) + '\n';
+        });
+        requestNetworkChange(props.userId, props.studyName, groovyCr).then(
```
jonenst, line 64: Needs to be updated, requestNetworkChange is now
```
export function requestNetworkChange(studyUuid, groovyScript) {
```

src/components/network/equipment-table.js
JacquesBorsenberger: need network to fetch the equipment,

jonenst (approved):

PR #323, opened 2021-04-12
Change language

etiennehomer:


src/index.js
```diff
@@ -33,17 +33,19 @@ const messages = {
...
-        <Provider store={store}>
-            <BrowserRouter basename={basename}>
-                <App />
-            </BrowserRouter>
-        </Provider>
-    </IntlProvider>,
-    document.getElementById('root')
-);
+const ApplicationWrapper = () => {
+    const [language, setLanguage] = useState();
```
ChamseddineBhd, line 39: what is the intial value of this state !  you should initialize it with SYSTEM ?

src/components/app.js
```diff
@@ -395,6 +416,11 @@ const App = () => {
         updateConfigParameter(PARAMS_USE_NAME_KEY, useName);
     };
 
+    const handleLanguageClick = (language) => {
```
ChamseddineBhd, line 419: you should also dispatch the new value to the store

ChamseddineBhd (changes requested):

src/components/app.js
```diff
@@ -157,13 +163,24 @@ const noUserManager = { instance: null, error: null };
 
 const STUDY_VIEWS = [StudyView.MAP, StudyView.SPREADSHEET, StudyView.RESULTS];
 
-const App = () => {
+const supportedLanguages = [FRENCH, ENGLISH];
+
+const getSystemLanguage = () => {
```
jonenst, line 168: rename to getComputedLanguage or something

src/index.js
jonenst: there should not be a state here since the language is stored in redux

src/components/app.js
etiennehomer: This method is really to get the language of the system. There's already a method named getComputedLanguage wheter it is a 'sys' code or a language code ('en', 'fr', etc)

src/components/app.js
jonenst: The value is dispatched to the store when the websocket notifies that the server has stored the new value. All configs behave this way. This PR should not change that so the code is correct like this I think

src/redux/reducer.js
```diff
@@ -55,6 +57,8 @@ const initialState = {
     selectedCase: null,
     selectedFile: null,
     useName: true,
+    language: 'sys',
+    computedLanguage: 'en',
```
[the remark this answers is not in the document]
etiennehomer: Done

jonenst (approved):

PR #328, opened 2021-04-20
Reduce latency in the parameter dialog when a change is made

SlimaneAmar:
Reduce latency in the parameter dialog when a change is made by using a local state
Implementation of a custom hook for parameters to reduce the duplicated code


niconoir (approved):

etiennehomer (approved):

src/components/network/network-table.js
```diff
@@ -140,10 +143,25 @@ const NetworkTable = (props) => {
     const handleSaveSelectedColumnNames = useCallback(() => {
         updateConfigParameter(
             COLUMNS_PARAMETER_PREFIX_IN_DATABASE + TABLES_NAMES[tabIndex],
-            JSON.stringify([...selectedColumnsNames])
+            JSON.stringify([...selectedColumnsNames]),
+            enqueueSnackbar,
+            intl.formatMessage({
+                id: 'paramsChangingError',
+            })
+        ).then(null, () =>
```
jonenst, line 151: use .catch instead

jonenst (approved):

PR #333, opened 2021-05-18
Correction in message on study creation error : display study name instead of study uuid

FranckLecuyer:


FranckLecuyer:
Already corrected in PR : https://github.com/gridsuite/gridstudy-app/pull/329

PR #349, opened 2021-06-09
[WIP] restore security analysis loader after first fetch

JacquesBorsenberger:


geofjamg (changes requested):
The second time we run a security analys, the loader is displayed even during the computation which is not what is expected. The loader should appear just during the time the result are refreshing.

PR #365, opened 2021-07-23
Display nominal voltage for equipments in tables

niconoir:


FranckLecuyer (approved):
Maybe this evolution could be done only in the front (so, without the PR in map-server), and could use the voltageLevels nominalVoltage values, already stored in the front network object ...

JacquesBorsenberger (approved):

FranckLecuyer (approved):

PR #367, opened 2021-07-23
use tooltip for long string in equipment table

JacquesBorsenberger:
Signed-off-by: jacques borsenberger <jacques.borsenberger@rte-france.com>


geofjamg (changes requested):
To improve performance and avoid adding many listeners to text cells, we should try to listen to a 'resize' event at table level and force a re-render in that case.

JacquesBorsenberger:
> To improve performance and avoid adding many listeners to text cells, we should try to listen to a 'resize' event at table level and force a re-render in that case.

seems that it works without listening to resize event... removed it

geofjamg (approved):

PR #382, opened 2021-10-06
Update contingency list to use element chooser

JacquesBorsenberger:
need an update of commons-ui

### DIFF
--- a/src/components/contingency-list-selector.js
+++ b/src/components/contingency-list-selector.js
@@ -5,7 +5,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import React, { useEffect, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { useParams } from 'react-router-dom';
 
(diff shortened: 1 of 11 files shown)


src/components/directory-item-selector.js
```diff
@@ -0,0 +1,100 @@
...
+
+    const directory2Tree = useCallback(
+        (newData) => {
+            const newNode = {
+                id: newData.elementUuid,
+                name: newData.elementName,
+                icon: getIconFor(newData.type, classes.icon),
+                children:
+                    newData.type === elementType.DIRECTORY
+                        ? [{ id: '', name: '' }]
```
sBouzols, line 37: I don't like this virtual/phantom child, we should make changes to TreeFinder component if there is a problem with empty directories which could be considered as leaf and then become selectable.

src/components/directory-item-selector.js
```diff
@@ -0,0 +1,100 @@
...
+            fetchRootFolders().then((roots) => {
+                setData(roots.map(directory2Tree));
+            });
+        }
+    }, [props.open, data, directory2Tree]);
+
+    const addToDirectory = useCallback(
+        (data, nodeId, content) => {
+            // create fake root
+            const node = nodeMap.current[nodeId];
```
sBouzols, line 56: why do you need to create a fake root ?

src/components/directory-item-selector.js
```diff
@@ -0,0 +1,100 @@
...
+            node.children = content.map(directory2Tree);
+            if (!node.children || node.children.length === 0) {
```
sBouzols, line 58: not a fan like already said

src/components/directory-item-selector.js
```diff
@@ -0,0 +1,100 @@
...
+    icon: {
+        marginRight: theme.spacing(1),
+        width: '18px',
+        height: '18px',
+    },
+}));
+
+const DirectoryItemSelector = (props) => {
+    const [data, setData] = useState([]);
+    const nodeMap = useRef({});
```
sBouzols, line 21: why do you need this nodeMap, you have the data state which is made for this usage.

src/components/contingency-list-selector.js
```diff
@@ -18,109 +18,212 @@ import CheckboxList from './util/checkbox-list';
...
+    const handleDelete = () => {
+        const toDelete = new Set(checkedContingencyListIds);
+        saveFavourite(
+            favoriteContingencyList.filter((item) => !toDelete.has(item))
+        );
+        setCheckedContingencyListIds([]);
+    };
+
+    const removeFromFavorite = (id) => {
+        saveFavourite(favoriteContingencyList.filter((item) => item !== id));
```
sBouzols, line 134: You don't reset setCheckedContingencyListIds here and remove this item.
You should factorize code here between this function and handleDelete I think.

src/components/util/checkbox-list.js
```diff
@@ -4,55 +4,111 @@
...
-            newChecked.push(value);
-        } else {
-            newChecked.splice(currentIndex, 1);
+    /* remove non absent selected items */
+    useEffect(() => {
+        const existingValues = new Set(props.values.map(props.id));
+        const newChecked = new Set(
+            [...checked].filter((id) => existingValues.has(id))
+        );
+        if (newChecked.size !== checked.size) {
```
sBouzols, line 72: This is bad, you could have add a value and remove one other then the sizes are equals but checked must be updated

sBouzols (changes requested):
Fine code review
No test of gridstudy-app AS, I will next time.

src/components/util/checkbox-list.js
JacquesBorsenberger: nope, the new elements must exists in the old set, if the size are equals then the set are equals

src/components/contingency-list-selector.js
```diff
@@ -18,109 +18,212 @@ import CheckboxList from './util/checkbox-list';
...
 import Button from '@material-ui/core/Button';
 import Alert from '@material-ui/lab/Alert';
-import Box from '@material-ui/core/Box';
 import Grid from '@material-ui/core/Grid';
+import DirectoryItemSelector from './directory-item-selector';
+import { PARAM_FAVORITE_CONTINGENCY_LISTS } from '../utils/config-params';
+import { useSelector } from 'react-redux';
+import { elementType } from '@gridsuite/commons-ui/lib/utils/elementType';
+
+function idsToObj(ids, cb) {
```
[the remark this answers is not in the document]
JacquesBorsenberger: it's not dependent of the class, an utilitarian function i can put it back in the class

src/redux/actions.js
```diff
@@ -280,3 +281,12 @@ export function changeDisplayedColumns(displayedColumnsParams) {
         displayedColumnsNamesParams: displayedColumnsParams,
     };
 }
+
+export const FAVORITE_CONTINGENCY_LISTS = 'FAVORITE_CONTINGENCY_LISTS';
+
+export function selectFavoriteContingencyList(favoriteContingencyList) {
```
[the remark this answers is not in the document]
JacquesBorsenberger: done

src/components/contingency-list-selector.js
```diff
@@ -18,109 +18,212 @@ import CheckboxList from './util/checkbox-list';
...
+    );
+
     const [contingencyList, setContingencyList] = useState([]);
 
     const [simulatedContingencyCount, setSimulatedContingencyCount] =
         useState(0);
 
-    const [checkedContingencyListNames, setCheckedContingencyListNames] =
-        useState([]);
+    const [checkedContingencyListIds, setCheckedContingencyListIds] = useState(
```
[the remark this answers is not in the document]
JacquesBorsenberger: done

src/components/contingency-list-selector.js
JacquesBorsenberger: done

PR #385, opened 2021-10-07
Correct spreadsheet refresh pb on load creation

FranckLecuyer:


src/components/network/network.js
```diff
@@ -109,27 +109,40 @@ export default class Network {
...
-        newEquipements.forEach((equipment1) => {
-            const found = currentEquipments.find(
-                (equipment2) => equipment2.id === equipment1.id
-            );
-            if (found === undefined) {
-                currentEquipments.push(equipment1);
-            }
-        });
+        let equipmentsAdded = false;
+        if (this.isResourceFetched(equipmentType)) {
```
sBouzols, line 123: if `isResourceFetched` use an `equipmentType` then we should rename its parameter which is `equipment` and which is not really undestandable. For homogenous code, we could rename equipmentType to equipment here but I prefer the first solution. What do you think of this ?

PR #388, opened 2021-10-11
Delete equipment dialog

FranckLecuyer:


FranckLecuyer:
> There's a problem when we try to delete an equipment with a '/' in the id. Some encoding issues...

I made a correction in study-server and network-modification-server to encode/decode the equipment id,
but there is still a problem in the network-store-server with id containing a '/' ...

src/components/dialogs/equipment-deletion-dialog.js
```diff
@@ -0,0 +1,237 @@
...
+    };
+
+    function handleDeleteEquipmentError(response, messsageId) {
+        const utf8Decoder = new TextDecoder('utf-8');
+        response.body
+            .getReader()
+            .read()
+            .then((value) => {
+                displayErrorMessageWithSnackbar({
+                    errorMessage: utf8Decoder.decode(value.value),
```
etiennehomer, line 95: Why do you need the decoder here ?

PR #392, opened 2021-10-18
Corrections on load creation dialog

niconoir:


src/components/dialogs/load-creation-dialog.js
```diff
@@ -399,7 +404,12 @@ const LoadCreationDialog = ({ open, onClose, network }) => {
                             filterOptions={(options, params) => {
                                 const filtered = filter(options, params);
 
-                                if (params.inputValue !== '') {
+                                if (
+                                    params.inputValue !== '' &&
+                                    !options.find(
```
sBouzols, line 409: `filtered.find` for optimization if we have a lot of options then the find could be greedy

src/components/dialogs/load-creation-dialog.js
```diff
@@ -451,7 +462,12 @@ const LoadCreationDialog = ({ open, onClose, network }) => {
...
```
sBouzols, line 467: same

niconoir: Will be done when component will be mutualized

PR #400, opened 2021-11-04
Modify cursor from grab style to grabbing style when panning or zooming tree

niconoir:


src/components/network-modification-tree.js
```diff
@@ -38,6 +40,14 @@ const NetworkModificationTree = (props) => {
...
     };
 
+    const stylePointerGrab = {
+        cursor: 'grab',
+    };
+
+    const stylePointerGrabbing = {
+        cursor: 'grabbing',
+    };
+
```
SlimaneAmar, line 50: Perhaps it's better to regroup  styles into an object like this :
`    const styles = {
           container: {
            width: '100%',
            height: '100%',
        },
       flow: {
            cursor: isMoving ? 'grabbing' : 'grab',
        },
    };`

src/components/network-modification-tree.js
```diff
@@ -108,12 +126,17 @@ const NetworkModificationTree = (props) => {
             <Box style={style} display="flex" flexDirection="row">
                 <Box flexGrow={1}>
                     <ReactFlow
+                        style={
+                            isMoving ? stylePointerGrabbing : stylePointerGrab
+                        }
```
SlimaneAmar, line 131: styles.flow

src/components/network-modification-tree.js
```diff
@@ -108,12 +126,17 @@ const NetworkModificationTree = (props) => {
             <Box style={style} display="flex" flexDirection="row">
                 <Box flexGrow={1}>
```
SlimaneAmar, line 127: styles.container

package.json
```diff
@@ -35,7 +35,7 @@
         "redux": "^4.0.5",
         "typeface-roboto": "^1.0.0",
         "material-ui-nested-menu-item": "1.0.2",
-        "react-flow-renderer": "^9.6.6",
+        "react-flow-renderer": "^9.6.7",
```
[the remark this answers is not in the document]
niconoir: A bug concerning panning has been fixed in version 9.6.7

src/components/network-modification-tree.js
niconoir: I agree. Fixed.

src/components/network-modification-tree.js
niconoir: Fixed.

src/components/network-modification-tree.js
niconoir: Fixed.

niconoir (commented):
Since there is a bug with the 'grab' cursor in chromium (cf https://bugs.chromium.org/p/chromium/issues/detail?id=45335) which does not seem to be corrected, the behaviour will be slightly different in chromium based browsers and firefox:
- in chromium based browsers, the 'grab' cursor will not be displayed, it will be the 'default' pointer cursor, and the 'grabbing' cursor will be shown when moving the tree.
- in Firefox, the 'grab' cursor will be correctly displayed while not moving and the 'grabbing' cursor will also be displayed when moving the tree. 

This behaviour is the same when moving the network map in the app.

SlimaneAmar (approved):

JacquesBorsenberger (approved):

PR #406, opened 2021-11-22
US #689 Special characters in Elasticsearch for equipment names front part

laurgarn:


FranckLecuyer (approved):

src/utils/rest-api.js
```diff
@@ -325,14 +324,9 @@ export function fetchEquipmentsInfos(studyUuid, searchTerm, useName) {
...
-    let escapedSearchTerm = '*' + luceneEscapeQuery.escape(searchTerm) + '*';
     let urlSearchParams = new URLSearchParams();
-    urlSearchParams.append(
-        'q',
-        useName
-            ? `equipmentName:${escapedSearchTerm}`
-            : `equipmentId:${escapedSearchTerm}`
-    );
+    urlSearchParams.append('userInput', searchTerm);
+    urlSearchParams.append('what', useName ? 'name' : 'id');
```
SlimaneAmar, line 329: `userInput` is too specific to the front, the request  can be made from another application
Maybe keep  `searchTerm`
`What` is not explicit in the back side
Maybe keep  `useName`

laurgarn: "searchTerm" is misleading as we don't, at least for name and id, search for term(s) in the sense of elasticsearch (opposed to 'match' query). And it does not rush to my mind what would be a proper use where we would feed this search with something different from a quite direct user input, even after you remember about this possibility.

Sure "what" is not obvious, but on the other side, if useName is to be used and has value 'false', it is not more obvious what is a not name.
Would 'usesNameOverId' be not too pedantic ?

SlimaneAmar (approved):

PR #407, opened 2021-11-23
Fit and center modification tree

niconoir:


laurgarn (approved):
Noob noticed nothing bad in his review.

src/components/graph/util/center-graph-button.js
```diff
@@ -0,0 +1,33 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+import CenterFocusIcon from '@material-ui/icons/CenterFocusStrong';
+import React, { useCallback } from 'react';
+import { ControlButton, useZoomPanHelper } from 'react-flow-renderer';
+
+const CenterGraphButton = ({ selectedNode }) => {
+    const { setCenter } = useZoomPanHelper();
+
+    const focusNode = useCallback(() => {
```
sBouzols, line 14: If we need to center over the selectedNode without using this ControlButton, for example when creating node, or selecting another node in a list in another component or anyway, how could we call this focusNode function.
Maybe we should externalized it in a utilitary js file and call it here, change which allow to call this function directly from network-modification-tree or others component if needed.
I see that coming...

src/components/graph/util/center-graph-button.js
```diff
@@ -0,0 +1,33 @@
...
+        // if no selected node, center on Root
+        const x = selectedNode ? selectedNode.position.x : 0;
```
sBouzols, line 16: If we have a way to unselect nodes (then to be in a state with no selected node), are we sure that the root node position is always (0,0) ? At first maybe but if it has at least two children, does it have the 0,0 position or a position between its two children (horizontaly at least). Does it depends of draw style chosen ? 
Anyway, I think it's always a bad idea to have raw values like that and hope it will work in all cases.

src/components/graph/util/center-graph-button.js
```diff
@@ -0,0 +1,33 @@
...
+        const y = selectedNode ? selectedNode.position.y : 0;
+        const zoom = 1;
```
sBouzols, line 18: Do we really want to overwrite the zoom value when centering on node ? that's a good question :thinking:

sBouzols (commented):
Code review
No tests (I will before approve)

src/components/graph/util/center-graph-button.js
niconoir: I agree modifying zoom might not be a good idea. I modified the code to keep the current zoom.

src/components/graph/util/center-graph-button.js
niconoir: Currently, the position of the root node is (0,0), calculated via the dagre layout, according to the direction and alignment for rank nodes specified for computation. Changing the computation configuration will affect the position of the root node and thus this part of code might be refactored. Anyway, the behaviour of the "center" button when no node is selected is currently being discussed so this behaviour might change in further versions.

src/components/graph/util/center-graph-button.js
sBouzols: Ok

src/components/graph/util/center-graph-button.js
sBouzols: Ok

sBouzols (approved):
Update with main before merging.
Ok for me !

src/components/graph/util/center-graph-button.js
sBouzols: Nice and simple fix GG!

PR #409, opened 2021-11-25
replace studyNotFound by real reason of failure

AbdelHedhili:

### DIFF
--- a/src/components/study-pane.js
+++ b/src/components/study-pane.js
@@ -264,6 +264,16 @@ const StudyPane = (props) => {
 
     const [studyNotFound, setStudyNotFound] = useState(false);
 
+    const [networkLoadingFail, setNetworkLoadingFail] = useState(false);
+
+    const [geoDataLoadingFail, setGeoDataLoadingFail] = useState(false);
+
+    const [errorMsgId, setErrorMsgId] = useState('');
+
+    const [studyExistsChecked, setStudyExistsChecked] = useState(false);
+
+    const [isNetworkLoadingDone, setIsNetworkLoadingDone] = useState(false);
+
     const [updatedLines, setUpdatedLines] = useState([]);
 
     const [displayedVoltageLevelId, setDisplayedVoltageLevelId] =
@@ -393,6 +403,10 @@ const StudyPane = (props) => {
 
     const updateLoadFlowResult = useCallback(() => {
         fetchStudy(studyUuid).then((study) => {
+            if (study.status === 400 || study.status === 404) {
+                setStudyNotFound(true);
+            }
+            setStudyExistsChecked(true);
             setLoadFlowStatus(getLoadFlowRunningStatus(study.loadFlowStatus));
             setLoadFlowResult(study.loadFlowResult);
         });
@@ -519,7 +533,7 @@ const StudyPane = (props) => {
                     studyUuid,
                     (error) => {
                         console.error(error.message);
-                        setStudyNotFound(true);
+                        setNetworkLoadingFail(true);
                     },
                     dispatch,
                     { equipments: [equipments.lines, equipments.substations] }
@@ -529,7 +543,7 @@ const StudyPane = (props) => {
                     studyUuid,
                     (error) => {
                         console.error(error.message);
-                        setStudyNotFound(true);
+                        setNetworkLoadingFail(true);
                     },
                     dispatch
                 );
@@ -537,6 +551,7 @@ const StudyPane = (props) => {
                 // lazy loading will do the job (no glitches to avoid)
                 dispatch(networkCreated(network));
             }
+            setIsNetworkLoadingDone(true);
         },
         [
             studyUuid,
@@ -586,7 +601,7 @@ const StudyPane = (props) => {
                 })
                 .catch(function (error) {
                     console.error(error.message);
-                    setStudyNotFound(true);
+                    setNetworkLoadingFail(true);
                 });
             // Note: studyUuid don't change
         },
@@ -624,7 +639,8 @@ const StudyPane = (props) => {
             })
             .catch(function (error) {
                 console.error(error.message);
-                setStudyNotFound(true);
+                setWaitingLoadGeoData(false);
+                setGeoDataLoadingFail(true);
             });
         // Note: studyUuid and dispatch don't change
     }, [studyUuid, dispatch]);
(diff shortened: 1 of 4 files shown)


src/translations/fr.json
```diff
@@ -41,6 +41,8 @@
 
     "studyNotFound": "L'étude avec l'identifiant \"{studyUuid}\" n'a pas été trouvée",
     "svgNotFound": "L'image poste n'a pas été trouvée : {error}; {svgUrl} ",
+    "networkLoadingFail": "Le réseau de l'étude avec l'identifiant \"{studyUuid}\" n'a pas pu être chargé",
+    "geoDataLoadingFail": "Les données géographiques du réseau de l'étude avec l'identifiant \"{studyUuid}\" n'ont pas pu être chargée",
```
etiennehomer, line 45: chargéeS

src/components/study-pane.js
```diff
@@ -393,6 +403,10 @@ const StudyPane = (props) => {
 
     const updateLoadFlowResult = useCallback(() => {
         fetchStudy(studyUuid).then((study) => {
+            if (study.status === 400 || study.status === 404) {
+                setStudyNotFound(true);
```
etiennehomer, line 407: Is it ok to check if the study exists here only ?
The LF infos could be moved and we lose the check on study existance.

etiennehomer (changes requested):

src/components/study-pane.js
```diff
@@ -391,6 +401,15 @@ const StudyPane = (props) => {
         }
     }
 
+    const checkStudyExists = useCallback(() => {
+        fetchStudy(studyUuid).then((study) => {
```
sBouzols, line 405: fetchStudy Get the study if exists, it doesn't only check the study existance.
Maybie we should have a `fetchStudyExist` rest api function, which define `{method: HEAD}` when calling backendFetch ? (see recent conversations about checking element existancy with HEAD REST API method)

src/components/study-pane.js
```diff
@@ -709,6 +731,7 @@ const StudyPane = (props) => {
         loadNetwork,
         loadGeoData,
         loadTree,
+        checkStudyExists,
```
sBouzols, line 734: add in upper note that `checkStudyExists` don't change as well. It's alos true for `loadTree`

src/components/study-pane.js
```diff
@@ -264,6 +264,16 @@ const StudyPane = (props) => {
 
     const [studyNotFound, setStudyNotFound] = useState(false);
 
+    const [networkLoadingFail, setNetworkLoadingFail] = useState(false);
```
sBouzols, line 267: why not have an enumerate of `errorTypes` somewhere which consist of , 
```
{ type: STUDY_NOT_FOUND, errorMsgId: 'study_not_found'}
{ type: NETWORK_LOADING_FAILURE, errorMsgId: 'network_loading_failure'}
{ type: GEODATA_LOADING_FAILURE, errorMsgId: 'geodata_loading_failure'}
{ type: NONE, errorMsgId: ''}
...
```
Then you add only one state in here which keep track of the last error :
`const [error, setError] = useState(errorTypes.NONE);`
and use `error.type` or `error.errorMsgId` where needed.
It is then easier to add a new error type, without adding a new state, and without states conditional mess.

src/components/study-pane.js
```diff
@@ -1591,12 +1614,31 @@ const StudyPane = (props) => {
         return waitingLoadGeoData || waitingLoadReport;
     }, [waitingLoadGeoData, waitingLoadReport]);
 
-    if (studyNotFound) {
+    useEffect(() => {
+        if (!waitingLoadGeoData && studyExistsChecked && isNetworkLoadingDone) {
+            if (studyNotFound) {
```
sBouzols, line 1619: This is complex meaning code and not easy upgradable.

src/components/study-pane.js
```diff
@@ -689,6 +710,7 @@ const StudyPane = (props) => {
     useEffect(() => {
         websocketExpectedCloseRef.current = false;
         dispatch(openStudy(studyUuid));
+        checkStudyExists();
         loadNetwork();
```
sBouzols, line 714: Is it interesting to call loadNetwork, loadGeoData, loadTree if study doesn't exists ?
I know we would like to speed up loading with parallelization but, maybe checking Study existence is a prerequisite.

sBouzols (changes requested):
Code review
I think we should set up a more scalable error management system here.
No tests

etiennehomer (approved):
Tests ok.

AbdelHedhili:
Since error are handle this way in all our apps and the states are needed because the errors depend on each other (not using all theses state require a probably much more complex and even less readable code) we agreed to let it that way at least for now.

src/components/study-pane.js
```diff
@@ -689,9 +711,22 @@ const StudyPane = (props) => {
...
-        loadNetwork();
-        loadGeoData();
-        loadTree();
+
+        fetchStudyExists(studyUuid).then((response) => {
+            if (response.status === 400 || response.status === 404) {
+                setStudyNotFound(true);
+                //since the study is not found no need to try to load network and geoData
+                setStudyExistsChecked(true);
+                setIsNetworkLoadingDone(true);
```
sBouzols, line 720: Strange to put this. Shouldn't be : 
`setIsNetworkLoadingDone(false);` ?

src/utils/rest-api.js
```diff
@@ -119,6 +119,13 @@ export function fetchStudy(studyUuid) {
     return backendFetch(fetchStudiesUrl).then((response) => response.json());
 }
 
+export function fetchStudyExists(studyUuid) {
+    console.info(`Fetching study '${studyUuid}' ...`);
```
sBouzols, line 123: adapt log: 
fetch study Uuid **existence** ... ?

src/components/study-pane.js
```diff
@@ -689,9 +711,22 @@ const StudyPane = (props) => {
     useEffect(() => {
         websocketExpectedCloseRef.current = false;
         dispatch(openStudy(studyUuid));
-        loadNetwork();
-        loadGeoData();
-        loadTree();
+
+        fetchStudyExists(studyUuid).then((response) => {
```
sBouzols, line 715: Why not use `checkStudyExists` here ? which is no more used otherwise... It's useEffect dependency has no meaning anymore.
Remove checkStudyExists method or make it return a promise and manage reponse here ?

sBouzols (changes requested):

src/components/study-pane.js
AbdelHedhili: nope because the point is we don't want to do the loading of a network we know doesn't exists

src/components/study-pane.js
AbdelHedhili: forgot to clean up that unused function indeed

sBouzols (approved):
Ok for me.
Test with geo-data-server down OK -> specific message on an existing study

PR #417, opened 2021-12-24
create_substation

AbdelHedhili:

### DIFF
--- a/src/components/dialogs/substation-creation-dialog.js
+++ b/src/components/dialogs/substation-creation-dialog.js
@@ -1,0 +1,273 @@
+/**
+ * Copyright (c) 2021, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+import { InputBase, TextField } from '@material-ui/core';
+import Button from '@material-ui/core/Button';
+import Dialog from '@material-ui/core/Dialog';
+import DialogActions from '@material-ui/core/DialogActions';
+import DialogContent from '@material-ui/core/DialogContent';
+import DialogTitle from '@material-ui/core/DialogTitle';
+import Grid from '@material-ui/core/Grid';
+import { makeStyles, withStyles } from '@material-ui/core/styles';
+import { useSnackbar } from 'notistack';
+import PropTypes from 'prop-types';
+import React, { useCallback, useState } from 'react';
+import { FormattedMessage, useIntl } from 'react-intl';
+import { useParams } from 'react-router-dom';
+import {
+    displayErrorMessageWithSnackbar,
+    useIntlRef,
+} from '../../utils/messages';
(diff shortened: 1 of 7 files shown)


src/components/dialogs/substation-creation-dialog.js
```diff
@@ -0,0 +1,273 @@
...
+        borderColor: 'grey',
+        borderRadius: 4,
+        position: 'relative',
+        border: '1px solid',
+        fontSize: 16,
+        paddingLeft: 14,
+        paddingRight: 14,
+        paddingTop: 18.5,
+        paddingBottom: 18.5,
+    },
```
niconoir, line 64: How did you choose these values ? I tried to modify these values, but it seems to have no effect on the way the component using this style is displayed. I think BootstrapInput can be removed.

src/components/dialogs/substation-creation-dialog.js
```diff
@@ -0,0 +1,273 @@
...
+                                handleChangeCountry(newValue);
+                            }}
+                            input={<BootstrapInput />}
+                            options={Object.keys(countriesList.object())}
+                            style={BootstrapInput.input}
+                            getOptionLabel={(code) => countriesList.get(code)}
+                            renderInput={(props) => (
+                                <TextField
+                                    {...props}
+                                    variant="outlined"
```
niconoir, line 241: `variant` should be `"filled"` and a property `size="small"` should be added in order for the component to be the same height as other fields in the row.

src/components/dialogs/substation-creation-dialog.js
```diff
@@ -0,0 +1,273 @@
...
+                            variant="filled"
+                        />
+                    </Grid>
+                    <Grid item xs={4} align="start">
+                        <Autocomplete
+                            id="select-countries"
+                            onChange={(event, newValue) => {
+                                handleChangeCountry(newValue);
+                            }}
+                            input={<BootstrapInput />}
```
niconoir, line 234: Can be removed.

src/components/dialogs/substation-creation-dialog.js
```diff
@@ -0,0 +1,273 @@
...
+                            options={Object.keys(countriesList.object())}
+                            style={BootstrapInput.input}
```
niconoir, line 236: Can be removed.

niconoir (changes requested):

niconoir (approved):

src/components/dialogs/substation-creation-dialog.js
```diff
@@ -0,0 +1,255 @@
...
+                substationId,
+                substationName,
+                substationCountry
+            )
+                .then(() => {
+                    handleCloseAndClear();
+                })
+                .catch((errorMessage) => {
+                    displayErrorMessageWithSnackbar({
+                        errorMessage: errorMessage,
```
etiennehomer, line 117: errorMessage is empty there. We want the error message from the back nope ?

AbdelHedhili: fixed the problem was backend

etiennehomer (approved):

PR #422, opened 2022-01-11
Implement delete network modification

JacquesBorsenberger:


src/components/graph/menus/modification-list-item.js
```diff
@@ -34,7 +35,18 @@ export const ModificationListItem = ({ modification, ...props }) => {
     return (
         <>
             <ListItem {...props}>
-                <OverflowableText text={getLabel()} />
+                <OverflowableText style={{ flexGrow: '1' }} text={getLabel()} />
+                <IconButton
+                    onClick={() => onDelete(modification.uuid)}
+                    size={'small'}
+                    style={{ alignSelf: 'right' }}
```
sBouzols, line 42: You will certainly need padding or margin later then maybe use a style class for the button and for the `OverflowableText` like in other components like that

src/utils/rest-api.js
```diff
@@ -710,6 +710,25 @@ export function updateTreeNode(node) {
     );
 }
 
+export function deleteModification(studyUuid, node, modificationUuid) {
+    const modificationDeleteUrl =
+        PREFIX_STUDY_QUERIES +
+        '/v1/studies/' +
+        studyUuid +
```
sBouzols, line 717: This came from a `decodeURIComponent` then you should `encodeURIComponent` as well for studyUuid I think.

src/utils/rest-api.js
```diff
@@ -710,6 +710,25 @@ export function updateTreeNode(node) {
...
+        '/nodes/' +
+        encodeURIComponent(node.id) +
+        '/network-modification/' +
```
sBouzols, line 720: /network-modification**s**/ (sync with back modification)

src/components/graph/menus/node-editor.js
```diff
@@ -47,8 +47,11 @@ const NodeEditor = ({ onClose, className }) => {
         if (!selectedNodeUuid) return;
         if (
             !selectedNodeUuidRef.current ||
-            (studyUpdatedForce?.eventData?.headers?.node === selectedNodeUuid &&
-                studyUpdatedForce?.eventData?.headers?.updateType === 'study')
+            (studyUpdatedForce?.eventData?.headers?.updateType ===
+                'nodeUpdated' &&
+                studyUpdatedForce.eventData.headers.nodes.find(
+                    (s) => s === selectedNodeUuid
+                ) !== -1)
```
sBouzols, line 54: replace with `!== undefined` because `find` return undefined if nothing found

src/components/graph/menus/network-modification-node-editor.js
```diff
@@ -74,6 +80,10 @@ const NetworkModificationNodeEditor = ({ selectedNode }) => {
         setOpenNetworkModificationsDialog(false);
     };
 
+    const doDeleteModification = (uuid) => {
+        deleteModification(studyUuid, selectedNode, uuid);
```
sBouzols, line 84: `selectedNode` will not change when deleting a modification... then how `modifications` state is updated after this ?

sBouzols (changes requested):
some comments
no tests for now

src/components/graph/menus/network-modification-node-editor.js
JacquesBorsenberger: there is a notification when the modification is deleted, we update on notification

src/utils/rest-api.js
JacquesBorsenberger: there is already a put without s
    @PutMapping(value = "/studies/{studyUuid}/nodes/{nodeUuid}/network-modification/loads")
I use the same path but with a delete

JacquesBorsenberger: for a change like this it will need a separate PR

sBouzols: I'm telling this because of API rules we defined you know. Here for the new API and a fix for the other one or in another PR for both of them it's ok for me.

src/components/graph/menus/network-modification-node-editor.js
sBouzols: Ok, I see.

src/components/graph/menus/modification-list-item.js
JacquesBorsenberger: done

src/utils/rest-api.js
JacquesBorsenberger: done

src/components/graph/menus/node-editor.js
JacquesBorsenberger: changed to indexOf for more compact code :D

src/utils/rest-api.js
niconoir: I think it would be better to make it right in this PR

sBouzols: @niconoir , if we do this and then merge the feature, it can't work until we update the API from back side.
We have a ticket to apply our API REST best practices in everyf services, it will be done in this context.

niconoir (approved):

PR #429, opened 2022-01-20
Fix favorites contingencies lists not found

FranckLecuyer:


SlimaneAmar (approved):

sBouzols (approved):

sBouzols:
Maybe it's here that we must add a check between `favoriteContingencyListUuids.size` and `mapCont.size` in `fetchContingencyLists` REST API function use ? Just to display a message ? Or showing the user that his favorites ContingencyLists don't exist anymore ? and calling `saveFavorites` to update his lis.

PR #430, opened 2022-01-20
Fix pb with false pmin and pmax values when editing a generator in spreadsheet

FranckLecuyer:


sBouzols (approved):
Code review ok
Can't test for now

PR #437, opened 2022-01-26
[WIP] Node realization and visualization

niconoir:
New proper version after study pane refactoring


niconoir:
Commits merged in main branch within PR: Select first tree model node on study load #438

PR #448, opened 2022-02-10
Fix sld glitch pb when doing a topology modification

FranckLecuyer:


src/components/single-line-diagram-pane.js
```diff
@@ -185,34 +185,41 @@ export function SingleLineDiagramPane({
...
-            } else {
-                updateSld();
+                        ]
+                ) {
+                    historyRef.current.replace(
+                        '/studies/' + encodeURIComponent(studyUuid)
+                    );
+                    setDisplayedSubstationId(null);
+                } else if (
+                    //If the SLD of the deleted voltage level or the SLD of its substation is open, we close it
```
etiennehomer, line 203: This comment was incorrect before your PR. It's more something like:
//If we delete the open voltage level or its substation, we close its SLD

etiennehomer (approved):

PR #456, opened 2022-02-18
Disable modifications dialogs closing when clicking outside dialog

niconoir:


JacquesBorsenberger (approved):
Hum, the same changes on every dialogs... 
I wonder if we could do something about it ;) 
Maybe after dialog refactoring

PR #462, opened 2022-02-21
Fix VL choice menu position

etiennehomer:

### DIFF
--- a/src/components/network-map-tab.js
+++ b/src/components/network-map-tab.js
@@ -231,7 +231,7 @@ export const NetworkMapTab = ({
                 handleClose={closeChoiceVoltageLevelMenu}
                 onClickHandler={choiceVoltageLevel}
                 substation={choiceVoltageLevelsSubstation}
-                position={[position[0] + 200, position[1]]}
+                position={[position[0] + 100, position[1]]}
             />
         );
     }
--- a/src/components/voltage-level-choice.js
+++ b/src/components/voltage-level-choice.js
@@ -16,7 +16,12 @@ import Button from '@material-ui/core/Button';
 import Typography from '@material-ui/core/Typography';
 import { useSelector } from 'react-redux';
 import { getNominalVoltageColor } from '../utils/colors';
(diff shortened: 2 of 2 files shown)


src/components/voltage-level-choice.js
```diff
@@ -57,6 +62,25 @@ const VoltageLevelChoice = ({
 }) => {
     const classes = useStyles();
     const useName = useSelector((state) => state[PARAM_USE_NAME]);
+    const displayMode = useSelector((state) => state[PARAM_MAP_TREE_DISPLAY]);
```
sBouzols, line 65: Ok but does this component which knows about drawers and their status (openned closed) and all ? Maybe you should move this computation of the position in `NetworkMapTab` component instead ?

sBouzols: Keep this component simple. It only needs a position Object which give `top` and a `left` value and that's all.

src/components/network-map-tab.js
```diff
@@ -231,7 +231,7 @@ export const NetworkMapTab = ({
                 handleClose={closeChoiceVoltageLevelMenu}
                 onClickHandler={choiceVoltageLevel}
                 substation={choiceVoltageLevelsSubstation}
-                position={[position[0] + 200, position[1]]}
+                position={[position[0] + 100, position[1]]}
```
sBouzols, line 234: Is this a magical number ? seems like it is :stuck_out_tongue:

sBouzols (changes requested):
Code Review : Maybe move this fix in NetworkMapTab
No tests for now

src/components/network-map-tab.js
etiennehomer: I changed it with a constant

src/components/voltage-level-choice.js
etiennehomer: Yes that's right. I moved this code to NetworkMapTap

FranckLecuyer (approved):

sBouzols (approved):
Code review ok
Test ok

PR #471, opened 2022-02-25
Fit tree view on load

etiennehomer:

### DIFF
--- a/src/components/network-modification-tree.js
+++ b/src/components/network-modification-tree.js
@@ -282,9 +282,9 @@ const NetworkModificationTree = ({ treeModel, studyMapTreeDisplay }) => {
         }
     }, [isModificationsDrawerOpen]);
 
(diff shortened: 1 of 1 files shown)


src/components/network-modification-tree.js
```diff
@@ -282,6 +282,10 @@ const NetworkModificationTree = ({ treeModel, studyMapTreeDisplay }) => {
         }
     }, [isModificationsDrawerOpen]);
 
+    const onLoad = (reactFlowInstance) => {
```
FranckLecuyer, line 285: use 'useCallback', like onMove, onPaneClick ...

FranckLecuyer (changes requested):

src/components/network-modification-tree.js
etiennehomer: done

FranckLecuyer (approved):

PR #473, opened 2022-03-01
Fix useNodeData hook to avoid multiple unneccessary fetches

niconoir:


JacquesBorsenberger (approved):

src/components/study-container.js
```diff
@@ -79,11 +79,14 @@ export function useNodeData(
         const headers = studyUpdatedForce?.eventData?.headers;
         const updateType = headers && headers['updateType'];
         const node = headers && headers['node'];
+        const nodes = headers && headers['nodes'];
         const isUpdateForUs =
             lastUpdateRef.current !== studyUpdatedForce &&
             updateType &&
-            (node === undefined || node === nodeUuid) &&
-            invalidations.find((e) => updateType === e) !== -1;
+            ((node === undefined && nodes === undefined) ||
```
sBouzols, line 86: it begins to be a mess in here. Maybe we should create a function `forceUpdate(studyUpdatedForce)` which manage all this code and conditions with comments for each case.

PR #478, opened 2022-03-04
minor fixes on searchElementDialog

AbdelHedhili:

### DIFF
--- a/src/components/dialogs/input-hooks.js
+++ b/src/components/dialogs/input-hooks.js
@@ -93,9 +93,9 @@ const useStyles = makeStyles((theme) => ({
     },
 }));
 
-export const gridItem = (field, size = 6) => {
+export const gridItem = (field, size = 6, align = 'start') => {
     return (
-        <Grid item xs={size} align="start">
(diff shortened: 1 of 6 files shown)


AbdelHedhili:
Any helpers ? I can't for the life of me align the button on the right so that it's aligned with inputs underneath.

src/components/dialogs/input-hooks.js
```diff
@@ -454,7 +454,7 @@ const DELAY = 1000;
 export const useSearchEquipmentField = ({ handleOpenSearchDialog, label }) => {
```
sBouzols, line 454: strange naming. Seems like it's a very generic custom hook. Maybe `useButton` is a better name. Then change `handleOpenSearchDialog` by `handleClick` maybe ?

sBouzols (commented):
Code Review ok, but we must change the hook name.
No Test

src/components/dialogs/input-hooks.js
AbdelHedhili: done

sBouzols (approved):

PR #484, opened 2022-03-08
Fix pb when no built node in study

FranckLecuyer:


laurgarn (approved):

src/components/graph/util/model-functions.js
```diff
@@ -5,13 +5,25 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-export function getIdFirstNodeOfType(elements, nodeType) {
-    if (elements.type === nodeType) {
-        return elements.id;
+export function getFirstNodeOfType(elements, nodeType, buildStatus) {
+    if (
+        elements.type === nodeType &&
+        (elements.buildStatus === undefined ||
```
niconoir, line 11: Maybe change to `buildStatus === undefined` instead of `elements.buildStatus === undefined` here, to get first node of given type if no buildStatus parameter is passed.

niconoir (changes requested):

niconoir (approved):

PR #493, opened 2022-03-28
[Fix] Do not load network if working node and study uuid are not defined

niconoir:


laurgarn (approved):

src/components/study-container.js
```diff
@@ -199,24 +199,29 @@ export function StudyContainer({ view, onChangeTab }) {
...
-                        console.error(error.message);
-                        setNetworkLoadingFailMessage(error.message);
-                        //setIsNetworkPending(false);
-                    },
-                    dispatch,
-                    { equipments: [equipments.lines, equipments.substations] }
-                );
-            } else {
-                if (workingNode !== null) {
+            if (workingNode && studyUuid) {
```
JacquesBorsenberger, line 202: I would prefer something like if (!workingNode || !studyUuid) return;
as we don't do anything and add a level to indentation make the code less clear

JacquesBorsenberger (approved):

src/components/study-container.js
niconoir: Modified.

PR #501, opened 2022-04-07
Material UI v5 migration

sBouzols:


src/components/dialogs/connectivity-edition.js
```diff
@@ -232,7 +233,11 @@ const ConnectivityEdition = ({
                         id="bus"
                         disabled={!voltageLevel}
                         options={busOrBusbarSectionOptions}
-                        getOptionLabel={(bbs) => bbs?.id || ''}
+                        getOptionLabel={(bbs) => {
```
laurgarn, line 236: Isn't it the same as before, but with more code ?

laurgarn (approved):

src/components/dialogs/connectivity-edition.js
sBouzols: Not exactly, we can't set `busOrBusbarSection` to null to clear the field anymore. Then I replaced null with ' ' (empty string) and manage this case in getOptionLabel (which is called in this case) to clear the field.

JacquesBorsenberger (approved):

PR #504, opened 2022-04-08
remove useless drawer shift

etiennehomer:


JacquesBorsenberger (approved):

src/components/network-modification-tree.js
```diff
@@ -155,8 +155,9 @@ const NetworkModificationTree = ({ treeModel, studyMapTreeDisplay }) => {
     );
 
     const onPaneClick = useCallback(() => {
-        setSelectedNode(undefined);
-    }, []);
+        setSelectedNode(null);
```
niconoir, line 158: Maybe it's not necessary to keep the selectedNode as a state in this component. Instead of having:
`const [selectedNode, setSelectedNode] = useState(null);` 
maybe use
`const selectedNode = useSelector((state) => state.selectedTreeNode);`
and remove all calls to `setSelectNode` that are redundant with the `dispatch(selectTreeNode(...))`

niconoir (changes requested):

src/components/network-modification-tree.js
etiennehomer: done

niconoir (approved):

PR #519, opened 2022-04-15
New parameter used to modify generators and batteries's power display

EstherDarkish:


src/components/parameters.js
```diff
@@ -429,6 +439,36 @@ const Parameters = ({ showParameters, hideParameters, user }) => {
         );
     }
 
+    function NetworkParameters() {
```
JacquesBorsenberger, line 442: I think one day we should explode parameter.js to 1 component per tab (and per Internal components); but that is for another PR I think

JacquesBorsenberger, line 442: I think one day we should explode parameter.js to 1 component per tab (and per Internal components); but that is for another PR I think

JacquesBorsenberger (approved):

src/components/network/equipment-table.js
```diff
@@ -53,6 +55,8 @@ export const EquipmentTable = ({
 
     useEffect(() => setLineEdit({}), [tableDefinition]);
 
+    const fluxConvention = useSelector((state) => state[PARAM_FLUX_CONVENTION]);
```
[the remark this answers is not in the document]
EstherDarkish: Ok, I moved it to the NetworkTable.

archendel (approved):

jonenst (approved):

PR #521, opened 2022-04-22
fix the size of geolocation button

YenguiSeddik:


src/components/app-top-bar.js
```diff
@@ -78,8 +78,11 @@ const CustomSuffixRenderer = ({ props, element }) => {
         element.type === EQUIPMENT_TYPE.VOLTAGE_LEVEL.name
     )
         return (
-            <IconButton onClick={(e) => enterOnSubstationCB(e, element)}>
-                <GpsFixedIcon />
+            <IconButton
+                onClick={(e) => enterOnSubstationCB(e, element)}
+                size={'small'}
```
JacquesBorsenberger, line 83: sill larger (42px vs 36px), there is padding; can you use one of the classes of equipmentClasses or put the padding to 0 (if there is a class for tag in equipmentClasses that is better as, we won't have to keep the style in 2 places

JacquesBorsenberger (changes requested):

sBouzols (approved):
Code Review Ok
Tests ok

JacquesBorsenberger (approved):

PR #526, opened 2022-04-26
fix input hooks adornment

JacquesBorsenberger:
Signed-off-by: jacques borsenberger <jacques.borsenberger@rte-france.com>


src/components/util/text-field-with-adornment.js
```diff
@@ -79,13 +80,14 @@ const TextFieldWithAdornment = (props) => {
...
+        classes.inputRight,
     ]);
 
     const withStartAdornmentText = useCallback(() => {
-        return value
+        return value !== '' || isFocused
             ? {
-                  startAdornment: isFocused && getTextAdornment('start'),
+                  startAdornment:
+                      value && isFocused && getTextAdornment('start'),
```
niconoir, line 90: Should be only `getTextAdornment('start')` here, no need to check `value && isFocused`.
Should be `endAdornment: value && clearable && getClearAdornment('end')` on the following line

niconoir (changes requested):

src/components/util/text-field-with-adornment.js
JacquesBorsenberger: done

niconoir (approved):

sBouzols (approved):
code Review OK
Tests Done :
* each form of hypothesis creation checked OK

PR #536, opened 2022-05-11
Add case export on node

sBouzols:


FranckLecuyer (approved):

src/translations/en.json
```diff
@@ -253,6 +253,7 @@
     "createNetworkModificationNode": "Create a network modification node",
     "createModelNode": "Create a model node",
     "removeNode": "Remove node",
+    "exportNode": "Export node",
```
SlimaneAmar, line 256: It is not the node that is exported but the associated network

sBouzols: New proposal

SlimaneAmar: Export network ?

sBouzols: Export Case PO approval

PR #543, opened 2022-05-16
Table overflow x

EstherDarkish:
Replaces the VirtualizedTable in the Spreadsheet by a MultiGrid. This allows a more flexible display of data's columns, and allows an horizontal scrolling on the table.
Adds the possibility to "lock" columns to always displays them on the left of the screen.


src/components/network/network-table.js
```diff
@@ -144,31 +358,385 @@ const NetworkTable = (props) => {
...
+                        <LockOpenIcon
+                            className={classes.columnConfigOpenLock}
+                        />
+                    );
+                }
+            }
+        }
+    };
+
+    const setSort = useCallback(
```
laurgarn, line 398: May be we could ask the PO whether we want to be able to return to 'not sorted' and a have a three state there (unsorted -> sorted -> sorted reverse -> not sorted) ?

src/components/network/network-table.js
```diff
@@ -144,31 +358,385 @@ const NetworkTable = (props) => {
...
+                    props.workingNode?.id,
+                    groovyCr
+                ).then((response) => {
+                    if (response.ok) {
+                        Object.entries(lineEdit.newValues).forEach(
+                            ([key, cr]) => {
+                                rowData[key] = cr.value;
+                            }
+                        );
+                        // TODO When the data is saved, we should force an update of the table's data. As is, it takes too long to update.
```
laurgarn, line 493: Personally agree.

src/components/network/network-table.js
```diff
@@ -144,31 +358,385 @@ const NetworkTable = (props) => {
...
+                    </div>
+                );
+            } else {
+                return (
+                    <div key={key} style={style}>
+                        <div className={classes.editCell}>
+                            <IconButton
+                                size={'small'}
+                                disabled={
+                                    lineEdit !== undefined &&
```
laurgarn, line 533: Could just use `lineEdit?.id !== undefined`

src/components/network/network-table.js
```diff
@@ -144,31 +358,385 @@ const NetworkTable = (props) => {
...
+                </div>
+            );
+        },
+        [classes.tableCell, formatCell]
+    );
+
+    const registerChangeRequest = useCallback(
+        (data, changeCmd, value) => {
+            // save original value, dont erase if exists
+            if (!lineEdit.oldValues[data.dataKey])
```
laurgarn, line 586: Beware of the zero ?

src/components/network/network-table.js
```diff
@@ -144,31 +358,385 @@ const NetworkTable = (props) => {
...
+        let generatedTableColumns = TABLES_DEFINITION_INDEXES.get(tabIndex)
+            .columns.filter((c) => {
+                return selectedColumnsNames.has(c.id);
+            })
+            .map((c) => {
+                let column = {
+                    ...c,
+                    label: intl.formatMessage({ id: c.id }),
+                    locked: lockedColumnsNames.has(c.id),
+                    columnWidth: c.columnWidth
```
laurgarn, line 645: Coud use `c.columnWidth ?? MIN_COLUMN_WIDTH`

src/components/network/network-table.js
```diff
@@ -144,31 +358,385 @@ const NetworkTable = (props) => {
...
+                    ) =>
+                        editableCellRender(
+                            rowData,
+                            columnDefinition,
+                            c.changeCmd,
+                            key,
+                            style
+                        );
+                } else {
+                    column.cellRenderer = (
```
laurgarn, line 664: Could just shorten to column.`cellRenderer = defaultCellRender`

src/components/network/network-table.js
```diff
@@ -144,31 +358,385 @@ const NetworkTable = (props) => {
...
+                        style
+                    ) =>
+                        defaultCellRender(
+                            rowData,
+                            columnDefinition,
+                            key,
+                            style
+                        );
+                }
+                delete column.changeCmd;
```
laurgarn, line 677: Why remove column.changeCmd ? Not obvious to me.

src/components/network/network-table.js
```diff
@@ -144,31 +358,385 @@ const NetworkTable = (props) => {
...
+        if (generatedTableColumns.length > 0 && isEditColumnVisible()) {
+            generatedTableColumns.unshift({
+                locked: true,
+                editColumn: true,
+                columnWidth: HEADER_CELL_WIDTH,
+                cellRenderer: (rowData, columnDefinition, key, style) =>
+                    editCellRender(rowData, columnDefinition, key, style),
+            });
+        }
+        generatedTableColumns.headerCellRender = (
```
laurgarn, line 700: Could just have `generatedTableColumns.headerCellRender = headerCellRender;` too

src/components/network/network-table.js
```diff
@@ -285,6 +900,45 @@ const NetworkTable = (props) => {
...
+        const localisedTabName = intl.formatMessage({ id: tabName });
+        return localisedTabName
+            .trim()
+            .replace(/[\\/:"*?<>|\s]/g, '-') // Removes the filesystem sensible characters
+            .substring(0, 27); // Best practice : limits the filename size to 31 characters (27+'.csv')
+    }, [intl, tabIndex]);
+
+    const getCSVColumnNames = () => {
+        let tempHeaders = [];
+        const columns = generateTableColumns(tabIndex).filter(
```
laurgarn, line 914: Why filter out edit column ?

src/components/network/network-table.js
```diff
@@ -285,6 +900,45 @@ const NetworkTable = (props) => {
...
+            (c) => !c.editColumn
+        );
+        columns.forEach((col) => {
```
laurgarn, line 917: Could use .map instead of temp.forEach and push

src/components/network/network-table.js
EstherDarkish: If we don't filter it out, then it becomes an empty column in the CSV file.

src/components/network/config-tables.js
```diff
@@ -74,10 +80,12 @@ export const TABLES_DEFINITIONS = {
             {
                 id: 'ID',
                 dataKey: 'id',
+                columnWidth: 220,
```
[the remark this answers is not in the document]
EstherDarkish: True. I replaced each one with a new const.

src/components/network/config-tables.js
```diff
@@ -722,22 +675,27 @@ export const TABLES_DEFINITIONS = {
...
             },
             {
                 id: 'Name',
                 dataKey: 'name',
+                columnWidth: 220,
             },
             {
                 id: 'ConvertersMode',
                 dataKey: 'convertersMode',
+                columnWidth: 340,
```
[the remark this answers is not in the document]
EstherDarkish: Same, new constant here.

src/components/network/network-table.js
```diff
@@ -75,40 +204,125 @@ const NetworkTable = (props) => {
...
+            for (let i = 0; i < datasourceRows.length; i++) {
+                const row = datasourceRows[i];
+                if (!rowFilter || filter(row)) {
+                    result.push(row);
+                }
+            }
+
+            function compareValue(a, b, isNumeric, reverse) {
+                const mult = reverse ? 1 : -1;
+                if (a === undefined && b === undefined) return 0;
```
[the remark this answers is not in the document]
EstherDarkish: Yes, I replaced it.

src/components/network/network-table.js
EstherDarkish: Changed it to a three state sort. It is better in the case where you start to edit something, click on the table name and "lose" your line in the 2000 rows of the table. If you can undo the sort, then your line comes back where it was and you can resume your edit.

src/components/network/network-table.js
EstherDarkish: I reused some existing code to do this, but by removing it, the code is more complex.
So I changed it, stopped removing it, and now the code is simpler.

src/components/network/network-table.js
EstherDarkish: Yes, done.

src/components/network/network-table.js
EstherDarkish: Done.

laurgarn (approved):

src/components/network/network-table.js
```diff
@@ -75,40 +204,125 @@ const NetworkTable = (props) => {
...
+            );
+        },
+        [rowFilter, selectedDataKey]
+    );
+
+    const formatCell = useCallback(
+        (rowData, columnDefinition) => {
+            let value = rowData[columnDefinition.dataKey];
+            if (columnDefinition.cellDataGetter) {
+                value = columnDefinition.cellDataGetter(rowData, props.network);
```
jonenst, line 271: keep the same API cellDataGetter as the previous one ?

EstherDarkish: Can you elaborate ? I do not understand.


By "previous one", you mean the old definition in config-tables.js ? Before modifying it, I scoured the code and checked with @niconoir to make sure there would not be negative side effects.


Or maybe you mean that we could use an else to not execute line 269 if columnDefinition.cellDataGetter is defined ?

jonenst (approved):

PR #550, opened 2022-05-20
make node name unique

YenguiSeddik:


src/components/network-modification-tree-pane.js
```diff
@@ -138,8 +140,21 @@ export const NetworkModificationTreePane = ({
...
+                        errorMessage: errorMessage,
+                        enqueueSnackbar: enqueueSnackbar,
+                        headerMessage: {
+                            headerMessageId: 'NodeCreateError',
+                            intlRef: intlRef,
+                        },
+                    });
+                });
+
             createTreeNode(studyUuid, element.id, insertMode, {
```
sBouzols, line 156: You should call `createTreeNode` in the `then()` proc of the `getUniqueNodeName`.
Because here you don't wait for the response to create the new node with the correct name. Then you could even remove unecessary `newName` state

src/components/dialogs/input-hooks.js
```diff
@@ -819,3 +820,58 @@ export const useExpandableValues = ({
...
+                    );
+                    setIsValidName(false);
+                }
+            });
+        },
+        [studyUuid, intl]
+    );
+
+    useEffect(() => {
+        if (name === '' && !timer.current) return; // initial render
```
sBouzols, line 866: This is the same if statement 2 times

src/components/dialogs/input-hooks.js
```diff
@@ -819,3 +820,58 @@ export const useExpandableValues = ({
...
+        if (name === '' && !timer.current) {
+            return;
+        }
+        clearTimeout(timer.current);
+        setIsValidName(true);
```
sBouzols, line 871: by default it's valid ? it should be false by default then true if HEAD request return not found I think

sBouzols (changes requested):
some remarks will test later

sBouzols (changes requested):

SlimaneAmar (changes requested):

src/components/dialogs/input-hooks.js
```diff
@@ -767,3 +770,100 @@ export const useExpandableValues = ({
...
+        adornment,
+        error,
+        triggerReset,
+    });
+
+    const validName = useCallback(
+        (name) => {
+            if (name !== defaultValue) {
+                isNodeExists(studyUuid, name).then((response) => {
+                    if (response.status === 403) {
```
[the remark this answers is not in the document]
YenguiSeddik: ok, done.

SlimaneAmar (approved):

sBouzols (approved):
Code Review OK
Tests OK
* rename with already existing name
* add/insert nodes OK
Console Warning check OK

PR #554, opened 2022-05-23
Add values to transformer spreadsheet

bendaamerahm:
- Add values under  Two Windings Transformers spreadsheet
- Add values under  Three Windings Transformers spreadsheet
- create new param to columns definition (boolean: true | false)
- handle the case when data is boolean (true/false) by rendring a new Element (checkbox)


src/translations/fr.json
```diff
@@ -394,5 +395,12 @@
...
 
     "LoadStudyAndParentsInfoError": "Erreur lors de la récupération du nom et chemin de l'étude en cours",
-    "LogOnlySingleNode": "Log noeud seul"
+    "LogOnlySingleNode": "Log noeud seul",
+
+    "RegulationMode" : "MODE DE RÉGLAGE DEPHASEUR",
+    "RegulatingPhase" : "DEPHASEUR EN RÉGLAGE",
+    "RegulationValue" : "CONSIGNE DU COURANT (A) OU DE TRANSIT (MW)",
+    "LoadTapChangingCapabilities" : "RÉGLEUR EN CHARGE",
+    "RegulatingRatio" : "RÉGLEUR  EN RÉGLAGE",
```
sBouzols, line 404: remove one whitespace here

src/components/network/network-table.js
```diff
@@ -671,6 +713,9 @@ const NetworkTable = (props) => {
                 };
                 if (c.changeCmd !== undefined) {
                     column.cellRenderer = editableCellRender;
+                } else if (column.boolean) {
+                    // check if param boolean is true then render a boolean cell(checkbox)
```
sBouzols, line 717: seems obvious comment but ok

src/components/network/network-table.js
```diff
@@ -599,6 +599,48 @@ const NetworkTable = (props) => {
...
+                                columnDefinition.canBeInvalidated &&
+                                props.loadFlowStatus !== RunningStatus.SUCCEED,
+                        })}
+                    >
+                        {isChecked !== undefined && (
+                            <Checkbox
+                                style={{ margin: '-10%', cursor: 'initial' }}
+                                checked={isChecked}
+                                disableRipple={true} //to change by using dynamic value when handling events
+                                //onClick={handleCheckboxClick} #TODO handle change event later
```
sBouzols, line 628: Remove dead code

src/components/network/config-tables.js
```diff
@@ -214,6 +214,27 @@ export const TABLES_DEFINITIONS = {
                 fractionDigits: 1,
                 canBeInvalidated: true,
             },
+            {
+                id: 'LoadTapChangingCapabilities',
+                dataKey: 'loadTapChangingCapabilities',
+                boolean: true, // new param to indicate that this cell data is boolean(true/false)
```
sBouzols, line 220: It's a new param for now then it will not :smile: 
I think you should remove this comment, we understand that this column if a boolean type.

sBouzols, line 220: You don't need a `cellDataGetter` here for `LoadTapChangingCapabilities`?

JacquesBorsenberger (approved):
not sure adding a new value for boolean type is the right approach, maybe we should use un columnType (for numeric, boolean...) but it is ok.

sBouzols (approved):
Code Review OK
Tests front ok
No warnings
Ok for the front, but hard to approve without data
No tests for sort, filter, or csv export because no data in columns

bendaamerahm:
> Code Review OK Tests front ok No warnings Ok for the front, but hard to approve without data No tests for sort, filter, or csv export because no data in columns

im working on adding data to json from back, ill informe you when data is ready to use

bendaamerahm:
> not sure adding a new value for boolean type is the right approach, maybe we should use un columnType (for numeric, boolean...) but it is ok.

ill do the refactoring for this change later  to avoid conflicts

sBouzols (changes requested):

sBouzols (approved):

PR #557, opened 2022-05-24
US-823 sequels

laurgarn:
* Force icon in our autocomplete hook
 * remove filledTextField from line-split fields
 * show the right busbar/bus id on re-edit
 * avoids showing twice the available busbars/busses
 * add a workaround for simple case to show split line ends on re-edit
 * change to more explicit label for the voltage level where to split
 * change BusbarOrNodeID english translation


src/components/dialogs/line-split-with-voltage-level-dialog.js
```diff
@@ -548,11 +539,22 @@ const LineSplitWithVoltageLevelDialog = ({
...
-            : lineToDivide.voltageLevelId2;
-        const mayVl = allVoltageLevelOptions.filter((vl) => vl.id === vlId);
-        if (mayVl && mayVl.length) return mayVl[0].name;
+        let vlId = '';
+        if (typeof lineToDivide === 'object') {
+            vlId = isFirst
+                ? lineToDivide.voltageLevelId1
+                : lineToDivide.voltageLevelId2;
+        } else if (isFirst) {
+            const mayLine = lineOptions.filter((l) => l?.id === newLine1Id);
```
sBouzols, line 548: Maybe use find instead of filter in all this function to avoid access to array[0], a find by id which is unique by the way.

laurgarn: done

sBouzols (approved):

PR #567, opened 2022-06-02
fix style of select menu under N-K tab

bendaamerahm:
fix style of select menu under N-K tab


FranckLecuyer (approved):
Review: OK
Tested: OK

PR #568, opened 2022-06-02
fix a bug in voltage level creation dialog

YenguiSeddik:


src/components/dialogs/voltage-level-creation-dialog.js
```diff
@@ -139,7 +139,7 @@ const BusBarConnexion = ({
             defaultValue && defaultValue.fromBBS
                 ? getBusbarSectionById(fieldProps, defaultValue?.fromBBS)
                 : '',
-        values: fieldProps,
+        values: fieldProps.filter((val) => val?.id !== ''),
```
laurgarn, line 142: '?.' could return undefined, so filtering against empty string may be too precise, no ?

laurgarn (approved):

PR #574, opened 2022-06-03
Network modification tree visual design

Meklo:


src/components/graph/nodes/network-modification-node.js
```diff
@@ -11,37 +11,70 @@ import makeStyles from '@mui/styles/makeStyles';
...
         color: theme.palette.primary.contrastText,
         '&:hover': {
             background: theme.palette.primary.main,
         },
+        overflow: 'hidden',
     },
     networkModification: {
-        variant: 'outlined',
-        background: theme.palette.primary.light,
+        background: '#c5c8d1',
```
EstherDarkish, line 31: In dark mode this works fine, but in light mode, the display is really hard to see.
Maybe use the theme.palette to help with contrast ?

`background: theme.palette.text.secondary,`
works fine here.

src/components/graph/nodes/network-modification-node.js
```diff
@@ -11,37 +11,70 @@ import makeStyles from '@mui/styles/makeStyles';
...
-    margin: {
-        marginLeft: '10px',
+    buildBannerOK: {
+        display: 'flex',
+        height: '100%',
+        width: '15%',
+        position: 'absolute',
+        top: '0px',
+        left: '0px',
+        background: '#74a358',
```
EstherDarkish, line 59: Why not use the same green as in the minimap ?
#74a358 vs #70d136

Also, maybe use properties for the colors, instead of direct values.
This way, we only need to change the properties in one place to adapt the colors of the whole tree.

EstherDarkish: There is also the possibility to use the theme colors here. This way, the contrast is correct in dark and light mode.

`background: theme.palette.success.main,`
works fine here.

src/components/network-modification-tree.js
```diff
@@ -222,7 +249,18 @@ const NetworkModificationTree = ({
                     showInteractive={false}
                 >
                     <CenterGraphButton selectedNode={selectedNode} />
+                    <ControlButton
+                        onClick={() =>
+                            setIsMinimapOpen(isMinimapOpen ? false : true)
```
EstherDarkish, line 254: Use 
`setIsMinimapOpen(!isMinimapOpen)`
instead ?

src/components/study-container.js
```diff
@@ -277,6 +278,14 @@ export function StudyContainer({ view, onChangeTab }) {
                 networkModificationTreeModel.setTreeElements(tree);
                 networkModificationTreeModel.updateLayout();
 
+                fetchCaseInfos(studyUuid)
+                    .then((res) => {
+                        if (res) {
+                            networkModificationTreeModel.setCaseName(res);
+                        }
+                    })
+                    .catch((err) => console.log(err.message));
```
EstherDarkish, line 287: use
`console.error`
instead ?

src/components/network-modification-tree.js
JacquesBorsenberger: or setIsMinimapOpen( oldval => !oldVal);  and make it a callback with no dependency ? 
const toggleMinimapOpen = useCallback( () => setIsMinimapOpen( oldval => !oldVal));

src/components/study-container.js
JacquesBorsenberger: i think we should give the user a better feedback if there is an error, (in snack-bar for example)

src/components/graph/nodes/network-modification-node.js
Meklo: I created constants to hold the banners background color.
We've carefully chosen the green tint thus I think it's best to keep the same from one theme to another

src/components/graph/nodes/network-modification-node.js
Meklo: Agreed `background: theme.palette.text.secondary,` works well in both themes for the nodes background

src/components/network-modification-tree.js
Meklo: I went for the toggle callback method

src/components/network-modification-tree.js
```diff
@@ -64,12 +67,34 @@ const NetworkModificationTree = ({
...
+            }
+            switch (node.data.buildStatus) {
+                case 'BUILT':
+                    return '#70d136';
+                case 'BUILT_INVALID':
+                    return '#9196a1';
+                case 'NOT_BUILT':
+                    return '#9196a1';
+                default:
+                    return '#9196a1';
```
EstherDarkish, line 94: Why not using const INVALID_NODE_BANNER_COLOR = '#9196a1'; from the other file ?

EstherDarkish (approved):

JacquesBorsenberger (approved):

PR #575, opened 2022-06-03
Adds new parameters to the LoadFlow settings

EstherDarkish:
Adds missing parameters in LoadFlowParameters.
- New entity fields
- Liquibase migration


src/components/parameters.js
```diff
@@ -319,6 +348,27 @@ const Parameters = ({ showParameters, hideParameters, user }) => {
         );
     }
 
+    function MakeAdvancedParameterButton(prop, label, callback) {
```
FranckLecuyer, line 351: Maybe rename 'prop' parameter name to 'showAdvancedParameter' ?

FranckLecuyer (changes requested):

src/components/parameters.js
EstherDarkish: Ok. Renamed it to "showOpenIcon"

FranckLecuyer (approved):

PR #591, opened 2022-06-14
show selected node in topbar

bendaamerahm:
* show selected node in topbar
* no default root node pre-selected
* synchronise the nodename in topbar with all events on the  selected node (update node name , delete node ...)

### DIFF
--- a/src/components/app-top-bar.js
+++ b/src/components/app-top-bar.js
@@ -153,6 +153,8 @@ const AppTopBar = ({ user, tabIndex, onChangeTab, userManager }) => {
 
     const workingNode = useSelector((state) => state.workingTreeNode);
 
+    const selectedTreeNode = useSelector((state) => state.selectedTreeNode);
+
     const [showParameters, setShowParameters] = useState(false);
     const [, showVoltageLevel, showSubstation] = useSingleLineDiagram();
     // Equipments search bar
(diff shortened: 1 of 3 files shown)


JacquesBorsenberger (approved):

src/redux/reducer.js
```diff
@@ -354,29 +356,13 @@ export const reducer = createReducer(initialState, {
 });
 
 function synchWorkingNodeAndSelectedNode(state) {
```
sBouzols, line 358: This function doesn't sync workingNode anymore, rename it or I don't know... Maybe it's because workingNode will be removed soon ?

src/components/app-top-bar.js
```diff
@@ -153,6 +153,8 @@ const AppTopBar = ({ user, tabIndex, onChangeTab, userManager }) => {
 
     const workingNode = useSelector((state) => state.workingTreeNode);
```
sBouzols, line 154: Strange to keep workingNode here then.
It means that we show the selectedNode in topBar but we use the workingNode for the search feature... it's a mess.

sBouzols (changes requested):

src/redux/reducer.js
```diff
@@ -354,29 +356,13 @@ export const reducer = createReducer(initialState, {
...
-            buildStatus: workingNode?.buildStatus,
-        };
-    }
     const selectedNode = state.networkModificationTreeModel?.treeElements.find(
         (entry) => entry?.id === state.selectedTreeNode?.id
     );
     // handle the case of selectedNode not in the TreeModel anymore.
     if (selectedNode === undefined) state.selectedTreeNode = null;
+    else {
+        const selectedNodeCopie = Object.assign({}, selectedNode);
```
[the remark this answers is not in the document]
bendaamerahm: my bad , ill use spread operator instead

src/redux/reducer.js
bendaamerahm: ill keep managing synchronization of both kind of nodes (selected, working) until we decide witch solution is better to do the refactor (example: using one type of node in global scope of the app)

JacquesBorsenberger (approved):

PR #592, opened 2022-06-15
fix:handles non boolean fields change

laurgarn:


src/components/dialogs/export-dialog.js
```diff
@@ -71,6 +71,14 @@ const useMeta = (metasAsArray) => {
         });
     };
 
+    const onFieldChange = (event, paramName) => {
```
JacquesBorsenberger, line 74: why not use a callback ?
and use prevInst instead of inst ?

src/components/dialogs/export-dialog.js
```diff
@@ -71,6 +71,14 @@ const useMeta = (metasAsArray) => {
...
+        setInst((prevInst) => {
+            const nextInst = { ...inst };
```
JacquesBorsenberger, line 76: why use inst instead of prevInst ?

JacquesBorsenberger (approved):

PR #606, opened 2022-06-20
fix:build status not refreshed as when loaded

laurgarn:


sBouzols (approved):
Code Review OK
Test OK

PR #611, opened 2022-06-21
Fix icons appearence in topbar when node with long name is selected

Walid-Sahnoun:


src/components/app-top-bar.js
```diff
@@ -220,6 +220,11 @@ const AppTopBar = ({ user, tabIndex, onChangeTab, userManager }) => {
     function hideParameters() {
         setShowParameters(false);
     }
+    function formatTreeNodeLabel(selectedTreeNode) {
```
etiennehomer, line 223: Maybe give the string selectedTreeNode.data.label to this method and not the object

PR #642, opened 2022-06-28
Refactor TreeModel and optimization of modifications management

sBouzols:
refactor(NetworkModificationNodeEditor): Simplify use of currentTreeNode, and manage fetchModifications procs only when needed

refactor(NetworkModificationTreeModel): Add parentNodeUuid for reverse course, variantId, reportUuid, modificationGroupUuid in the Tree model to avoid useless fetchTreeNode call only to get modificationGroupUuid for example before getting modifications


FranckLecuyer (changes requested):
Code review:OK
Tests: PB
When we open a switch in an SLD of a built node, the modification is not shown immediately in the modifications list: we  need to switch to another node and return to the previous node to see it ...

sBouzols:
@FranckLecuyer : it was fixed really recently in https://github.com/gridsuite/study-server/pull/231, could you check again with your study-server updated ?

FranckLecuyer (approved):

PR #645, opened 2022-06-28
US-1054 more useful message on tree load 404.

laurgarn:


archendel (approved):

src/utils/rest-api.js
```diff
@@ -774,11 +774,15 @@ export function fetchNetworkModificationTree(studyUuid) {
...
-    );
+    return backendFetch(url, { method: 'get' }).then((response) => {
+        if (response.ok) {
+            return response.json();
+        } else {
+            return response.json().then((text) => {
+                return Promise.reject(text);
+            });
+        }
+    });
```
AbdelHedhili, line 785: why this change ? It's exactly the same or i'm misssing something ?

src/components/study-container.js
```diff
@@ -316,16 +316,26 @@ export function StudyContainer({ view, onChangeTab }) {
...
-                    },
-                })
-            )
+            .catch((errorMessage) => {
+                if (errorMessage.status === 404)
+                    displayErrorMessageWithSnackbar({
+                        errorMessage: '',
+                        enqueueSnackbar: enqueueSnackbar,
+                        headerMessage: {
+                            headerMessageId: 'StudyAwfulStateRecreate',
```
AbdelHedhili, line 325: not a big fan of the naming "awful" but whatever

AbdelHedhili (changes requested):

src/utils/rest-api.js
laurgarn: Not exactly. For "exception path", content was awaited as flat string and is in fact json, as seen in browser console. This json is to be exploited in the caller.

src/components/study-container.js
laurgarn: Not very proud of, but it is internal. If you are inspired, feel free to propose another name ;-)

archendel: StudyUnrecoverableStateRecreate could be a better property name

AbdelHedhili: +1

AbdelHedhili (approved):

SlimaneAmar:
Use `useSnackMessage`

PR #646, opened 2022-06-28
select parent node when selected node is deleted

YenguiSeddik:


src/redux/reducer.js
```diff
@@ -348,13 +356,18 @@ export const reducer = createReducer(initialState, {
...
-    else {
+    // handle the case of currentTreeNode is deleted
+    if (currentNode === undefined) {
+        const parentNode =
+            state.networkModificationTreeModel?.treeElements.find(
+                (treeElement) => treeElement?.id === parentNodeId
+            );
+        state.currentTreeNode = { ...parentNode };
+    } else {
         state.currentTreeNode = { ...currentNode };
```
sBouzols, line 371: Add a comment that we need to overwrite state.currentTreeNode line 371 to consider label change for example.

src/redux/reducer.js
```diff
@@ -348,13 +356,18 @@ export const reducer = createReducer(initialState, {
     },
 });
 
-function synchCurrentTreeNode(state) {
+function synchCurrentTreeNode(state, parentNodeId = undefined) {
```
sBouzols, line 359: rename to parentNodeUuid

src/redux/reducer.js
```diff
@@ -174,10 +174,18 @@ export const reducer = createReducer(initialState, {
         if (state.networkModificationTreeModel) {
             let newModel =
                 state.networkModificationTreeModel.newSharedForUpdate();
+            let parentNode;
```
sBouzols, line 177: rename to parentNodeUuid

src/redux/reducer.js
```diff
@@ -174,10 +174,18 @@ export const reducer = createReducer(initialState, {
...
+            if (
+                action.networkModificationTreeNodes.includes(
+                    state.currentTreeNode?.id
+                )
+            ) {
+                parentNode = state.currentTreeNode?.data?.parentNodeUuid;
```
sBouzols, line 183: Add a TODO comment here :
Today we manage action.networkModificationTreeNodes which size is always 1 and then to delete one node at a time.
If tomorrow we need to delete multiple nodes, we need to check that the parentNode here isn't in the action.networkModificationTreeNodes list

src/redux/reducer.js
sBouzols: Maybe we could even factorize more this function
function synchCurrentTreeNode(state,nextCurrentNodeUuid) {
// check if its in model and get Node objetc
// overwrite state.currentTreeNode

use synchCurrentTreeNode(state, parentNodeUuid) in NETWORK_MODIFICATION_TREE_NODES_REMOVED
if currentTreeNode is included in nodes deleted
otherwise
use synchCurrentTreeNode(state, undefined)

But ok will see later

sBouzols (changes requested):
Some renamings

src/redux/reducer.js
YenguiSeddik: Ok, done.

src/redux/reducer.js
YenguiSeddik: Ok, done.

bendaamerahm (approved):
Code review OK 
Test OK

src/redux/reducer.js
YenguiSeddik: Ok, done.

PR #660, opened 2022-06-30
Disables the OverloadedLines panel when the current node is not built.

EstherDarkish:
Disables the OverloadedLines panel when the current node is not built.


sBouzols (approved):
Code review OK
Tests OK

etiennehomer (approved):

PR #662, opened 2022-07-01
fix:promise to reject with response if not ok

laurgarn:


src/utils/rest-api.js
```diff
@@ -789,9 +789,7 @@ export function fetchNetworkModificationTree(studyUuid) {
         if (response.ok) {
             return response.json();
         } else {
-            return response.json().then((text) => {
-                return Promise.reject(text);
-            });
+            return Promise.reject(response);
```
bendaamerahm, line 792: it will be great if we refactor the code to something like that 
.then((response) =>
        response.ok
            ? response.json()
            : response.text().then((text) => Promise.reject(text))
    );
but it's ok

niconoir (approved):

bendaamerahm (approved):

src/utils/rest-api.js
laurgarn: Nope, I want the response in the caller, in order to get to the 404 status on which I want a different behaviour. Body is secondary.

bendaamerahm: okey i understand

bendaamerahm (approved):

src/utils/rest-api.js
jonenst: we should find a global solution for this. The body often contains useful information about the error when our services manage to produce it. Of course sometimes the error is so bad that we can't even produce an error message, and in this case it shouldn't crash, but in other cases it should use the body to display the error information. So i wouldn't say it's secondary. However the code is hard to write and test to do all of this correctly, hence the need for a more global solution.. The simple solution in this PR is a good first step though

PR #665, opened 2022-07-01
Fixes the bug preventing a SLD of an unbuilt node to be closed.

EstherDarkish:
When a SLD is open and the currentNode is not built, there was a bug that prevented the "close" icon of the SLD to close the SLD window.


bendaamerahm (approved):

sBouzols (approved):
Code Review Ok
Tests OK : with Nicolas, Kevin, Charly
Multiple test cases OK

PR #672, opened 2022-07-04
Add alphabetical sorting to picking modal related to directory items

Meklo:


sBouzols:
waiting for the next version of commons-ui then.

src/components/directory-item-selector.js
```diff
@@ -64,12 +65,22 @@ const DirectoryItemSelector = (props) => {
         });
     };
 
+    function sortHandlingDirectories(a, b) {
+        if (a.type === elementType.DIRECTORY && a.type !== b.type) {
```
sBouzols, line 69: Add a comment that this code is needed to put nodes with children (aka directories) first then other types of nodes (leaves) alphabetically sorted.

src/components/directory-item-selector.js
```diff
@@ -28,6 +28,7 @@ const DirectoryItemSelector = (props) => {
             const newNode = {
                 id: newData.elementUuid,
                 name: newData.elementName,
+                type: newData.type,
```
sBouzols, line 31: do not pass type to TreeViewFinder nodes data which will not be consume in there.
Maybe check what you want by checking if children is undefined or not.

sBouzols (changes requested):

src/components/directory-item-selector.js
Meklo: Done

src/components/directory-item-selector.js
Meklo: Also done

bendaamerahm (approved):

PR #673, opened 2022-07-04
Fix: disable updateNetwork and updateGeoData on currentNode renaming

sBouzols:

### DIFF
--- a/src/components/graph/util/model-functions.js
+++ b/src/components/graph/util/model-functions.js
@@ -75,8 +75,8 @@ export function isNodeReadOnly(node) {
 
 export function isNodeBuilt(node) {
     if (!node) return false;
-    if (node?.type === 'ROOT') return true;
-    return node?.data?.buildStatus === 'BUILT';
+    if (node.type === 'ROOT') return true;
+    return node.data?.buildStatus === 'BUILT';
 }
 
 export function isSameNode(node1, node2) {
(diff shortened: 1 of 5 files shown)


bendaamerahm (approved):
test ok 
=> that resolve the actual problem but we need to figure out a better solution later (handle specific Notification by action)

bendaamerahm (approved):
Code review: OK
Test: 
- renaming node don't trigger fetch of GeoData and Network : OK

src/components/graph/util/model-functions.js
```diff
@@ -74,6 +74,18 @@ export function isNodeReadOnly(node) {
 }
 
 export function isNodeBuilt(node) {
+    if (!node) return false;
     if (node?.type === 'ROOT') return true;
```
laurgarn, line 78: You could get rid of the ? in the ?. due to the test you added above

src/components/study-container.js
```diff
@@ -317,20 +322,18 @@ export function StudyContainer({ view, onChangeTab }) {
...
-                            : getFirstNodeOfType(tree, 'ROOT')
-                    )
-                );
-
+                let firstSelectedNode =
+                    getFirstNodeOfType(tree, 'NETWORK_MODIFICATION', 'BUILT') ||
+                    getFirstNodeOfType(tree, 'ROOT');
+
+                // To get positions we must get the node from the model class
+                let ModelFirstSelectedNode = {
```
laurgarn, line 330: Why let and not const ? (can not remember a language where constness cost less typing than mutability, though :-/ )

laurgarn (approved):

PR #681, opened 2022-07-06
auto scroll to selected row

YenguiSeddik:
When there is scrollbar in the spreadsheet, the selected equipment  is displayed at the top with different background color .
If there is no scrollbar, the selected equipment just change the background color.

### DIFF
--- a/src/components/network/equipment-table.js
+++ b/src/components/network/equipment-table.js
@@ -13,9 +13,7 @@ export const EquipmentTable = (props) => {
 
     const [fixedColumnsCount, setFixedColumnsCount] = useState(0);
     const theme = useTheme();
-    const [scrollTopIndex, setScrollTopIndex] = useState(
-        props.scrollTop !== -1
-    );
+    const [scrollTopLock, setScrollTopLock] = useState(true);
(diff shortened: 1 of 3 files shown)


src/components/network/equipment-table.js
```diff
@@ -6,11 +6,18 @@ import {
...
     ROW_HEIGHT,
 } from './config-tables';
+import { useTheme } from '@mui/material/styles';
 
 export const EquipmentTable = (props) => {
     const gridRef = useRef();
 
     const [fixedColumnsCount, setFixedColumnsCount] = useState(0);
+    const theme = useTheme();
+    const [scrollTopIndex, setScrollTopIndex] = useState(
```
Meklo, line 16: Maybe rename it to something like `scrollTopLock` since its usage is more about locking/unlocking the scroll and the value of the index doesn't seem to be consumed

Meklo (approved):
Except for the naming of the state everything is in order. 
Approved to don't block the merge with a request change since I'm not here next week

sBouzols (approved):
Code Review OK
Tests OK with you

PR #689, opened 2022-07-07
hide edition on hypothesis

bendaamerahm:
hide temporarily the possibility of editing hypothesis throught the update button


sBouzols (approved):
Code Review OK

PR #691, opened 2022-07-07
Us 978 tableur corrections diverses

dbraquart:
This US contains various fixes, then please refer to US description.

A special point of attention: to reset the filter value (not the regexp itself, but the string/input value), I had to use a "useRef": please tell me if there is a better way.

### DIFF
--- a/src/components/network/network-table.js
+++ b/src/components/network/network-table.js
@@ -406,20 +406,16 @@ const NetworkTable = (props) => {
         ]
     );
 
-    const onTabChange = useCallback(
-        (index) => {
-            console.debug('onTabChange from ' + tabIndex + ' to ' + index);
-            // when we change Tab, we dont want to keep/apply the search criteria
-            if (
-                !searchTextInput.current.value ||
-                searchTextInput.current.value !== ''
-            ) {
-                searchTextInput.current.value = '';
-                setFilterValue('');
-            }
-        },
-        [tabIndex]
-    );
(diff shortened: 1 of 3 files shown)


Meklo (approved):
Code Review: OK
Tests: OK

src/components/network/network-table.js
```diff
@@ -830,9 +910,8 @@ const NetworkTable = (props) => {
         return generatedTableColumns;
     }
 
-    function renderTable() {
+    function renderTable(rows) {
```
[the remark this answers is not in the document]
dbraquart: rows.length was needed **before** renderTable(), to disable the CSV export button when there is no data/rows.
So I compute rows, then I can disable the CSV button if the size is 0, and then I pass the rows to renderTable().

src/components/network/network-table.js
```diff
@@ -360,6 +406,21 @@ const NetworkTable = (props) => {
         ]
     );
 
+    const onTabChange = useCallback(
+        (index) => {
+            console.debug('onTabChange from ' + tabIndex + ' to ' + index);
```
YenguiSeddik, line 411: I don't think we need to log the index of the tab when we change it

dbraquart: ok, I will remote it

YenguiSeddik (approved):
Code review: OK
Tests: OK

PR #697, opened 2022-07-08
fix: Search equipment Infos with 3 chars doesn't work sometimes 

sBouzols:
if searchTerm with 3 char then close and reopen the searchBar then with the same searchTerms it doesn't proc fetch


YenguiSeddik (approved):
Test: Ok

PR #701, opened 2022-07-12
fixing name label of the SJB

anas-drissi:


bendaamerahm (approved):
code review: ok

src/translations/fr.json
```diff
@@ -382,7 +382,7 @@
     "Substation": "Site",
     "BusBarSections": "Sections de jeu de barre",
     "BusBarSectionID": "ID",
-    "BusBarSectionName": "Nom",
+    "BusBarSectionName": "Nom (optionnel)",
```
sBouzols, line 385: Maybe check if you could use `NameOptional` key and remove `BusBarSectionName` because it seems like it is the same string :)

sBouzols (changes requested):
If the string already exists then use the existing key

src/translations/fr.json
anas-drissi: It's fixed.

sBouzols (approved):
Code Review OK

PR #704, opened 2022-07-13
fetch voltage level from back when creating network modif

AbdelHedhili:

### DIFF
--- a/src/components/dialogs/input-hooks.js
+++ b/src/components/dialogs/input-hooks.js
@@ -284,30 +284,30 @@ export const useConnectivityValue = ({
     }, [inputForm.toggleClear]);
 
     useEffect(() => {
-        if (fetchedVoltageLevelOptions) {
-            fetchedVoltageLevelOptions.then((values) =>
-                setVoltageLevelOptions(
-                    values.sort((a, b) => a.id.localeCompare(b.id))
-                )
-            );
-        }
+        if (!fetchedVoltageLevelOptions) return;
+
+        fetchedVoltageLevelOptions.then((values) =>
+            setVoltageLevelOptions(
+                values.sort((a, b) => a.id.localeCompare(b.id))
+            )
+        );
     }, [fetchedVoltageLevelOptions]);
 
     useEffect(() => {
-        if (voltageLevelOptions) {
-            setConnectivity({
-                voltageLevel: voltageLevelIdDefaultValue
-                    ? voltageLevelOptions.find(
-                          (value) => value.id === voltageLevelIdDefaultValue
-                      )
-                    : null,
-                busOrBusbarSection: busOrBusbarSectionIdDefaultValue
-                    ? {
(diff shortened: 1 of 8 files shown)


src/components/graph/menus/network-modification-node-editor.js
```diff
@@ -353,6 +356,18 @@ const NetworkModificationNodeEditor = () => {
...
+        if (currentTreeNode !== undefined) {
+            fetchVoltageLevels(studyUuid, currentTreeNode.id).then((array) => {
+                setVoltageLevelOptions(
+                    array.sort((a, b) => a.id.localeCompare(b.id))
+                );
+            });
+        } else {
+            setVoltageLevelOptions([]);
+        }
+    }, [studyUuid, currentTreeNode]);
```
sBouzols, line 369: We don't need to fetch VLs for form dialogs components when navigating between nodes I think. `withVLs` is called when needed, i.e when the concerned dialog is openned.
I adapted the code of `withEquipmentModificationOptions`, which works like a charm, for `withSubstations` and `withLines` I think it's better to be homogenous here.

sBouzols (changes requested):

src/components/graph/menus/network-modification-node-editor.js
AbdelHedhili: fixed

src/components/dialogs/input-hooks.js
```diff
@@ -283,18 +284,30 @@ export const useConnectivityValue = ({
...
-                      (value) => value.id === voltageLevelIdDefaultValue
-                  )
-                : null,
-            busOrBusbarSection: busOrBusbarSectionIdDefaultValue
-                ? {
-                      id: busOrBusbarSectionIdDefaultValue,
-                  }
-                : null,
-        });
+        if (fetchedVoltageLevelOptions) {
```
sBouzols, line 287: this seems interesting to test `fetchedVoltageLevelOptions` definition here.
Maybe prefer to do
`if(!fetchedVoltageLevelOptions) return;`
to avoid savage indentation in the following code (dedication @laurgarn )

src/components/dialogs/input-hooks.js
```diff
@@ -283,18 +284,30 @@ export const useConnectivityValue = ({
...
+            fetchedVoltageLevelOptions.then((values) =>
+                setVoltageLevelOptions(
+                    values.sort((a, b) => a.id.localeCompare(b.id))
+                )
+            );
+        }
+    }, [fetchedVoltageLevelOptions]);
+
+    useEffect(() => {
+        if (voltageLevelOptions) {
```
sBouzols, line 297: Maybe prefer to do
`if(!voltageLevelOptions) return;`
to avoid savage indentation in the following code (dedication @laurgarn )

sBouzols (approved):
Code review OK
will test

PR #706, opened 2022-07-13
Display an error when loading study case name, if case no longer exists

FranckLecuyer:


src/components/study-container.js
```diff
@@ -309,13 +309,12 @@ export function StudyContainer({ view, onChangeTab }) {
                         }
                     })
                     .catch((err) => {
-                        console.error(err.message);
+                        console.error(err);
                         displayErrorMessageWithSnackbar({
-                            errorMessage: err.message,
+                            errorMessage: err,
```
sBouzols, line 314: You could use `useSnackMessage` custom hook here

sBouzols (approved):
a little comment to improve code simplicity

src/components/study-container.js
FranckLecuyer: Done

niconoir (approved):

PR #713, opened 2022-07-19
US-1146 add autocomplete to delete dialog. Usable

laurgarn:


src/components/dialogs/equipment-deletion-dialog.js
```diff
@@ -59,24 +70,72 @@ const useStyles = makeStyles(() => ({
...
+            false,
+            true,
+            equipmentType
+        )
+            .then((infos) => {
+                let autoItems = makeItems(infos, useNameLocal);
+                setEquipmentsFound(autoItems);
+            })
+            .catch((errorMessage) =>
+                displayErrorMessageWithSnackbar({
```
FranckLecuyer, line 105: use useSnackMessage

src/components/dialogs/input-hooks.js
```diff
@@ -412,25 +413,27 @@ export const useAutocompleteField = ({
...
     const validationRef = useRef();
     validationRef.current = validation;
 
     useEffect(() => {
-        if (defaultValue) {
-            setValue(defaultValue);
-        }
-    }, [defaultValue]);
+        setValue(defaultValue);
+    }, [defaultValue, values]);
```
FranckLecuyer, line 425: why values in the dependency array ?

src/components/dialogs/input-hooks.js
```diff
@@ -412,25 +413,27 @@ export const useAutocompleteField = ({
...
             const res = validateField(value, validationRef.current);
             setError(res?.errorMsgId);
             return !res.error;
         }
-        inputForm.addValidation(id ? id : label, validate);
-    }, [label, validation, inputForm, value, id]);
+        if (inputForm) {
+            inputForm.addValidation(id ? id : label, validate);
+        }
+    }, [label, validation, inputForm, value, id, error, errorMsg]);
```
FranckLecuyer, line 436: why error and errorMsg in the dependency array ?

FranckLecuyer (changes requested):

FranckLecuyer (commented):
Tests: the search of VSC or LCC converter stations does not work

src/components/dialogs/input-hooks.js
laurgarn: Removed

src/components/dialogs/equipment-deletion-dialog.js
laurgarn: Done

src/components/dialogs/input-hooks.js
FranckLecuyer: ok

FranckLecuyer (approved):

src/components/dialogs/input-hooks.js
```diff
@@ -412,24 +413,26 @@ export const useAutocompleteField = ({
     previousValue,
     loading = false,
 }) => {
+    const [presentedOptions, setPresentedOptions] = useState([]);
+    const [isLoading, setIsLoading] = useState(loading);
```
archendel, line 417: I think it's weird here, "loading" is now the default value of "isLoading", I don't understand how it's supposed to work

src/components/dialogs/input-hooks.js
```diff
@@ -442,7 +445,35 @@ export const useAutocompleteField = ({
         }
     }, [selectedValue]);
 
+    useEffect(() => {
```
archendel, line 448: Misunderstanding this useEffect, values can now be a promise or a function ?
Can we make this code more clear ?

archendel (changes requested):
Test : OK
Code : Some clarifications needed

src/components/util/search-matching-equipments.js
```diff
@@ -26,40 +27,42 @@ export const useSearchMatchingEquipments = (
     const [useNameLocal] = useParameterState(PARAM_USE_NAME);
 
     const searchMatchingEquipments = useCallback(
-        (searchTerm) => {
+        (searchTerm, sooner = false) => {
```
archendel, line 30: "sooner" variable name is unclear, change it to give some clue about it's behaviour

src/components/util/search-matching-equipments.js
```diff
@@ -26,40 +27,42 @@ export const useSearchMatchingEquipments = (
...
+                        inUpstreamBuiltParentNode,
+                        equipmentType
+                    )
+                        .then((infos) => {
+                            if (searchTerm === lastSearchTermRef.current) {
+                                setEquipmentsFound(
+                                    makeItems(infos, useNameLocal)
+                                );
+                            } // else ignore results of outdated fetch
+                            timer.current = undefined;
```
archendel, line 50: why ? What is the difference with "clearTimeout(timer.current)" ?

src/components/dialogs/voltage-level-creation-dialog.js
```diff
@@ -353,7 +354,7 @@ const VoltageLevelCreationDialog = ({
         formProps: filledTextField,
         values: substationOptions?.sort(compareById),
         allowNewValue: true,
-        getLabel: getId,
+        getLabel: getIdOrSelf,
```
archendel, line 357: leave a comment to explain why we now need to get Id sometimes, and Self other times

src/components/dialogs/equipment-deletion-dialog.js
```diff
@@ -34,20 +34,28 @@ const equipmentTypes = [
     'BATTERY',
     'DANGLING_LINE',
     'HVDC_LINE',
-    'LCC_CONVERTER_STATION',
-    'VSC_CONVERTER_STATION',
+    'HVDC_CONVERTER_STATION',
```
archendel, line 37: have some equipment types disappeared ?

src/components/dialogs/input-hooks.js
```diff
@@ -442,51 +476,131 @@ export const useAutocompleteField = ({
...
+        (term) => {
+            if (!onSearchTermChange) return;
+            setIsLoading(true);
+            setExpanded(true);
+            onSearchTermChange(term, true);
+        },
+        [onSearchTermChange]
+    );
+
+    const myOnOpen = useCallback(() => {
```
archendel, line 505: change function name to something more clear

src/components/dialogs/input-hooks.js
```diff
@@ -442,51 +476,131 @@ export const useAutocompleteField = ({
...
+                }
+                return term;
+            });
+        },
+        [values, minCharsBeforeSearch, onSearchTermChange]
+    );
+
     const field = useMemo(() => {
+        const handleKeyDown = (e) => {
+            if (e.ctrlKey && e.code === 'Space') {
```
archendel, line 535: this key combination seems to be inspired from eclipse usage, I don't think it's well known our users
Also, is it a behaviour we need ?

src/components/dialogs/input-hooks.js
```diff
@@ -442,51 +476,131 @@ export const useAutocompleteField = ({
...
+                handleForcedSearch(userStr);
+            }
+        };
+
+        const optionEqualsToValue = (option, input) =>
```
archendel, line 540: function returning Boolean should start with "is"

src/components/dialogs/input-hooks.js
```diff
@@ -397,13 +399,39 @@ export const useConnectivityValue = ({
 
 const filter = createFilterOptions();
 
+const QUESTIONABLE_SIZE = 1000;
```
archendel, line 402: change constant name ?

src/components/dialogs/equipment-deletion-dialog.js
```diff
@@ -59,65 +67,91 @@ const useStyles = makeStyles(() => ({
...
+        useSearchMatchingEquipments(
+            studyUuid,
+            currentNodeUuid,
+            true,
+            equipmentType,
+            makeItems
+        );
+
+    const [equipmentOrId, equipmentField] = useAutocompleteField({
+        allowNewValue: true,
```
archendel, line 96: should we allow new values for a deletion form ?

archendel (changes requested):
some comments, and "loading" functionality still looks weird in my opinion

src/components/util/search-matching-equipments.js
laurgarn: What is unclear about adjective 'sooner', especially given a default value of 'false' ?
https://www.google.com/search?channel=fs&q=traduire+sooner

src/components/util/search-matching-equipments.js
laurgarn: Uh, it was just for printf-debugging, it allowed to know if a request was on air.
I am removing it.
For the difference with `clearTimeout(timer.current)`, I guess there is more than one, actually

src/components/dialogs/input-hooks.js
laurgarn: Ja wohl !

src/components/dialogs/input-hooks.js
laurgarn: This is a convention for boolean values too, and should not be applied blindly, though it could have been applied profitably to what is named freeSolo not far from here

src/components/dialogs/input-hooks.js
laurgarn: May be, but change it to what ? I came using this name, because it was the size for which it is better to ask again the new term having previous one as prefix.

src/components/dialogs/equipment-deletion-dialog.js
laurgarn: The parameter in useAutocompleteField is called allowNewValue.
Though I think a better name could be found, I am not sure it would suit any use case.

src/components/dialogs/input-hooks.js
laurgarn: It seems used in some MS software for code-completion : is this fashionable this way ? When not editing (a text), it is usually used to help entering short data.
So user can use it or not be aware, it is not dangerous. And as a developper (that tests the software also by using it) I am happy not to have to remember ids for each equipment type in every network situation format and geographical area.

src/components/dialogs/voltage-level-creation-dialog.js
laurgarn: Done

src/components/util/search-matching-equipments.js
archendel: my point here is the purpose of the variable, what does "sooner" means for "searchMatchingEquipment" ? It should search sooner ? Sooner than what ?

src/components/dialogs/input-hooks.js
archendel: I understand it's useful as a developer, but is it a behaviour that was asked for production ? Hidden undocumented behaviour can lead to problems, I'm not sure we should create shortcuts that will go to production to help us test the app

src/components/dialogs/input-hooks.js
archendel: I just thought it would improve readability here

src/components/dialogs/equipment-deletion-dialog.js
archendel: my point here is we are deleting items that do exist, and will be listed as options. What is the purpose to let the user write it's own input ?

laurgarn: It is a point requested by the PO.
It can be explain like that : this autocomplete is an help, it doesn't have to forbid you to create a suppression of an equipment created previously in the node or one of its ancestor, and when the node is not built, the equipment is not available in the list.

archendel: alright I understand

src/components/util/search-matching-equipments.js
laurgarn: sooner than later -> do not wait much before really asking

src/components/dialogs/input-hooks.js
laurgarn: Yes, values can now be either an array, a Promise or a function.
Can I make it clearer ? With pure code, no.
With a comment reminding that Promise.resolve given a parameter that is not a Promise creates a Promise that will call `then` with that value. Can not see a better way. What do you propose ?

archendel (approved):

PR #721, opened 2022-07-21
Add new editable fields in the LOADS tab in the spreadsheet : name, type, p0, q0.

dbraquart:
We also want to do it using the 'Update Load' existing hypothesis rather than a generic groovy update.

### DIFF
--- a/src/components/network/config-tables.js
+++ b/src/components/network/config-tables.js
@@ -6,8 +6,14 @@
  */
 
 import { equipments } from './network-equipments';
-import { NumericalField, TapChangerSelector } from './equipment-table-editors';
+import {
+    NumericalField,
+    NameField,
+    EnumField,
+    TapChangerSelector,
+} from './equipment-table-editors';
 import { FluxConventions } from '../dialogs/parameters/network-parameters';
+import { LOAD_TYPES } from './constants';
 
 const nominalVoltage = (network, voltageLevelId) => {
     return network.getVoltageLevel(voltageLevelId)?.nominalVoltage;
@@ -664,6 +670,7 @@ export const TABLES_DEFINITIONS = {
         index: 6,
         name: 'Loads',
         resource: equipments.loads,
+        modifiableEquipmentType: 'load',
         columns: [
             {
                 id: 'ID',
@@ -674,10 +681,21 @@ export const TABLES_DEFINITIONS = {
                 id: 'Name',
                 dataKey: 'name',
                 columnWidth: MEDIUM_COLUMN_WIDTH,
+                changeCmd: "equipment.setName('{}')\n",
+                editor: ({ equipment, ...props }) =>
+                    NameField({
+                        ...props,
+                    }),
             },
             {
                 id: 'LoadType',
                 dataKey: 'type',
+                changeCmd: 'equipment.setLoadType(LoadType.{})\n',
+                editor: ({ equipment, ...props }) =>
+                    EnumField({
+                        enumList: LOAD_TYPES.slice(1), // dont use/display the first empty entry
+                        ...props,
+                    }),
             },
             {
                 id: 'VoltageLevelId',
@@ -711,12 +729,26 @@ export const TABLES_DEFINITIONS = {
                 dataKey: 'p0',
                 numeric: true,
                 fractionDigits: 1,
+                changeCmd: 'equipment.setP0({})\n',
+                editor: ({ equipment, ...props }) =>
+                    NumericalField({
+                        min: undefined,
+                        max: undefined,
+                        ...props,
+                    }),
             },
             {
                 id: 'ConstantReactivePower',
                 dataKey: 'q0',
                 numeric: true,
                 fractionDigits: 1,
+                changeCmd: 'equipment.setQ0({})\n',
+                editor: ({ equipment, ...props }) =>
+                    NumericalField({
+                        min: undefined,
+                        max: undefined,
+                        ...props,
+                    }),
             },
         ],
     },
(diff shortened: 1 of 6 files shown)


src/components/network/config-tables.js
```diff
@@ -711,12 +729,26 @@ export const TABLES_DEFINITIONS = {
                 dataKey: 'p0',
                 numeric: true,
                 fractionDigits: 1,
+                changeCmd: 'equipment.setP0({})\n',
+                editor: ({ equipment, ...props }) =>
```
archendel, line 733: undefined props don't need to be specified, I think you can simplify this to :
editor: NumericalField,

src/components/network/config-tables.js
```diff
@@ -674,10 +681,21 @@ export const TABLES_DEFINITIONS = {
                 id: 'Name',
                 dataKey: 'name',
                 columnWidth: MEDIUM_COLUMN_WIDTH,
+                changeCmd: "equipment.setName('{}')\n",
+                editor: ({ equipment, ...props }) =>
```
archendel, line 685: I think you can simplify to :
editor: NameField,

src/components/network/config-tables.js
```diff
@@ -711,12 +729,26 @@ export const TABLES_DEFINITIONS = {
...
+                        ...props,
+                    }),
             },
             {
                 id: 'ConstantReactivePower',
                 dataKey: 'q0',
                 numeric: true,
                 fractionDigits: 1,
+                changeCmd: 'equipment.setQ0({})\n',
+                editor: ({ equipment, ...props }) =>
```
archendel, line 746: same as above

src/components/network/equipment-table-editors.js
```diff
@@ -86,8 +89,69 @@ export const NumericalField = ({
...
+        <TextField
+            defaultValue={defaultValue}
+            onChange={(ev) => setter(ev.target.value)}
+            {...props}
+            size={'small'}
+            margin={'none'}
+            style={{ ...style, padding: 0 }}
+            inputProps={{
+                style: {
+                    textAlign: 'center',
```
archendel, line 124: I think it looks better without centering the text, it's my opinion though

src/components/network/equipment-table-editors.js
```diff
@@ -86,8 +89,69 @@ export const NumericalField = ({
...
+                    fontSize: 'small',
+                    flexGrow: 1,
```
archendel, line 126: what is the flexGrow useful for ?

src/components/network/equipment-table-editors.js
```diff
@@ -86,8 +89,69 @@ export const NumericalField = ({
...
+};
+
+export const EnumField = ({ enumList, setter, defaultValue, ...props }) => {
+    return (
+        <Select
+            defaultValue={defaultValue || ''}
+            onChange={(ev) => setter(ev.target.value)}
+            size={'medium'}
+            margin={'none'}
+            MenuProps={{
```
archendel, line 140: why is it needed ?

archendel (changes requested):
Test : OK
Code review : some small changes

src/components/network/equipment-table-editors.js
dbraquart: it is useless. I copy/paste all the style from another TextField Editor to look like the same. I will remove it (in both)

src/components/network/equipment-table-editors.js
dbraquart: I dont know. I copy/paste this part from another Select Editor to look like the same. To be verified

src/components/network/config-tables.js
dbraquart: yes, right. I will simplify as requested

src/components/network/config-tables.js
dbraquart: yes, right. I will simplify as requested

src/components/network/config-tables.js
dbraquart: yes, right. I will simplify as requested

src/components/network/equipment-table-editors.js
dbraquart: I got the answer: it was to limit the height of the select one opened. I dont need it in my "enum" case, so I'll remove it.

archendel (approved):
One small visual trouble, the dropdown list field seems to be slightly bigger than the textfields

src/components/network/equipment-table-editors.js
```diff
@@ -82,12 +90,76 @@ export const NumericalField = ({
...
+                    step: 'any',
                 }}
             />
-        </Tooltip>
+        );
+    }
+
+    return (
+        <div>
+            {min !== undefined || max !== undefined ? (
```
EstherDarkish, line 127: If there is no "min" values, there should not be "min: " in the tooltip.

src/components/network/equipment-table-editors.js
```diff
@@ -82,12 +90,76 @@ export const NumericalField = ({
...
+    );
+};
+
+export const NameField = ({ defaultValue, setter, style, ...props }) => {
+    const [error, setError] = useState(false);
+
+    const validateChange = useCallback(
+        (ev) => {
+            const newVal = ev.target.value;
+            setError(newVal.length === 0);
```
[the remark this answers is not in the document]
dbraquart: yes in deed. cf previous commit ffa9e9e

src/components/network/equipment-table-editors.js
dbraquart: agreed. to be changed

EstherDarkish: Maybe change the new code to not have a lot of ternary operators ?

dbraquart: agreed. to be changed in next commit

PR #727, opened 2022-07-22
Display of nodes with very long text

anas-drissi:


src/components/graph/nodes/network-modification-node.js
```diff
@@ -81,6 +82,9 @@ const useStyles = makeStyles((theme) => ({
     margin: {
         marginLeft: theme.spacing(1.25),
     },
+    tooltip: {
+        maxWidth: '1000px',
```
laurgarn, line 86: Seems big to me. Is there no way to wrap the text in the tooltip in stead of trying to have it on one line ?

anas-drissi: The goal is to have it on one line.

bendaamerahm (approved):

PR #728, opened 2022-07-25
Change message and switch severity from error to warning when loading a study with not found case

FranckLecuyer:


Meklo (approved):
Code : OK
Tests : OK

PR #737, opened 2022-07-28
Disable multiple successive click on switch before the svg is updated

FranckLecuyer:


bendaamerahm (approved):
code review: ok
test: ok

PR #738, opened 2022-07-29
Display info snackbar on study rename or move

Meklo:


src/utils/compute-title.js
```diff
@@ -8,6 +8,16 @@
 const MAX_TITLE_LENGTH = 106;
 const SEPARATOR = ' | ';
 
+export const computeFullPath = (parents) => {
+    let testedPath = '';
+    let path = '';
```
FranckLecuyer, line 13: 'path' variable seems not necessary

FranckLecuyer (changes requested):
Tests:OK

src/utils/compute-title.js
Meklo: One of the variables is indeed useless, I removed testedPath and kept path for semantics reasons

FranckLecuyer (approved):

src/components/study-container.js
```diff
@@ -356,22 +358,45 @@ export function StudyContainer({ view, onChangeTab }) {
...
                 document.title = computePageTitle(
                     initialTitle,
-                    study?.elementName,
+                    studyName,
                     parents
                 );
+
+                studyMetadata.path = path;
+                studyMetadata.name = studyName;
+                setStudyMetadata(studyMetadata);
```
archendel, line 393: is it safe to call setStudyMetadata here whereas the useCallBack has studyMetadata in dependency array ?

archendel (approved):
One small remark

Meklo:
> One test which does not work, but maybe it's out of this PR : If you rename one of the study parents, it does not update anything on gridstudy
> 
> Also, one remark below

It will indeed be the subject of a future PR

src/components/study-container.js
Meklo: Upon research it wasn't best practice compliant so I've overhauled the process through a custom hook

PR #739, opened 2022-08-01
Read only mode for search dialog (for tree node not built)

FranckLecuyer:
Wait for next commons-ui release


SlimaneAmar (approved):

laurgarn (approved):
Works OK

PR #746, opened 2022-08-08
Force edited row on top of table

EstherDarkish:
The edited row in the spreadsheet is now always on top of the table.
The table can be scrolled again, but the edited row will still be displayed on top of other rows.

### DIFF
--- a/src/components/network/network-table.js
+++ b/src/components/network/network-table.js
@@ -447,16 +447,7 @@ const NetworkTable = (props) => {
                 // When modifying a row, a shallow copy of the edited row is created and put on top of the table.
                 // The edition is then done on the copy, on top of the table, and not below, to prevent
                 // bugs of rows changing position in the middle of modifications.
-                let editRowPosition;
-                for (
-                    editRowPosition = 0;
-                    editRowPosition < result.length;
-                    editRowPosition++
-                ) {
-                    if (isLineOnEditMode(result[editRowPosition])) {
-                        return [result[editRowPosition], ...result];
-                    }
-                }
+                return [result[parseInt(lineEdit.rowIndex) - 1], ...result];
             }
             return result;
         },
@@ -467,8 +458,8 @@ const NetworkTable = (props) => {
             filter,
             formatCell,
             props.disabled,
-            isLineOnEditMode,
             isModifyingRow,
+            lineEdit,
         ]
     );
 
@@ -813,6 +804,7 @@ const NetworkTable = (props) => {
                                         oldValues: {},
                                         newValues: {},
                                         id: rowData.id,
+                                        rowIndex: rowIndex,
                                         errors: new Map(),
                                         equipmentType:
                                             TABLES_DEFINITION_INDEXES.get(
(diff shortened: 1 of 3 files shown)


src/components/network/network-table.js
```diff
@@ -421,6 +442,22 @@ const NetworkTable = (props) => {
                     );
                 });
             }
+
+            if (isModifyingRow()) {
+                // When modifying a row, a shallow copy of the edited row is created and put on top of the table.
+                // The edition is then done on the copy, on top of the table, and not below, to prevent
+                // bugs of rows changing position in the middle of modifications.
+                let editRowPosition;
```
dbraquart, line 450: I wonder if we could have memorized the rowPosition when we click on edition button (cf setLineEdit), rather than compute it.

dbraquart (commented):
Tests: OK
Review: OK, just 2 minor comments

src/components/network/network-table.js
```diff
@@ -380,6 +397,10 @@ const NetworkTable = (props) => {
         [rowFilter, selectedDataKey, formatCell]
     );
 
+    const isModifyingRow = useCallback(() => {
```
SlimaneAmar, line 400: Callback really necessary ?

src/components/network/network-table.js
```diff
@@ -430,6 +467,8 @@ const NetworkTable = (props) => {
             filter,
             formatCell,
             props.disabled,
+            isLineOnEditMode,
```
SlimaneAmar, line 470: Slight confusion between these 2 terms : `isLineOnEditMode` and `isModifyingRow`

src/components/network/network-table.js
```diff
@@ -920,43 +1012,57 @@ const NetworkTable = (props) => {
...
-                            defaultValue={currentValue}
-                            setColumnError={(k) => setColumnInError(k)}
-                            resetColumnError={(k) => resetColumnInError(k)}
-                            forceLineUpdate={forceLineUpdate}
-                            columnDefinition={columnDefinition}
-                            setter={(val) => changeRequest(val)}
-                            style={style}
-                        />
-                    );
+                if (rowIndex === 1) {
```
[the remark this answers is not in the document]
EstherDarkish: Yes

src/components/network/network-table.js
EstherDarkish: Yes it is. Without it, I have multiple warnings and lint do not pass.

src/components/network/network-table.js
EstherDarkish: Fixed.

src/components/network/network-table.js
```diff
@@ -135,6 +136,22 @@ const useStyles = makeStyles((theme) => ({
             borderBottom: '1px solid ' + theme.palette.divider,
         },
     },
+    leftFade: {
+        background:
+            'linear-gradient(to right, ' +
+            theme.palette.primary.main +
+            ', rgba(0,0,0,0) 10%)',
```
Meklo, line 143: It's up to personal preferences but with a slight border radius on the left side I reckon it looks a bit better. I tried with the following : `border-radius: 4px 0 0 4px;`

src/components/network/network-table.js
```diff
@@ -713,22 +741,49 @@ const NetworkTable = (props) => {
...
+                        style={style}
+                        className={clsx(
+                            classes.referenceEditRow,
+                            classes.leftFade
+                        )}
+                    >
+                        <div className={classes.editCell}>
+                            <IconButton size={'small'}>
+                                <MoreHorizIcon />
                             </IconButton>
```
Meklo, line 787: Since it's a button but there's no event tied to it, it causes a bit of friction when hovering then clicking and nothing happens
Maybe it could be fixed by centering MoreHorizIcon without using IconButton or by disabling the change of background color on hover ?

Meklo (approved):
Code : OK
Tests : OK

src/components/network/network-table.js
EstherDarkish: Yes, that looks better.

src/components/network/network-table.js
EstherDarkish: Ok, fixed it.

PR #761, opened 2022-08-29
Fix map loading in Tree display mode

sBouzols:


src/components/horizontal-toolbar.js
```diff
@@ -179,9 +182,6 @@ export function HorizontalToolbar({ setStudyDisplayMode, studyDisplayMode }) {
     );
 }
 
-HorizontalToolbar.propTypes = {
-    setStudyDisplayMode: PropTypes.func,
-    studyDisplayMode: PropTypes.string,
-};
+HorizontalToolbar.propTypes = {};
```
etiennehomer, line 185: We keep the empty props ?

etiennehomer (approved):

etiennehomer (changes requested):
We get Lines and Substations when we change the node in Tree display mode. Maybe we could avoid those requests ?

src/components/horizontal-toolbar.js
sBouzols: removed

etiennehomer (approved):

src/components/app-top-bar.js
```diff
@@ -294,14 +311,11 @@ const AppTopBar = ({
...
-                searchTermDisabled={!isNodeBuilt(currentNode)}
-                initialSearchTerm={
-                    !isNodeBuilt(currentNode)
-                        ? intl.formatMessage({
-                              id: 'InvalidNode',
-                          })
-                        : ''
+                searchTermDisabled={
+                    !isNodeBuilt(currentNode) ||
+                    studyDisplayMode === STUDY_DISPLAY_MODE.TREE
```
FranckLecuyer, line 316: The test could be : getDisableReason() !== ''   ?

FranckLecuyer (changes requested):

src/components/app-top-bar.js
sBouzols: :+1: done

FranckLecuyer (approved):

PR #769, opened 2022-09-07
Improve multi sld display

etiennehomer:


EstherDarkish:
There are bugs with the "fullscreen" button : 

Depending on the zoom level of the web browser, the fullscreen button can be out of range of the shown area.
This occures each time the displayed circuit is bigger than the surounding display box.

Moreover, the button lost its fullscreen effect. When we click on it, instead of showing the SLD in full screen mode, it only sligthly change its size, and it is not consistent.

EstherDarkish:
There is a bug with the minimize function : when a SLD is minimized, we should click on its up arrow to show it again. Sometimes, clicking on the arrow makes the element disapear and is not shown anymore. When clicking on another element's up arrow, the first one comes back.

EstherDarkish (changes requested):
Bugs with the fullscreen function and the minimize function.

etiennehomer:
> There are bugs with the "fullscreen" button :
> 
> Depending on the zoom level of the web browser, the fullscreen button can be out of range of the shown area. This occures each time the displayed circuit is bigger than the surounding display box.
> 
> Moreover, the button lost its fullscreen effect. When we click on it, instead of showing the SLD in full screen mode, it only sligthly change its size, and it is not consistent.

Those bugs could be fixed

etiennehomer:
> There is a bug with the minimize function : when a SLD is minimized, we should click on its up arrow to show it again. Sometimes, clicking on the arrow makes the element disapear and is not shown anymore. When clicking on another element's up arrow, the first one sometimes comes back, but not always.

I couldn't reproduce this one. And maybe it's more related to the other US on the SLD

jonenst (approved):

PR #770, opened 2022-09-07
Move anjd split input hooks in inputs sub dir

laurgarn:


src/components/dialogs/inputs/use-expandable-values.js
```diff
@@ -0,0 +1,151 @@
+import { useStyles } from '../dialogUtils';
```
sBouzols, line 1: license header

sBouzols (changes requested):
license header missing
will test a bit

sBouzols (approved):

PR #771, opened 2022-09-08
fixing tooltip for see always part of the path name

anas-drissi:


src/components/study-container.js
```diff
@@ -389,7 +389,8 @@ export function StudyContainer({ view, onChangeTab }) {
                     .slice(1)
                     .map((parent) => parent.elementName);
 
-                const studyName = response[0]?.elementName;
+                const studyName =
```
laurgarn, line 392: Why adding the ellipsis even when the name is short enough ?

src/components/study-container.js
```diff
@@ -351,6 +351,10 @@ export function StudyContainer({ view, onChangeTab }) {
         // Note: studyUuid and dispatch don't change
     }, [studyUuid, dispatch, snackError, snackWarning]);
 
+    function limitChar(str) {
```
laurgarn, line 354: Would be better to define it out of the component function, as it avoids a new parsing of it at each call of the component function.

anas-drissi: it's fixed

laurgarn (approved):

src/components/study-container.js
```diff
@@ -389,7 +393,7 @@ export function StudyContainer({ view, onChangeTab }) {
                     .slice(1)
                     .map((parent) => parent.elementName);
 
-                const studyName = response[0]?.elementName;
+                const studyName = limitChar(response[0]?.elementName, 30);
```
bendaamerahm, line 396: i think we don't do that here , there is a file that contain already a method computePageTitle in compute-title.js, just put your modification there

bendaamerahm (changes requested):

bendaamerahm (approved):
code review: ok
test: ok

PR #773, opened 2022-09-08
refresh result tab

anas-drissi:
Refresh result tab for LoadFlow and Security Analysis.


bendaamerahm (approved):
test: ok
code review: ok

src/components/result-view-tab.js
```diff
@@ -52,10 +53,16 @@ export const ResultViewTab = ({
...
 
+    const loadflowNotif = useSelector((state) => state.loadflowNotif);
+
     function renderLoadFlowResult() {
         return (
             <Paper className={classes.table}>
-                <LoadFlowResult result={loadFlowInfos?.loadFlowResult} />
+                <LoadFlowResult
+                    result={
+                        loadflowNotif ? loadFlowInfos?.loadFlowResult : null
```
laurgarn, line 63: You don't need to use the ternary when you use the optional chaining.
-> loadFlowInfos?.loadFlowResult ?? null
if you just want to avoid undefined.

anas-drissi: This is to avoid having results when there is no notification and it is also to empty the table of results when the calculation is invalidated when you log or trigger a line

anas-drissi: it's fixed

laurgarn (approved):

PR #776, opened 2022-09-15
fix:wraps SLD title more freely and keep icons stuck to top

laurgarn:


src/components/diagrams/singleLineDiagram/single-line-diagram.js
```diff
@@ -1011,26 +1012,36 @@ const SizedSingleLineDiagram = forwardRef((props, ref) => {
...
-                    >
-                        {pinned ? <PushPinIcon /> : <PushPinOutlinedIcon />}
-                    </IconButton>
-                    <IconButton
-                        className={classes.close}
-                        onClick={onCloseHandler}
-                    >
-                        <CloseIcon />
-                    </IconButton>
+                    <Box>
```
etiennehomer, line 1015: We can remove this box nope ?

laurgarn: Nope, if we want to keep the icons stuck to the top and have the label to uses more than one line of text, we need the two levels, outer for vertical concern and inner for horizontal alignment.

etiennehomer: Ah yes, you're right

niconoir (approved):

etiennehomer (approved):

PR #796, opened 2022-10-07
extracts SLD viewer to a typescript module

CBiasuzzi:
The target is to extract the SLD viewer from gridstudy-app to the typescript repository powsybl-diagram-viewer, alongside the network-area-diagram-viewer. 
Besides the SVG to be rendered, the typescript module takes as input the SVG metadata, the callback for the navigation between VLs, the callback around the user click on a switch and a callback related to the feeder's context menu.

The extracted viewer component is available in the powsybl-diagram-viewer's repo  (version 0.1.5).

Outdated note:  Currently, the extracted and typescript-ed code lives in a branch (sld) of repo powsybl-diagram-viewer, therefore it is assumed to have a local copy of that project's branch (in package.json, the dependency is declared with a local, relative path
   "@powsybl/diagram-viewer": "file:../powsybl-diagram-viewer", to be changed to a normal version number, as soon as the branch is merged and published).


jonenst:
good. Minor point, this changes the current behavior that uses the browser to compute the size and relies on the size computed on the server (which is "wrong" when there are long names or ids because the server doesn't take text into account), but I think this is what we want to do.

jonenst (approved):
as you mentioned, don't forget to use the new version in package.json

etiennehomer:
@CBiasuzzi The powsybl-diagram-viewer v0.1.4 has been published. You can use it in this PR

etiennehomer (approved):

etiennehomer (changes requested):
There are problems with the breakers and disconnectors interactions. 
There's a delay between the click and the action.
And the breaker position change, get back to initial position and then move to the final position after the server response.

etiennehomer:
> There are problems with the breakers and disconnectors interactions. There's a delay between the click and the moment the loader appears. And the breaker position change, get back to initial position and then move to the final position after the server response.

CBiasuzzi:
thx, you're right. This locallySwitchedBreaker behavior was added recently, I can see. ...
I'll try to fix the conflicts ; the PR is still in WIP

etiennehomer:
> thx, you're right. This locallySwitchedBreaker behavior was added recently, I can see. ... I'll try to fix the conflicts ; the PR is still in WIP

Yes the merge is not obvious...

CBiasuzzi:
> > thx, you're right. This locallySwitchedBreaker behavior was added recently, I can see. ... I'll try to fix the conflicts ; the PR is still in WIP
> 
> Yes the merge is not obvious...

I believe I've fixed the merge issue: now when a switch is clicked it should mimic the original behavior.
In the meantime, though, I found a problem with larger svg (for example, displaying the BE_microgrid's PP_ substation);
I patched it, but unfortunately I had to touch the viewer project: I opened a pull request there ,too ( I updated the viewr's version in the package* files)

etiennehomer:
> > > thx, you're right. This locallySwitchedBreaker behavior was added recently, I can see. ... I'll try to fix the conflicts ; the PR is still in WIP
> > 
> > 
> > Yes the merge is not obvious...
> 
> I believe I've fixed the merge issue: now when a switch is clicked it should mimic the original behavior. In the meantime, though, I found a problem with larger svg (for example, displaying the BE_microgrid's PP_ substation); I patched it, but unfortunately I had to touch the viewer project: I opened a pull request there ,too ( I updated the viewr's version in the package* files)

Interactions work fine ! 
Yes the code for large svg is quite sensitive to changes. Ok i'll check the other PR. There's no problem with releasing new versions of powsybl-diagram-viewer, it's rather quick

etiennehomer:
@CBiasuzzi  The powsybl-diagram-viewer v0.1.5 has been published. You can use it in this PR

CBiasuzzi:
> @CBiasuzzi The powsybl-diagram-viewer v0.1.5 has been published. You can use it in this PR

thank you. I updated the package* files with the new version. 
I'm removing the WIP tag in this PR, at this point.

etiennehomer (approved):

PR #804, opened 2022-10-13
Fix substation creation when name is empty or contain only spaces

Walid-Sahnoun:


FranckLecuyer (approved):

src/utils/rest-api.js
```diff
@@ -1658,6 +1658,13 @@ export function createTwoWindingsTransformer(
     });
 }
 
+function checkEquipementNameEmptyOrContainsOnlySpaces(name) {
```
SlimaneAmar, line 1661: Remove it

src/utils/rest-api.js
```diff
@@ -1690,7 +1697,8 @@ export function createSubstation(
         },
         body: JSON.stringify({
             equipmentId: substationId,
-            equipmentName: substationName,
+            equipmentName:
+                checkEquipementNameEmptyOrContainsOnlySpaces(substationName),
```
SlimaneAmar, line 1701: substationName.trim() === '' ? null : substationName.trim()

SlimaneAmar (changes requested):

src/utils/rest-api.js
Walid-Sahnoun: done

src/utils/rest-api.js
Walid-Sahnoun: done

PR #805, opened 2022-10-13
Add map manual refresh parameter

Meklo:


src/translations/en.json
```diff
@@ -83,6 +83,8 @@
     "Limit": "Limit",
     "load": "load",
     "Load": "Load",
+    "MapManualRefresh": "Manual update of geographical view\u00a0:",
```
EstherDarkish, line 86: There is no space before colons in english.

EstherDarkish (approved):

src/translations/en.json
Meklo: ✓

src/components/study-container.js
```diff
@@ -542,44 +556,57 @@ export function StudyContainer({ view, onChangeTab }) {
...
-                const substationsIds =
-                    studyUpdatedForce.eventData.headers['substationsIds'];
-                const tmp = substationsIds.substring(
-                    1,
-                    substationsIds.length - 1
-                ); // removing square brackets
-                if (tmp && tmp.length > 0) {
-                    updateNetwork(tmp.split(', '));
-                }
+        if (!mapManualRefresh || !reloadMapNeeded) {
```
niconoir, line 559: I am not sure if the modifications made in this file are relevant. If the network is not reloaded when currentNode changes or when network has been updated on the back side, the data displayed in the equipment table might be wrong.

niconoir (changes requested):

src/components/study-container.js
Meklo: I implemented a fix which reload network data when the user switch to the spreadsheet
If the user is in manual mode, the spreadsheet is already open, and new data comes in therefore the data is invalidated

src/components/study-container.js
```diff
@@ -352,6 +364,18 @@ export function StudyContainer({ view, onChangeTab }) {
         // Note: studyUuid and dispatch don't change
     }, [studyUuid, dispatch, snackError, snackWarning]);
 
+    useEffect(() => {
```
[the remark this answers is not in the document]
Meklo: One useEffect is there to handle manual mode triggers and the other is there to handle automatic mode triggers
In practice I have verified that during network and equipment manipulations there are no overlap between the two, the network reload only once when needed.
It's on the app startup that there might be multiple call to network loading.
I'm currently looking for a fix.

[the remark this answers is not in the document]
Meklo: It's fixed

niconoir (approved):

PR #811, opened 2022-10-17
Fix user disconnect

flomillot:
Unselect the current node when the user is disconnected
Fix npm warnings

### DIFF
--- a/src/components/app.js
+++ b/src/components/app.js
@@ -42,6 +42,7 @@ import {
     selectFluxConvention,
     selectFavoriteSensiVariablesFiltersLists,
     selectFavoriteSensiBranchFiltersLists,
+    selectMapManualRefresh,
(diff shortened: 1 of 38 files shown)


flomillot:
Je viens de tomber sur #766 du coup j'abandonne la PR

src/components/app.js
```diff
@@ -338,6 +339,11 @@ const App = () => {
         });
     });
 
+    // Manage user (both manual and auto) disconnection
+    useEffect(() => {
```
niconoir, line 343: I would have put this code in app-top-bar.js since it seems strange to me to handle the currentTreeNode redux storage so high in the hierarchy.

niconoir (changes requested):

niconoir (approved):

PR #812, opened 2022-10-17
Delete fields in generator modification

YenguiSeddik:
Delete the new fields added to generator creation dialogue from generator modification dialogue


anas-drissi:
The creation is done even without having entered anything in the form, is this normal ? 
Because in case of edition the form is not editable, we have an error 500.

SlimaneAmar (approved):

anas-drissi (approved):

PR #813, opened 2022-10-18
Update load connectivity

bendaamerahm:


src/components/dialogs/connectivity-edition.js
```diff
@@ -138,7 +138,7 @@ export const useConnectivityValue = ({
         label: 'ConnectionDirection',
         validation: { isFieldRequired: false },
         inputForm: inputForm,
-        formProps: filledTextField,
+        formProps: outlinedTextField,
```
FranckLecuyer, line 141: Are you sure you have to give explicitly this property ?
I see in the documentation that 'outlined' is the default variant for a TextField

FranckLecuyer (changes requested):

src/components/dialogs/connectivity-edition.js
bendaamerahm: yes you are right, ill delete this prop in this case

FranckLecuyer (approved):

PR #855, opened 2022-11-08
fix targetV or targetQ non editable fields

FranckLecuyer:


src/components/network/network-table.js
```diff
@@ -1004,7 +1004,12 @@ const NetworkTable = (props) => {
         (rowData, columnDefinition, key, style, rowIndex) => {
             if (
                 isLineOnEditMode(rowData) &&
-                rowData[columnDefinition.dataKey] !== undefined &&
+                (rowData[columnDefinition.dataKey] !== undefined ||
```
bendaamerahm, line 1007: it will be more readable if we externalize those conditions into a function

bendaamerahm (changes requested):

src/components/network/network-table.js
FranckLecuyer: Done

PR #857, opened 2022-11-09
Generator form update : Reactive Capability Curve

EstherDarkish:
Updates the generator form's Reactive Capability Curve segment.


src/components/dialogs/inputs/input-hooks.js
```diff
@@ -864,3 +736,81 @@ export const useCSVReader = ({ label, header }) => {
...
+    validation = {},
+    inputForm,
+    doTranslation = true,
+}) => {
+    const [value, setValue] = useState(defaultValue);
+    const intl = useIntl();
+
+    useEffect(() => {
+        // Updates the component when the correct default value is given by the parent component.
+        if (defaultValue !== undefined && defaultValue.length > 0) {
```
laurgarn, line 754: Can be simplified to : `defaultValue?.length > 0`

src/components/dialogs/inputs/reactive-capability-curve-table.js
```diff
@@ -0,0 +1,212 @@
...
+        const newValues = [
+            ...values.slice(0, newPosition),
+            minimalValue,
+            ...values.slice(newPosition),
+        ];
+
+        setValues(newValues);
+
+        // Adds a unique ID on each displayed line, to prevent a bad rendering effect
+        let newValuesForDisplay = [];
```
laurgarn, line 108: Could be written with Array.prototype.map
```
const newValuesForDisplay = newValues.map((v) => {
  v.id = getNewId();
  return v;
});
```

src/components/dialogs/inputs/reactive-capability-curve-table.js
```diff
@@ -0,0 +1,212 @@
...
+            const enforcedMinimumDefaultValues =
+                enforceMinimumValues(defaultValues);
+            setValues(enforcedMinimumDefaultValues);
+            setDisplayedValues(enforcedMinimumDefaultValues);
+        }
+    }, [defaultValues]);
+
+    const handleDeleteItem = useCallback(
+        (index) => {
+            const newValues = [
```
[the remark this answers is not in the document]
sBouzols: even shorter
`const newValues = values.slice(index, 1);`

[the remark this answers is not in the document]
laurgarn: Not for this one, I guess.
`[1, 2, 3, 4, 5].slice(2, 3)`
gives
`[ 3 ]`
splice is mainly used for its side effect, that is removing entries, more than for returning the slice removed.

src/components/dialogs/generator-creation-dialog.js
```diff
@@ -460,86 +489,76 @@ const GeneratorCreationDialog = ({
...
+
                         {/* Connectivity part */}
-                        <Grid container spacing={2}>
-                            <Grid item xs={12}>
-                                <h3 className={classes.h3}>
-                                    <FormattedMessage id="Connectivity" />
-                                </h3>
-                            </Grid>
-                        </Grid>
+                        <GridSection title="Connectivity" />
```
sBouzols, line 494: it's a translation IDthere,  then it should be clearer for the propName
maybe something like `titleMessageDescriptor` with this format :
https://formatjs.io/docs/react-intl/components/#formattedmessage
then you pass it throught FormattedMessage component

src/components/dialogs/generator-creation-dialog.js
```diff
@@ -460,86 +489,76 @@ const GeneratorCreationDialog = ({
...
+                            {!!isReactiveCapabilityCurveOn &&
+                                rCCurveError.length > 0 && (
+                                    <Grid container spacing={2}>
+                                        <Grid item xs={12}>
+                                            {rCCurveError.map((value) => (
+                                                <div
+                                                    className={classes.rccError}
+                                                >
+                                                    <FormattedMessage
+                                                        id={value}
```
sBouzols, line 524: it's not a value, maybe rename this as `messageDescriptorId` then `rCCurveError` is an array of `messageDescriptorId` as well

src/components/dialogs/inputs/input-hooks.js
```diff
@@ -18,8 +18,11 @@ import { validateField } from '../../util/validation-functions';
 import {
     CircularProgress,
     FormHelperText,
+    FormLabel,
     InputLabel,
     MenuItem,
+    Radio,
+    RadioGroup,
```
sBouzols, line 25: maybe move the import of `FormControlLabel` here as well ?

src/components/dialogs/inputs/input-hooks.js
```diff
@@ -864,3 +736,81 @@ export const useCSVReader = ({ label, header }) => {
...
+
+    useEffect(() => {
+        function validate() {
+            return true;
+        }
+
+        inputForm.addValidation(id ? id : label, validate);
+    }, [label, validation, inputForm, value, id]);
+
+    const handleChangeValue = useCallback((event) => {
```
sBouzols, line 767: Why to use a useCallback here ?

src/components/dialogs/inputs/input-hooks.js
```diff
@@ -864,3 +736,81 @@ export const useCSVReader = ({ label, header }) => {
...
+        label,
+        handleChangeValue,
+        id,
+        defaultValue,
+        doTranslation,
+        possibleValues,
+        value,
+    ]);
+
+    useEffect(
```
sBouzols, line 810: You already have a useEffect line 752 which is proc by `defaultValue` which do the exact same action. Why not using only one UseEffect ?

src/components/dialogs/inputs/input-hooks.js
```diff
@@ -864,3 +736,81 @@ export const useCSVReader = ({ label, header }) => {
...
 };
+
+export const useRadioValue = ({
+    label,
+    possibleValues = [],
+    defaultValue,
+    id,
+    validation = {},
+    inputForm,
+    doTranslation = true,
```
sBouzols, line 747: Is there a use case when we want to avoid translation ?

src/components/dialogs/inputs/input-hooks.js
```diff
@@ -864,3 +736,81 @@ export const useCSVReader = ({ label, header }) => {
...
+                    value={value ?? defaultValue}
+                    onChange={handleChangeValue}
+                >
+                    {possibleValues.map((value) => (
+                        <FormControlLabel
+                            value={value}
+                            control={<Radio />}
+                            label={
+                                doTranslation
+                                    ? intl.formatMessage({ id: value })
```
sBouzols, line 791: I think you should manage separately the value of the Radio and its label.

src/components/dialogs/reactive-capability-curve-table.js
```diff
@@ -21,9 +21,10 @@ export const ReactiveCapabilityCurveTable = ({
     inputForm,
     isFieldRequired,
     disabled = false,
+    customPLabel = undefined,
```
sBouzols, line 24: Why do you need a customPLabel ?

src/components/util/validation-functions.js
```diff
@@ -61,6 +88,61 @@ export function validateField(value, toValidate) {
...
+        errorMessages.push('ReactiveCapabilityCurveCreationErrorMissingPoints');
+    }
+
+    // Each P must be a unique valid number
+    const everyValidP = reactiveCapabilityCurve
+        .map((element) =>
+            validateValueIsANumber(element.p) ? element.p : null
+        )
+        .filter((p) => p !== null);
+    const setOfPs = [...new Set(everyValidP)];
```
sBouzols, line 112: why not `const setOfPs = new Set(everyValidP);` ?

src/components/util/validation-functions.js
```diff
@@ -61,6 +88,61 @@ export function validateField(value, toValidate) {
...
+
+    if (setOfPs.length !== everyValidP.length) {
```
sBouzols, line 114: then setOfPs.size

sBouzols (changes requested):

src/components/dialogs/generator-creation-dialog.js
```diff
@@ -54,11 +57,14 @@ const useStyles = makeStyles((theme) => ({
...
         color: theme.palette.error.main,
         fontSize: 'small',
-        textAlign: 'center',
-        margin: theme.spacing(2),
+        marginLeft: theme.spacing(2),
+        marginRight: theme.spacing(2),
     },
 }));
 
+export const REACTIVE_LIMIT_TYPE_MIN_MAX = 'ReactiveLimitsKindMinMax';
```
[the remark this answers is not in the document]
EstherDarkish: Ok, removed the export clause.

src/components/dialogs/generator-creation-dialog.js
```diff
@@ -87,7 +93,10 @@ const GeneratorCreationDialog = ({
 
     const [reactivePowerRequired, setReactivePowerRequired] = useState(false);
 
-    const [rCCurveError, setRCCurveError] = useState();
+    const [isReactiveCapabilityCurveOn, setReactiveCapabilityCurveOn] =
+        useState(true);
+
+    const [rCCurveError, setRCCurveError] = useState([]);
```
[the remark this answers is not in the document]
EstherDarkish: Ok, set to reactiveCapabilityCurveErrors.

src/components/dialogs/generator-creation-dialog.js
```diff
@@ -213,14 +222,26 @@ const GeneratorCreationDialog = ({
...
+                    : REACTIVE_LIMIT_TYPE_CURVE,
+            possibleValues: [
+                REACTIVE_LIMIT_TYPE_MIN_MAX,
+                REACTIVE_LIMIT_TYPE_CURVE,
+            ],
         });
 
+    useEffect(() => {
+        setReactiveCapabilityCurveOn(
+            reactiveCapabilityCurveChoice === REACTIVE_LIMIT_TYPE_CURVE
```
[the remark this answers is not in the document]
EstherDarkish: Yes, replaced with your function.

src/components/dialogs/generator-creation-dialog.js
EstherDarkish: `GridSubSection` is based on the old `GridSection` that was already there.
I could either change the old code and refactor the eight files that use GridSection, or I could use the same templating as GridSection.
What do you prefer ?

src/components/dialogs/generator-creation-dialog.js
EstherDarkish: Ok, changed "value" to "messageDescriptorId".
rCCurveError is now reactiveCapabilityCurveErrors.
Also changed the CSS class name to be more generic.

src/components/dialogs/inputs/input-hooks.js
EstherDarkish: Yes, thank you, now I'll check on this kind of stuff after using the auto include function of my IDE.

src/components/dialogs/inputs/input-hooks.js
EstherDarkish: Not yet, but in the future, if we re-use this component for something like "Select your age", we will not have to include every numbers in the en.json and fr.json files.

src/components/dialogs/inputs/input-hooks.js
EstherDarkish: Yes, thank you.

src/components/dialogs/inputs/input-hooks.js
EstherDarkish: Good point, it is unecessary. Removed it for a lambda inline.

src/components/dialogs/inputs/input-hooks.js
```diff
@@ -864,3 +736,81 @@ export const useCSVReader = ({ label, header }) => {
...
+    const handleChangeValue = useCallback((event) => {
+        setValue(event.target.value);
+    }, []);
+
+    const field = useMemo(() => {
+        return (
+            <FormControl>
+                {label && (
+                    <FormLabel id={id ? id : label}>
+                        {intl.formatMessage({ id: label })}
```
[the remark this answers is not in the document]
EstherDarkish: Ok, I changed it.

src/components/dialogs/inputs/input-hooks.js
EstherDarkish: Nice catch, the second one is an error. I removed it.

src/components/dialogs/inputs/reactive-capability-curve-table.js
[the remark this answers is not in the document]
EstherDarkish: Ok, I used @laurgarn 's solution

src/components/dialogs/inputs/reactive-capability-curve-table.js
```diff
@@ -0,0 +1,212 @@
...
+            setDisplayedValues(newValues);
+
+            inputForm.reset();
+        },
+        [values, inputForm]
+    );
+
+    const handleAddValue = useCallback(() => {
+        const newPosition = values.length - 1;
+        const newValues = [
```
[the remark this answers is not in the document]
EstherDarkish: Done.

src/components/dialogs/inputs/reactive-capability-curve-table.js
EstherDarkish: Yes, thank you.

src/components/util/validation-functions.js
EstherDarkish: Because I'd lose the ".length" function, and I want it so it is clearer in the code.

src/components/util/validation-functions.js
EstherDarkish: I'd rather have length on both side, it is clearer.

src/components/dialogs/inputs/input-hooks.js
EstherDarkish: Ok done.

src/components/dialogs/reactive-capability-curve-table.js
EstherDarkish: This is to override the default "P" label in the field. It is needed for the first and last lines of the reactive capability limit curves.

src/components/dialogs/generator-creation-dialog.js
```diff
@@ -238,14 +255,13 @@ const GeneratorCreationDialog = ({
...
 
     const [reactiveCapabilityCurve, reactiveCapabilityCurveField] =
-        useTableValues({
-            id: 'ReactiveCapabilityCurveOn',
+        useReactiveCapabilityCurveTableValues({
             tableHeadersIds: headerIds,
             inputForm: inputForm,
             Field: ReactiveCapabilityCurveTable,
             defaultValues: formValues?.reactiveCapabilityCurvePoints,
             isRequired: false,
```
sBouzols, line 263: `isRequired` seems not used anymore in this custom hook ?

src/components/dialogs/inputs/reactive-capability-curve-table.js
```diff
@@ -0,0 +1,205 @@
...
+    isReactiveCapabilityCurveOn,
+    disabled = false,
+}) => {
+    const classes = useStyles();
+
+    // Values sent back to the parent component.
+    const [values, setValues] = useState(enforceMinimumValues(defaultValues));
+
+    // Values displayed in the form. When updated, they also updates "values".
+    const [displayedValues, setDisplayedValues] = useState(
```
sBouzols, line 68: I don't know, we should avoid those two states which save the exact same thing...
I know it's a kind of a fix for the flickering problem, but seems a bad fix... I will think to another solution

src/components/dialogs/inputs/reactive-capability-curve-table.js
```diff
@@ -0,0 +1,205 @@
...
+
+    const handleAddValue = useCallback(() => {
+        const newValues = [...values];
+        newValues.splice(values.length - 1, 0, minimalValue);
+
+        setValues(newValues);
+
+        // Adds a unique ID on each displayed line, to prevent a bad rendering effect
+        const newValuesForDisplay = newValues.map((v) => {
+            v.id = getNewId();
```
sBouzols, line 103: then there it seems that `displayedValues` has another format than `values`

src/components/dialogs/inputs/reactive-capability-curve-table.js
```diff
@@ -0,0 +1,205 @@
...
+        }
+    }, [defaultValues]);
+
+    const handleDeleteItem = useCallback(
+        (index) => {
+            const newValues = [...values];
+            newValues.splice(index, 1);
+
+            setValues(newValues);
+            setDisplayedValues(newValues);
```
sBouzols, line 88: But when deleting a line we set DisplayedValues as the same as Values...

src/components/dialogs/inputs/reactive-capability-curve-table.js
```diff
@@ -0,0 +1,205 @@
...
+        // Adds a unique ID on each displayed line, to prevent a bad rendering effect
+        const newValuesForDisplay = newValues.map((v) => {
+            v.id = getNewId();
+            return v;
+        });
+        setDisplayedValues(newValuesForDisplay);
+    }, [values]);
+
+    const handleSetValue = useCallback((index, newValue) => {
+        setValues((oldValues) => [
```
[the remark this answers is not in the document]
sBouzols: But it's not, you don't only want to insert a newValue in the array, you want to overwrite the existing one with the same index as well

etiennehomer:
In input-hooks.js, function useRadioValue, you can put : 
const handleChangeValue = useCallback(
        (event) => {
            setValue(event.target.value)
            inputForm.setHasChanged(true);
        },
        [inputForm]
    );
and use it in onChange prop. 

So the validate button is enable when the user change the radio button state

src/components/dialogs/inputs/input-hooks.js
```diff
@@ -864,3 +736,63 @@ export const useCSVReader = ({ label, header }) => {
...
+                {label && (
+                    <FormLabel id={id ? id : label}>
+                        <FormattedMessage id={label} />
+                    </FormLabel>
+                )}
+                <RadioGroup
+                    row
+                    aria-labelledby={id ? id : label}
+                    value={value ?? defaultValue}
+                    onChange={(e) => setValue(e.target.value)}
```
etiennehomer, line 779: const handleChangeValue = useCallback(
(event) => {
setValue(event.target.value)
inputForm.setHasChanged(true);
},
[inputForm]
);
and use it in onChange prop.

So the validate button is enable when the user change the radio button state

src/components/dialogs/generator-creation-dialog.js
```diff
@@ -117,7 +121,13 @@ const GeneratorCreationDialog = ({
             stepUpTransformerReactance: generator.stepUpTransformerReactance,
             reactiveCapabilityCurvePoints:
                 generator.reactiveCapabilityCurvePoints,
-            minMaxReactiveLimits: generator?.minMaxReactiveLimits,
+            minMaxReactiveLimits: generator.minMaxReactiveLimits,
+            reactiveCapabilityCurve:
```
etiennehomer, line 125: reactiveCapabilityCurve: reactiveCapabilityCurve, 

Nope ?

etiennehomer (changes requested):

src/components/dialogs/generator-creation-dialog.js
```diff
@@ -213,13 +223,20 @@ const GeneratorCreationDialog = ({
...
-            defaultValue: formValues?.reactiveCapabilityCurve ?? true,
-        });
+    const [
+        reactiveCapabilityCurveChoice,
+        reactiveCapabilityCurveChoiceRadioButton,
+    ] = useRadioValue({
+        validation: fieldRequired,
+        inputForm: inputForm,
+        defaultValue:
+            formValues?.reactiveCapabilityCurve === false ? 'MINMAX' : 'CURVE',
```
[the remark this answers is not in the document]
etiennehomer: And use :
!formValues?.reactiveCapabilityCurve

src/components/dialogs/inputs/reactive-capability-curve-table.js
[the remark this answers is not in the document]
EstherDarkish: Ok

src/components/dialogs/inputs/reactive-capability-curve-table.js
```diff
@@ -0,0 +1,205 @@
...
+
+    const handleDeleteItem = useCallback(
+        (index) => {
+            const newValues = [...values];
+            newValues.splice(index, 1);
+
+            setValues(newValues);
+            setDisplayedValues(newValues);
+
+            inputForm.reset();
```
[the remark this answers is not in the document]
EstherDarkish: Yes, but each field is re-validated anyway after each render (it was done like this). Maybe should be changed in the future.

src/components/dialogs/generator-creation-dialog.js
EstherDarkish: Correct.

src/components/dialogs/inputs/reactive-capability-curve-table.js
EstherDarkish: Same format with additional value which is only used by React to update its table.

src/components/dialogs/inputs/reactive-capability-curve-table.js
```diff
@@ -0,0 +1,205 @@
+/*
```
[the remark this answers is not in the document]
EstherDarkish: The other one was changed so they do not share the same name anymore.

src/components/dialogs/inputs/input-hooks.js
EstherDarkish: Yes, OK.

src/components/dialogs/generator-creation-dialog.js
etiennehomer: Discussed with Charly and Sylvain: no correction to be done

sBouzols (approved):
Code Review OK
Tests OK

etiennehomer (approved):

PR #860, opened 2022-11-14
Cut and paste node

archendel:


src/components/network-modification-tree-pane.js
```diff
@@ -210,18 +218,51 @@ export const NetworkModificationTreePane = ({
...
+                cutTreeNode(
+                    studyUuid,
+                    selectedNodeIdForCopy,
+                    referenceNodeId,
+                    insertMode
+                )
+                    .then(() => {
+                        //After the first CUT / PASTE operation, it becomes a COPY / PASTE operation
+                        setSelectedNodeIdForCopy(null);
+                        setCopyType(null);
```
AbdelHedhili, line 249: if it's becoming a copy after the cut why does copyType and selectedNodeId are set to null ? Shouldn't this go to CopyType.COPY and the id stays ?

archendel: Forgot to modify comment, after cut/paste, we can't paste anymore

AbdelHedhili (approved):
don't forget to update or remove the comment and it's all ok for me

bendaamerahm (approved):
code review: ok
test: ok

PR #866, opened 2022-11-15
2-windings-transformer creation dialog's tab are now always displayed when scrolling, and turn red when a field is in error

archendel:
Signed-off-by: LE SAULNIER Kevin <kevin.lesaulnier@rte-france.com>


src/components/dialogs/two-windings-transformer/two-windings-transformer-creation-dialog.js
```diff
@@ -725,13 +742,16 @@ const TwoWindingsTransformerCreationDialog = ({
     const handleSave = () => {
         setCreationError();
         let isFormValid = true;
+        let tabWithError = [];
```
dbraquart, line 745: tabWithError is not plural, and has the exact same name than a style.
Can we change it ? tabsWithError, tabWithErrorList, allTabsWithError ...

src/components/dialogs/two-windings-transformer/two-windings-transformer-creation-dialog.js
```diff
@@ -725,13 +742,16 @@ const TwoWindingsTransformerCreationDialog = ({
...
         if (!characteristicsInputForm.validate()) {
             isFormValid = false;
+            tabWithError.push(0);
```
dbraquart, line 748: I would use constants like:
	const CHARACTERISTICS_TAB = 0;
	const RATIO_TAP_TAB = 1;
	const PHASE_TAP_TAB = 2;
and use them everywhere in the code rather than 0/1/2
(will ease if we swap tabs or insert a new one)

dbraquart (changes requested):
tests: OK
code: OK with minor changes requested

Meklo (approved):
I agree with David on plural variable name and maybe implementing constants in order to improve maintenance and readability. 
This said, everything is working neatly

src/components/dialogs/two-windings-transformer/two-windings-transformer-creation-dialog.js
archendel: fixed

dbraquart (approved):

PR #878, opened 2022-11-21
Removes a warning cause by a missing key in a map.

EstherDarkish:

### DIFF
--- a/src/components/dialogs/generator-creation-dialog.js
+++ b/src/components/dialogs/generator-creation-dialog.js
@@ -606,8 +606,12 @@ const GeneratorCreationDialog = ({
                                     <Grid container spacing={2}>
                                         <Grid item xs={12}>
                                             {reactiveCapabilityCurveErrors.map(
-                                                (messageDescriptorId) => (
+                                                (
+                                                    messageDescriptorId,
+                                                    index
+                                                ) => (
                                                     <div
+                                                        key={index}
                                                         className={
                                                             classes.midFormErrorMessage
                                                         }


src/components/dialogs/generator-creation-dialog.js
```diff
@@ -606,8 +606,12 @@ const GeneratorCreationDialog = ({
                                     <Grid container spacing={2}>
                                         <Grid item xs={12}>
                                             {reactiveCapabilityCurveErrors.map(
-                                                (messageDescriptorId) => (
+                                                (
+                                                    messageDescriptorId,
+                                                    index
+                                                ) => (
                                                     <div
+                                                        key={index}
```
jonenst, line 614: would be better to store a logical unique identifier for the error and use it as a key. For example, field X can only ever have one error of type "NonNegativeFoo" so the key may be "X-NonNegativeFoo"

EstherDarkish: Yes, we could do it, but is it necessary ? The keys are never used, they are only there to avoid a warning.
As per the documentation :

> Keys Must Only Be Unique Among Siblings
> 
> Keys used within arrays should be unique among their siblings. However, they don’t need to be globally unique. We can use the same keys when we produce two different arrays

Source : https://reactjs.org/docs/lists-and-keys.html#keys-must-only-be-unique-among-siblings

EstherDarkish: After talking about it, it is better to add the correct key because, among other reasons, its a better practice and do not cost anything.

src/components/dialogs/two-windings-transformer/two-windings-transformer-creation-dialog.js
```diff
@@ -438,7 +438,7 @@ const TwoWindingsTransformerCreationDialog = ({
                 isFieldRequired:
                     regulationMode === REGULATION_MODES.CURRENT_LIMITER.id &&
                     phaseTapRegulating,
-                isValueGreaterThan: '0',
+                valueGreaterThan: '0',
```
laurgarn, line 441: why use string rather than number zero ?

EstherDarkish: String or number, it makes no difference here.
I only changed the name of the param, not its value.

src/translations/fr.json
```diff
@@ -421,9 +421,11 @@
...
-    "ReactiveCapabilityCurveCreationErrorPInvalid": "Chaque valeur de puissance active doit être un nombre unique et valide.",
-    "ReactiveCapabilityCurveCreationErrorPOutOfRange": "Chaque valeur de puissance active doit être comprise entre les valeurs P min et P max.",
-    "ReactiveCapabilityCurveCreationErrorQminPQmaxPIncoherence": "Chaque valeur de puissance réactive minimale doit être inférieur ou égale à la puissance réactive maximale correspondante.",
+    "ReactiveCapabilityCurveCreationErrorPInvalid": "Chaque valeur de Puissance active doit être un nombre unique et valide.",
+    "ReactiveCapabilityCurveCreationErrorPOutOfRange": "Chaque valeur de Puissance active doit être comprise entre les valeurs P min et P max.",
+    "ReactiveCapabilityCurveCreationErrorMinPTooLow": "La valeur P min est incohérente avec la Puissance active min.",
+    "ReactiveCapabilityCurveCreationErrorMaxPTooHigh": "La valeur P max est incohérente avec la Puissance active max.",
+    "ReactiveCapabilityCurveCreationErrorQminPQmaxPIncoherence": "Chaque valeur de Puissance réactive minimale doit être inférieur ou égale à la Puissance réactive maximale correspondante.",
```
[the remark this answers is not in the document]
EstherDarkish: I used the same cases as the corresponding fields, I think it is easier to see what the message is talking about. But of course could change back to "normal" casing.
Do I change it back ?

src/translations/en.json
```diff
@@ -421,8 +421,10 @@
...
     "ReactiveCapabilityCurveCreationErrorMissingPoints": "The table of reactive capability curve must have at least four points.",
     "ReactiveCapabilityCurveCreationErrorPInvalid": "Each active power value must be a valid unique number.",
-    "ReactiveCapabilityCurveCreationErrorPOutOfRange": "Each active power value must be within the P min and P max values.",
-    "ReactiveCapabilityCurveCreationErrorQminPQmaxPIncoherence": "Each minimum reactive power value must be inferior or equal to the corresponding maximum reactive power value.",
+    "ReactiveCapabilityCurveCreationErrorPOutOfRange": "Each active power value must be within the Min P and Max P values.",
+    "ReactiveCapabilityCurveCreationErrorMinPTooLow": "The Min P value in incoherent with the Minimum active power value.",
+    "ReactiveCapabilityCurveCreationErrorMaxPTooHigh": "The Max P value in incoherent with the Maximimum active power value.",
+    "ReactiveCapabilityCurveCreationErrorQminPQmaxPIncoherence": "Each Minimum reactive power value must be inferior or equal to the corresponding Maximum reactive power value.",
```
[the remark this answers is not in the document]
EstherDarkish: I used the same cases as the corresponding fields, I think it is easier to see what the message is talking about. But of course could change back to "normal" casing.
Do I change it back ?

src/components/util/validation-functions.js
```diff
@@ -5,13 +5,42 @@
...
+        return /^\s*$/.test(value);
+    }
+    return false;
+}
+
+/*
+ * Returns true if the value is a valid number, per Gridsuite's standard (allows either coma or dots for decimal).
  */
 export function validateValueIsANumber(value) {
     if (value === undefined || value === '') {
```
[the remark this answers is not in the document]
EstherDarkish: I want it to be clear and easy to read

jonenst (approved):

PR #881, opened 2022-11-22
Fix error in spreadsheet when editing a transformer

YenguiSeddik:


laurgarn (approved):
Tested OK

src/components/network/equipment-table-editors.js
```diff
@@ -25,6 +25,21 @@ export const TapChangerSelector = ({
     defaultValue,
     ...props
 }) => {
+    const [arrayLength, setArrayLength] = useState(0);
+    const [value, setValue] = useState('');
```
jonenst, line 29: Why do we need a state for these ?
Is there really a need to handle the case where the props are like 5, then undefined, then 8, and we require that we still display 5 for the whole duration when the prop is undefined ?

EstherDarkish (changes requested):
The bug is not there anymore, but a new one shows its ugly head : when we change the value of the select field, then scroll to the left (enough to despawn the select field), then when we go back to the field, its possible values are not shown anymore.
The old value that was selected before the scroll is still saved, but there is no option for the user to change or review this change.

EstherDarkish:
Another issue is that when we edit a row that already have a value, the select input field do not display the current value (there is nothing pre-selected)

src/components/network/equipment-table-editors.js
YenguiSeddik: the problem here is when we have NaN as array length we get an error, so i added this state to check the value is not NaN before updating

Meklo (approved):
Tests : OK
Code : OK
Side note : We'll need to find a way to implement a more general way of setting properties in sub-objects from the spreadsheet

EstherDarkish (changes requested):
I now have the fatal error back : "invalid array length"

EstherDarkish (approved):
Ok

PR #887, opened 2022-11-29
Fix AutoComplete behaviour

niconoir:


laurgarn (approved):

src/components/dialogs/equipment-deletion-dialog.js
```diff
@@ -71,19 +111,29 @@ const EquipmentDeletionDialog = ({
...
-    const [searchMatchingEquipments, equipmentsFound] =
-        useSearchMatchingEquipments(
-            studyUuid,
-            currentNodeUuid,
-            true,
-            equipmentType,
-            makeItems
-        );
+    useEffect(() => {
+        setEquipmentsFound([]);
```
etiennehomer, line 122: We can avoid this set ?

src/components/dialogs/equipment-deletion-dialog.js
```diff
@@ -189,15 +238,20 @@ const EquipmentDeletionDialog = ({
...
-                            {equipmentTypes.map((item) => {
-                                return (
-                                    <MenuItem key={item} value={item}>
-                                        {intl.formatMessage({
-                                            id: item,
-                                        })}
-                                    </MenuItem>
-                                );
-                            })}
+                            {Object.entries(equipmentTypes).map(
```
etiennehomer, line 241: Object.values 
id is not used

etiennehomer (changes requested):

src/components/dialogs/equipment-deletion-dialog.js
```diff
@@ -71,19 +111,29 @@ const EquipmentDeletionDialog = ({
...
-        );
+    useEffect(() => {
+        setEquipmentsFound([]);
+        if (Array.isArray(equipmentType.fetcher)) {
+            Promise.all(
+                equipmentType.fetcher.map((fetchPromise) =>
+                    fetchPromise(studyUuid, currentNodeUuid)
+                )
+            ).then((vals) => {
+                setEquipmentsFound(vals?.flat().sort(compareById));
```
SlimaneAmar, line 129: vals may be null ?

src/components/dialogs/equipment-deletion-dialog.js
niconoir: If we remove this set, a glitch can appear when displaying equipments list in the autocomplete (when moving from one equipment type to another)

src/components/dialogs/equipment-deletion-dialog.js
niconoir: Right, corrected.

src/components/dialogs/equipment-deletion-dialog.js
```diff
@@ -4,53 +4,93 @@
...
-        .map((e) => {
-            let label = usesNames ? e.name : e.id;
-            return {
-                label: label,
-                id: e.id,
-                key: e.id,
-            };
-        })
-        .sort((a, b) => a.label.localeCompare(b.label));
+const equipmentTypes = {
```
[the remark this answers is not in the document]
niconoir: Currently only used in (and specific to) equipment-deletion-dialog

src/components/dialogs/equipment-deletion-dialog.js
niconoir: You're right, `vals` should not be null (removed the `?`). Also added a catch to handle fetching errors.

SlimaneAmar (approved):

etiennehomer (approved):

PR #888, opened 2022-11-29
Fix import regression

Meklo:


src/translations/en.json
```diff
@@ -559,8 +559,6 @@
     "ImportDephasingRule":"Import dephasing rule",
 
     "GenerateSkeleton": "Generate CSV skeleton",
-    "UploadCSV": "Uploadd CSV",
```
laurgarn, line 562: Upload, with only one 'd' at the end

Meklo: This locale was not needed so I removed it

laurgarn (approved):
Works

PR #890, opened 2022-11-30
Fix issue of subfolder not present in filter selection

YenguiSeddik:


FranckLecuyer (approved):

src/components/directory-item-selector.js
```diff
@@ -162,7 +158,7 @@ const DirectoryItemSelector = (props) => {
                     );
                 });
         },
-        [addToDirectory, contentFilter, props.equipmentTypes]
+        [addToDirectory, contentFilter, props]
```
archendel, line 161: sounds unsafe to have "props" as dependency. there are more properties than "types" and "equipmentTypes"
Also, react won't check if values of "props" object have changed, but only if the object reference is different between renders

Use props.types and props.equipmentTypes as dependencies

YenguiSeddik: Ok. done

archendel (approved):

PR #916, opened 2022-12-05
fix:allow 0 as current phase tap and negative target flow

laurgarn:


Meklo (approved):

src/components/dialogs/two-windings-transformer/two-windings-transformer-creation-dialog.js
```diff
@@ -708,7 +706,7 @@ const TwoWindingsTransformerCreationDialog = ({
                         id: 'IncoherentPhaseLowTapPositionError',
                     });
                 }
-                if (!tapValues.includes(phaseTapPosition)) {
+                if (!tapValues.includes(parseInt(phaseTapPosition))) {
```
dbraquart, line 709: with this fix, we can pass 0 : good.
side effect / regression : 
- before this fix, it was not possible to pass a double value, like 1.5 (we had a warning)
- with this fix, I think the double is cast to an integer: the test is ok, but "1.5" is sent to the back-end.

today, there is a numeric filter on this field (the user cannot enter letters for instance). Maybe we should use a more restrictive filter (only integer values, double not allowed).

dbraquart (changes requested):

src/components/dialogs/two-windings-transformer/two-windings-transformer-creation-dialog.js
laurgarn: There is another concern behind this side effect. All tap positions are currently handled through useDoubleValue allowing decimal values every where (you can even generate decimal, and so buggy, tap rows). Another US/TS is needed for these problems.
So I keep the code in this state for this PR.

dbraquart (approved):

PR #924, opened 2022-12-07
Fix two winding transformer tap position validation on edit or copy

Meklo:


src/components/dialogs/two-windings-transformer/two-windings-transformer-creation-dialog.js
```diff
@@ -373,7 +373,10 @@ const TwoWindingsTransformerCreationDialog = ({
         validation: {
             isFieldRequired: ratioTapChangerEnabled,
             valueGreaterThan: ratioLowTapPosition - 1,
-            valueLessThanOrEqualTo: ratioHighTapPosition,
+            valueLessThanOrEqualTo:
+                editData || isCopy
```
laurgarn, line 377: I think ratioHighTapPosition should be used in priority if it is defined. Don't you ?

Meklo: I agree, will change this

laurgarn (approved):

src/components/dialogs/two-windings-transformer/two-windings-transformer-creation-dialog.js
```diff
@@ -112,6 +112,20 @@ const TwoWindingsTransformerCreationDialog = ({
     const [ratioCellIndexError, setRatioCellIndexError] = useState(undefined);
     const [phaseCellIndexError, setPhaseCellIndexError] = useState(undefined);
 
+    const computePhaseHighTapPosition = () => {
```
laurgarn, line 115: Why define nearly twice the same function ?
You may well define one with a parameter and give it once phaseTapRows and ratioTapRows the other.
Especially as valueLessThanOrEqualTo is not expected to be a function, no ?

Meklo: There's no need, I implemented it your way. I thought it would cause issues at validation to pass a value computed at initialization instead of a reference but after testing it it works well

YenguiSeddik (approved):
Code Review: Ok
Test: Ok
No Console warnings

PR #926, opened 2022-12-07
Recalibrate Centering on Selected Node

anistouri:
Signed-off-by: TOURI ANIS <anis-1.touri@rte-france.com>


sBouzols (changes requested):
Seems not the good solution to change the DRAWER_NODE_EDITOR_WIDTH

PR #928, opened 2022-12-08
When sld reloads, we check if size has changed to refresh zoom or not

archendel:
Signed-off-by: LE SAULNIER Kevin <kevin.lesaulnier@rte-france.com>


src/components/diagrams/singleLineDiagram/single-line-diagram.js
```diff
@@ -296,6 +296,10 @@ const SingleLineDiagram = forwardRef((props, ref) => {
         []
     );
 
+    const hasSldSizeChanged = (oldWidth, newWidth, oldHeight, newHeight) => {
```
laurgarn, line 299: It is considered wise to avoid double negation.
There is a negation here and one at the unique call site.
Here you could inverse the test and name the function stillSameSizes, or even sameSizes.

archendel: I agree, fixed

YenguiSeddik (approved):

laurgarn (approved):

bendaamerahm (approved):

PR #930, opened 2022-12-08
Changing node won't reinitialize zoom unless SLD svg dimensions have changed

archendel:


YenguiSeddik:
It works correctly with SLD. But with NAD, the zoom still reinitialize when changing node

YenguiSeddik (approved):

PR #933, opened 2022-12-08
[WIP] fix: depth and full screen icon box hide when pane is bigger than given sizes

bendaamerahm:


bendaamerahm:
done in another pr

PR #936, opened 2022-12-09
Fix ratio and phase tap as double instead of integer in 2WT creation dialog

FranckLecuyer:


src/components/dialogs/two-windings-transformer/two-windings-transformer-creation-dialog.js
```diff
@@ -343,24 +344,31 @@ const TwoWindingsTransformerCreationDialog = ({
                 formValues?.ratioTapChanger?.regulatingTerminalId ?? '',
         });
 
-    const [ratioLowTapPosition, ratioLowTapPositionField] = useDoubleValue({
+    const [ratioLowTapPosition, ratioLowTapPositionField] = useIntegerValue({
         label: 'LowTapPosition',
         validation: {
             isFieldRequired: ratioTapChangerEnabled,
+            valueGreaterThan: '0',
```
laurgarn, line 351: Lower tap is allowed to be negative (from CGMES for example)

FranckLecuyer: Changes done

laurgarn (approved):

Meklo (approved):

bendaamerahm (approved):

PR #939, opened 2022-12-12
Fix unauthorised '0' value for tap position in transformer creation dialog

FranckLecuyer:


laurgarn (approved):
Code OK
Test OK
Console warnings OK

PR #955, opened 2022-12-14
Delete attaching line

bendaamerahm:


src/components/dialogs/delete-voltage-level-on-line.js
```diff
@@ -0,0 +1,198 @@
...
+        >
+            <GridSection title="Line1" />
+            <Grid container spacing={2} alignItems="center">
+                {gridItem(lineToAttachTo1Field, 5)}
+            </Grid>
+            <GridSection title="Line2" />
+            <Grid container spacing={2} alignItems="center">
+                {gridItem(lineToAttachTo2Field, 5)}
+            </Grid>
+            <GridSection title="ReplacingLines" />
```
dbraquart, line 173: the key is wrong, must be "ReplacingLine"

src/components/dialogs/delete-attaching-line-dialog.js
```diff
@@ -0,0 +1,215 @@
...
+            </Grid>
+            <GridSection title="Line2" />
+            <Grid container spacing={2} alignItems="center">
+                {gridItem(lineToAttachTo2Field, 5)}
+            </Grid>
+            <GridSection title="LineAttached" />
+            <Grid container spacing={2} alignItems="center">
+                {gridItem(attachedLineField, 5)}
+            </Grid>
+            <GridSection title="ReplacingLines" />
```
dbraquart, line 197: the key is wrong, must be "ReplacingLine"

src/components/dialogs/delete-attaching-line-dialog.js
```diff
@@ -0,0 +1,215 @@
...
+    const formValueAttachedLineId = useMemo(() => {
+        return formValues?.attachedLineId
+            ? { id: formValues?.attachedLineId }
+            : { id: '' };
+    }, [formValues]);
+
+    const [lineToAttachTo1, lineToAttachTo1Field] = useAutocompleteField({
+        id: 'lineToAttachTo1',
+        label: 'ID',
+        validation: { isFieldRequired: true },
```
dbraquart, line 86: the field is marked as mandatory, but it is possible to validate the dialog with an empty value for this field !

src/components/dialogs/delete-attaching-line-dialog.js
```diff
@@ -0,0 +1,215 @@
...
+            lineOptions.find(
+                (value) => value.id === formValues?.lineToAttachTo1Id
+            ) || formValueLineTo1AttachId,
+        loading: loadingLineOptions,
+    });
+
+    const [lineToAttachTo2, lineToAttachTo2Field] = useAutocompleteField({
+        id: 'lineToAttachTo2',
+        label: 'ID',
+        validation: { isFieldRequired: true },
```
dbraquart, line 101: the field is marked as mandatory, but it is possible to validate the dialog with an empty value for this field !

src/components/dialogs/delete-attaching-line-dialog.js
```diff
@@ -0,0 +1,215 @@
...
+            lineOptions.find(
+                (value) => value.id === formValues?.lineToAttachTo2Id
+            ) || formValueLineTo2AttachId,
+        loading: loadingLineOptions,
+    });
+
+    const [attachedLine, attachedLineField] = useAutocompleteField({
+        id: 'attachedLine',
+        label: 'ID',
+        validation: { isFieldRequired: true },
```
dbraquart, line 116: the field is marked as mandatory, but it is possible to validate the dialog with an empty value for this field !

src/utils/rest-api.js
```diff
@@ -2031,6 +2031,84 @@ export function linesAttachToSplitLines(
...
+    currentNodeUuid,
+    modificationUuid,
+    lineToAttachTo1Id,
+    lineToAttachTo2Id,
+    attachedLineId,
+    replacingLine1Id,
+    replacingLine1Name
+) {
+    const body = JSON.stringify({
+        type: MODIFICATION_TYPE.DELETE_VOLTAGE_LEVEL_ON_LINE,
```
dbraquart, line 2083: this type is Wrong!  should be DELETE_ATTACHING_LINE
so will create the wrong modification in the back / database

src/utils/rest-api.js
```diff
@@ -2031,6 +2031,84 @@ export function linesAttachToSplitLines(
...
+        attachedLineId,
+        replacingLine1Id,
+        replacingLine1Name,
+    });
+
+    let deleteVoltageLevelOnLineUrl =
+        getStudyUrlWithNodeUuid(studyUuid, currentNodeUuid) +
+        '/network-modifications';
+    if (modificationUuid) {
+        console.info('deleting voltage level on line update', body);
```
dbraquart, line 2095: log should be:  'Updating delete attaching line'

src/utils/rest-api.js
```diff
@@ -2031,6 +2031,84 @@ export function linesAttachToSplitLines(
...
+        deleteVoltageLevelOnLineUrl +=
+            '/' + encodeURIComponent(modificationUuid);
+    } else {
+        console.info('Delete voltage level on line', body);
```
dbraquart, line 2099: log should be:  'Creating delete attaching line'

dbraquart (changes requested):
actually it is not possible to create the correct DELETE_ATTACHING_LINE modification

src/components/dialogs/delete-attaching-line-dialog.js
bendaamerahm: yes it's mandatory, so the validation gonna fail if the field is empty

src/components/dialogs/delete-attaching-line-dialog.js
bendaamerahm: yes it's mandatory, so the validation gonna fail if the field is empty

src/components/dialogs/delete-attaching-line-dialog.js
bendaamerahm: yes it's mandatory, so the validation gonna fail if the field is empty

src/utils/rest-api.js
bendaamerahm: my bad

src/components/dialogs/delete-attaching-line-dialog.js
dbraquart: the validation does not fail for this field.
test case: create a new modification; fill only the optional field : only one field seems to be mandatory.
So we can send an empty value for 'lineToAttachTo1'

src/components/dialogs/delete-attaching-line-dialog.js
dbraquart: the validation does not fail for this field.
test case: create a new modification; fill only the optional field : only one field seems to be mandatory.
So we can send an empty value for 'lineToAttachTo2'

src/components/dialogs/delete-attaching-line-dialog.js
dbraquart: the validation does not fail for this field.
test case: create a new modification; fill only the optional field : only one field seems to be mandatory.
So we can send an empty value for 'attachedLine'

src/utils/rest-api.js
```diff
@@ -2039,6 +2039,84 @@ export function linesAttachToSplitLines(
...
+        attachedLineId,
+        replacingLine1Id,
+        replacingLine1Name,
+    });
+
+    let deleteVoltageLevelOnLineUrl =
+        getStudyUrlWithNodeUuid(studyUuid, currentNodeUuid) +
+        '/network-modifications';
+    if (modificationUuid) {
+        console.info('deleting attaching line update', body);
```
dbraquart, line 2103: log should be: 'Updating delete attaching line'

src/utils/rest-api.js
```diff
@@ -2039,6 +2039,84 @@ export function linesAttachToSplitLines(
...
+        deleteVoltageLevelOnLineUrl +=
+            '/' + encodeURIComponent(modificationUuid);
+    } else {
+        console.info('Delete attaching line', body);
```
dbraquart, line 2107: log should be: 'Creating delete attaching line'

dbraquart:
console error Error: Missing message: "DeleteAttachingLineError" for locale "en", using id as fallback.
=> DeleteAttachingLineError translation key does not exist in en/fr

src/components/graph/menus/modification-list-item.js
```diff
@@ -73,6 +73,10 @@ export const ModificationListItem = ({
             return modif.lineToAttachToId;
         } else if (modif.type === 'LINES_ATTACH_TO_SPLIT_LINES') {
             return modif.attachedLineId;
+        } else if (modif.type === 'DELETE_VOLTAGE_LEVEL_ON_LINE') {
+            return modif.attachedLineId;
+        } else if (modif.type === 'DELETE_ATTACHNING_LINE') {
```
dbraquart, line 78: there is a typo here: DELETE_ATTACHNING_LINE -> DELETE_ATTACHING_LINE
as a consequence, the modification name has no computed label

dbraquart (approved):
review ok
tests ok

src/components/dialogs/delete-voltage-level-on-line.js
```diff
@@ -0,0 +1,198 @@
...
+
+    useEffect(() => {
+        if (!lineOptionsPromise) return;
+        lineOptionsPromise.then((values) => {
+            setLineOptions(values);
+            setLoadingLineOptions(false);
+        });
+    }, [lineOptionsPromise]);
+
+    const formValueLineTo1AttachId = useMemo(() => {
```
flomillot, line 67: All these useMemo are really useful here ? I don't see any benefit, on the contrary, but maybe I missed something.

bendaamerahm: this is to avoid rerenders in our hooks, we use this with useAutocompleteField default value

flomillot: Ah indeed it's used in an autocomplete I missed it, my bad

flomillot (approved):

PR #957, opened 2022-12-15
Fix ratio and phase tap position pb

FranckLecuyer:


src/components/dialogs/two-windings-transformer/phase-tap-changer-pane.js
```diff
@@ -520,7 +520,10 @@ const PhaseTapChangerPane = (props) => {
                                 onClick={() => generateTapRows()}
                                 disabled={
                                     !phaseTapChangerEnabled ||
-                                    !(lowTapPosition && highTapPosition)
+                                    !(
```
laurgarn, line 523: Could simplify to `lowTapPosition === '' || highTapPosition === ''`, that would avoid double negation.

src/components/dialogs/two-windings-transformer/ratio-tap-changer-pane.js
```diff
@@ -505,7 +505,10 @@ const RatioTapChangerPane = (props) => {
                                 onClick={() => generateTapRows()}
                                 disabled={
                                     !ratioTapChangerEnabled ||
-                                    !(lowTapPosition && highTapPosition)
+                                    !(
```
laurgarn, line 508: Same as above

src/components/dialogs/two-windings-transformer/phase-tap-changer-pane.js
FranckLecuyer: you're right
Done

laurgarn (approved):

PR #959, opened 2022-12-15
fix sld display when selecting a node not built

YenguiSeddik:
When working on node not built, opening multiple SLD can lead to some of them being displayed out of the screen because we take the size of the last sld,
This PR is to make sure when opening multiple sld and the current node is not built, each sld will be displayed inside the screen.


EstherDarkish (approved):

EstherDarkish:
How to reproduce : On a built node, open two big SLD, then minimize them.
After that, go to an unbuilt node, then open one of the big minimized SLD, pin it, and open the other. After the fix, instead of a second SLD that goes beyond the limit of the screen, they both share the available space.

sBouzols (approved):
Code Review OK
Tests OK
division by 0 could be suspicious...

PR #961, opened 2022-12-16
Refacto NAD/SLD to use only one pane for both

EstherDarkish:


src/components/diagrams/diagram-pane.js
```diff
@@ -0,0 +1,578 @@
...
+                    );
+                }
+
+                if (displayedVoltageLevels.length === 0) return;
+                displayedVoltageLevels.forEach((voltageLevel) => {
+                    const name =
+                        paramUseName && voltageLevel?.name
+                            ? voltageLevel?.name
+                            : voltageLevel?.id;
+                    if (name !== undefined) {
```
anistouri, line 179: you can remplace it with `if (name != null){ `

src/components/diagrams/diagram-common.js
```diff
@@ -0,0 +1,323 @@
...
+export const computePaperAndSvgSizesIfReady = (
+    isFullScreenActive,
+    svgType,
+    totalWidth,
+    totalHeight,
+    svgPreferredWidth,
+    svgPreferredHeight,
+    headerPreferredHeight
+) => {
+    if (
```
anistouri, line 282: you can replace it by `svgPreferredWidth != null &&
         headerPreferredHeight != null`
!= null check also undefined

src/components/diagrams/diagram-common.js
```diff
@@ -0,0 +1,323 @@
...
+                tempMaxWidth = MAX_WIDTH_NETWORK_AREA_DIAGRAM;
+                tempMaxHeight = MAX_HEIGHT_NETWORK_AREA_DIAGRAM;
+            } else {
+                console.error('type inconnu svgType');
+            }
+            svgWidth = Math.min(
+                svgPreferredWidth,
+                svgType !== SvgType.NETWORK_AREA_DIAGRAM
+                    ? totalWidth //- MAP_RIGHT_OFFSET
+                    : totalWidth, // MAP_RIGHT_OFFSET = 120 pour SLD, 0 pour NAD
```
anistouri, line 310: why do you have a condition ?

src/components/diagrams/diagram.js
```diff
@@ -0,0 +1,911 @@
...
+        setDisplayedDiagramHeights,
+        diagramId,
+        svgType,
+        //onClose,
+        setDepth,
+        totalHeight,
+        totalWidth,
+    } = props;
+
+    const [, , , closeDiagramView] = useDiagram(); // We only use this function (the fourth)
```
anistouri, line 121: you can replace by `const { closeDiagramView } = useDiagram();`

src/components/diagrams/diagram.js
```diff
@@ -0,0 +1,911 @@
...
+        disabled,
+        modificationInProgress,
+        loadingState,
+        locallySwitchedBreaker,
+        diagramType,
+    ]);
+
+    useLayoutEffect(() => {
+        if (
+            typeof svgFinalWidth != 'undefined' &&
```
anistouri, line 558: the same remark

src/components/diagrams/diagram.js
```diff
@@ -0,0 +1,911 @@
...
+                if (svgEl != null) {
+                    svgEl.setAttribute('width', svgFinalWidth);
+                    svgEl.setAttribute(
+                        'height',
+                        computedHeight ? computedHeight : svgFinalHeight
+                    );
+                }
+            }
+            setModificationInProgress(false);
+        } else {
```
anistouri, line 573: To remove

src/components/diagrams/diagram.js
```diff
@@ -0,0 +1,911 @@
...
+            typeof svgFinalHeight != 'undefined'
+        ) {
+            const divElt = svgRef.current;
+            if (divElt != null) {
+                const svgEl = divElt.getElementsByTagName('svg')[0];
+                if (svgEl != null) {
+                    svgEl.setAttribute('width', svgFinalWidth);
+                    svgEl.setAttribute(
+                        'height',
+                        computedHeight ? computedHeight : svgFinalHeight
```
anistouri, line 568: you can remplace it by `computedHeight ?? svgFinalHeight`

src/components/diagrams/diagram.js
```diff
@@ -0,0 +1,911 @@
...
+            )
+        );
+    };
+
+    let sizeWidth,
+        sizeHeight = initialHeight;
+    if (svg.error) {
+        sizeWidth = errorWidth;
+    } else if (
+        typeof finalPaperWidth != 'undefined' &&
```
anistouri, line 659: the same remark

src/components/diagrams/diagram.js
EstherDarkish: Yes, and I also changed the return value of useDiagram.

src/components/diagrams/diagram-common.js
```diff
@@ -0,0 +1,323 @@
...
+        typeof headerPreferredHeight != 'undefined'
+    ) {
+        let paperWidth, paperHeight, svgWidth, svgHeight;
+        if (isFullScreenActive) {
+            paperWidth = totalWidth;
+            paperHeight = totalHeight;
+            svgWidth = totalWidth - BORDERS;
+            svgHeight = totalHeight - headerPreferredHeight - BORDERS;
+        } else {
+            let tempMaxWidth, tempMaxHeight;
```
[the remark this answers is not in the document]
EstherDarkish: Done

src/components/diagrams/diagram-pane.js
EstherDarkish: done

src/components/diagrams/diagram-common.js
EstherDarkish: This is from old code. It is now removed.

src/components/diagrams/diagram.js
EstherDarkish: done

src/components/diagrams/diagram.js
EstherDarkish: done

src/components/diagrams/diagram.js
EstherDarkish: done

src/redux/actions.js
```diff
@@ -502,7 +503,7 @@ export function togglePinSld(id) {
     };
 }
 
-export const CLOSE_SLD = 'CLOSE_SLD';
+export const CLOSE_SLD = 'CLOSE_SLD'; // TODO to remove after the SLD/NAD refactorization
```
flomillot, line 506: It's not what you're doing, there is an other refactorization after ?

flomillot (approved):

flomillot:
Clean !

src/components/app-top-bar.js
```diff
@@ -232,12 +224,12 @@ const AppTopBar = ({ user, tabIndex, onChangeTab, userManager }) => {
         (optionInfos) => {
             onChangeTab(STUDY_VIEWS.indexOf(StudyView.MAP)); // switch to map view
             if (optionInfos.type === EQUIPMENT_TYPE.SUBSTATION.name) {
```
etiennehomer, line 226: openDiagramView(optionInfos.id, optionInfos.type); ?
So you can remove the "if"

src/components/study-pane.js
```diff
@@ -155,20 +145,20 @@ const StudyPane = ({
         }
     }, [network, filteredNominalVoltages, dispatch]);
 
-    function openVoltageLevelDiagram(vlId, substationId) {
+    function openVoltageLevelDiagram(vlId) {
```
etiennehomer, line 148: Could be openVoltageLevelDiagramFromResults

src/components/study-pane.js
```diff
@@ -155,20 +145,20 @@ const StudyPane = ({
...
         // TODO code factorization for displaying a VL via a hook
         if (vlId) {
             props.onChangeTab(0); // switch to map view
-            showVoltageLevelDiagram(vlId); // show voltage level
+            openDiagramView(vlId, SvgType.VOLTAGE_LEVEL);
         }
     }
 
     const openVoltageLevel = useCallback(
```
etiennehomer, line 156: openVoltageLevelDiagramFromMap ?
Those names are a bit long but that could be benefic to make a difference between the 2 methods

src/components/diagrams/diagram-common.js
```diff
@@ -0,0 +1,274 @@
...
+        case 'THREE_WINDINGS_TRANSFORMER_LEG':
+            return equipments.threeWindingsTransformers;
+        default: {
+            console.log('bad feeder type ', feederType);
+            return null;
+        }
+    }
+}
+
+export const useDiagram = () => {
```
bendaamerahm, line 170: i suggest to optimise useDiagram by making some changes : 

- Combine similar useCallback functions: The openDiagramView, togglePinDiagramView, minimizeDiagramView, and closeDiagramView functions all dispatch the same type of action, but with different parameters. These functions can be combined into a single function that takes an additional parameter for the type of action to dispatch (open, togglePin, minimize, or close).

- Use the useEffect hook with dependencies: Currently, the useEffect hook is listening to changes in the diagramStates and studyUuid, but these values are not used inside the callback function. 
Instead, you can use the useEffect hook with a dependency array to only call the syncDiagramStateWithSessionStorage function when the specific action that causes the state to change is dispatched.

- Use useMemo instead of useCallback: Since all the functions are only calling dispatch with the same set of dependencies, useMemo is a more appropriate hook.

- Use a switch case inside the callback function of useEffect to call the appropriate function based on the action type

- consider using reselect to avoid recomputing the same thing everytime

- consider using Immer to update state immutablely

here's an example how this function can be:

```
export const useDiagram = () => {
    const dispatch = useDispatch();
    const {diagramStates, studyUuid, action} = useSelector((state) => ({
        diagramStates: state.diagramStates,
        studyUuid: state.studyUuid,
        action: state.action
    }));

    useEffect(() => {
        if(action === 'SYNC_DIAGRAM_STATE'){
           syncDiagramStateWithSessionStorage(diagramStates, studyUuid);
        }
    }, [action]);

    const handleDiagramView = useMemo(() => (id, type, action) => {
        dispatch({
            type: action,
            payload: {id, type}
        });
    }, [dispatch]);

    return {
        openDiagramView: (id, type) => handleDiagramView(id, type, 'OPEN_DIAGRAM'),
        minimizeDiagramView: (id, type) => handleDiagramView(id, type, 'MINIMIZE_DIAGRAM'),
        togglePinDiagramView: (id, type) => handleDiagramView(id, type, 'TOGGLE_PIN_DIAGRAM'),
        closeDiagramView: (id, type) => handleDiagramView(id, type, 'CLOSE_DIAGRAM'),
        closeDiagramViews: (idsToRemove) => handleDiagramView(idsToRemove, null, 'CLOSE_DIAGRAMS'),
    };
};
```

src/components/diagrams/diagram-footer.js
```diff
@@ -0,0 +1,127 @@
...
+    },
+    fullScreenIcon: {
+        bottom: theme.spacing(1),
+        right: theme.spacing(1),
+        position: 'absolute',
+        cursor: 'pointer',
+    },
+}));
+
+const DiagramFooter = (props) => {
```
bendaamerahm, line 42: - i suggest to Move the logic of checking the props exist outside of the callback functions, this check can be moved outside of the callback function to reduce the number of times of check.

- Use the useCallback hook to memoize the callback functions, our callback functions are passed as props, they will be recreated on every render of the parent component, by using the useCallback we can prevent unnecessary re-renders.

can be something like this: 

```
    const classes = useStyles();
    const {onStopFullScreen, onStartFullScreen, onIncrementCounter, onDecrementCounter} = props;
    const handleStopFullScreen = useCallback(() => onStopFullScreen && onStopFullScreen(),[onStopFullScreen]);
    const handleStartFullScreen = useCallback(() => onStartFullScreen && onStartFullScreen(),[onStartFullScreen]);
    const handleIncrementCounter = useCallback(() => onIncrementCounter && onIncrementCounter(),[onIncrementCounter]);
    const handleDecrementCounter = useCallback(() => onDecrementCounter && onDecrementCounter(),[onDecrementCounter]);
```

src/components/diagrams/diagram-footer.js
```diff
@@ -0,0 +1,127 @@
...
+            {props.showCounterControls && (
+                <>
+                    {props.showCounterValue && (
+                        <Typography className={classes.counterText}>
+                            {props.counterText + props.counterValue}
+                        </Typography>
+                    )}
+                    <AddCircleIcon
+                        onClick={handleIncrementCounter}
+                        className={classes.plusIcon}
```
bendaamerahm, line 77: Use a more specific classname for the icons,  instead of using classes like plusIcon and lessIcon, use more specific classnames that describe the context of the icons.

`className={classes.incrementCounterIcon}` and   `className={classes.decrementCounterIcon}` for example

src/components/diagrams/diagram-header.js
```diff
@@ -0,0 +1,127 @@
...
+    close: {
+        padding: 0,
+        borderRight: theme.spacing(1),
+    },
+}));
+
+const DiagramHeader = (props) => {
+    const classes = useStyles();
+
+    const handleMinimize = () => {
```
bendaamerahm, line 44: same suggestion as diagram-footer

src/components/app-top-bar.js
EstherDarkish: `EQUIPMENT_TYPE.SUBSTATION.name` and `SvgType.SUBSTATION` could be different.
This code is not mine, I don't think I should modify it in this refactoring.

src/components/diagrams/diagram-pane.js
```diff
@@ -0,0 +1,600 @@
...
+    isNodeBuilt,
+    isNodeInNotificationList,
+} from '../graph/util/model-functions';
+import { AutoSizer } from 'react-virtualized';
+import Diagram from './diagram';
+import { SLD_DISPLAY_MODE } from '../network/constants';
+import clsx from 'clsx';
+import { useNameOrId } from '../util/equipmentInfosHandler';
+
+const useDisplayView = (network, studyUuid, currentNode) => {
```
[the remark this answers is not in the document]
EstherDarkish: Yes, maybe for another US or TS.

src/components/study-pane.js
EstherDarkish: This is not my code, I don't think I should modify it in this refactoring. Maybe in another US/TS

src/components/study-pane.js
EstherDarkish: This is not my code, I don't think I should modify it in this refactoring. Maybe in another US/TS

etiennehomer:
To me, it's a problem that the SLD and NAD are fully mingled in the code. Moving them in a unique pane and managing their sizes together is a good thing, but they also have different behaviours and we should keep them independant

src/components/diagrams/diagram-common.js
EstherDarkish: Maybe in another TS, this would complexify the current refactoring too much.

src/components/diagrams/diagram-footer.js
EstherDarkish: Yes

src/components/diagrams/diagram-header.js
EstherDarkish: Done

src/components/diagrams/diagram.js
```diff
@@ -0,0 +1,870 @@
...
+
+    const currentNode = useSelector((state) => state.currentTreeNode);
+
+    const fullScreenDiagram = useSelector((state) => state.fullScreenDiagram);
+
+    const notificationIdList = useSelector((state) => state.notificationIdList);
+
+    const [forceState, updateState] = useState(false);
+
+    const networkAreaDiagramDepth = useSelector(
```
etiennehomer, line 123: That would be nice to separate hooks and function required for SLD or for NAD. I think it's  a bit damaging to have them in the same component and file

EstherDarkish:
> To me, it's a problem that the SLD and NAD are fully mingled in the code. Moving them in a unique pane and managing their sizes together is a good thing, but they also have different behaviours and we should keep them independant

I agree. I tried several time to externalize their stuff in separate files, but without success. Each time it creates a load of bugs, which, I think, are related to the references that are passed between components.
I think we should try again to refactorize this aspect, but maybe not today and in this current US, as this would take (me) too much time.

archendel (commented):
Test: OK except some behaviours we've discussed 
Code: OK

archendel (approved):

src/components/diagrams/diagram-pane.js
```diff
@@ -0,0 +1,602 @@
...
+     */
+
+    const [computedHeight, setComputedHeight] = useState();
+
+    useEffect(() => {
+        let displayedDiagramHeights_ =
+            displayedDiagramHeightsRef.current?.filter((displayedHeight) =>
+                views
+                    .filter((sld) => sld.state !== ViewState.MINIMIZED)
+                    .map((sld) => sld.id)
```
etiennehomer, line 476: Here, we'll have the same id for the SLD and the NAD of a same voltage level

src/components/diagrams/diagram-pane.js
```diff
@@ -0,0 +1,602 @@
...
+                                    SvgType.NETWORK_AREA_DIAGRAM && (
+                                    <div className={classes.separator}></div>
+                                )
+                            }
+                            <Diagram
+                                diagramTitle={diagramView.name}
+                                ref={diagramView.ref}
+                                disabled={disabled}
+                                pinned={diagramView.state === ViewState.PINNED}
+                                diagramId={diagramView.id}
```
etiennehomer, line 529: Same here. The type is not used to build a unique id

src/components/diagrams/diagram.js
```diff
@@ -0,0 +1,870 @@
...
+        dispatch(decrementNetworkAreaDiagramDepth());
+    };
+
+    /**
+     * RENDER
+     */
+
+    return !svg.error ? (
+        <Paper
+            ref={ref}
```
jonenst, line 698: I don't think this is right

src/components/diagrams/diagram.js
```diff
@@ -0,0 +1,870 @@
...
+            elevation={4}
+            square={true}
+            className={classes.paperBorders}
+            style={{
+                flewGrow: 0, // allows a separator to take all the available space
```
jonenst, line 703: typo ??

EstherDarkish: Ok, will change the comment to "allows a separator (in the parent pane) to take all the available space"

src/components/diagrams/diagram-pane.js
EstherDarkish: Thanks, it is now fixed.

src/components/diagrams/diagram.js
EstherDarkish: After testing you are correct, I removed it.

src/components/diagrams/diagram.js
EstherDarkish: Agree, but maybe not today and in this current US, as this would take (me) too much time.

src/components/diagrams/diagram.js
etiennehomer: fleXGrow

EstherDarkish: Ok my bad, you are correct, this is a typo. And it was not needed because 0 is already the default value. I removed it.

jonenst (approved):

PR #968, opened 2022-12-22
Display the errors from the notifications

flomillot:


sBouzols (approved):
Code Review OK
Seems like there are different syntax for translation keys
lower and upper Camel Case... we will need to fix that later

src/components/study-container.js
```diff
@@ -242,31 +245,38 @@ export function StudyContainer({ view, onChangeTab }) {
 
     const wsRef = useRef();
 
-    const checkFailNotifications = useCallback(
+    const displayErrorNotifications = useCallback(
         (eventData) => {
             const updateTypeHeader = eventData.headers[UPDATE_TYPE_HEADER];
+            const errorMessage = eventData.headers[ERROR_HEADER];
+            const userId = eventData.headers[USER_HEADER];
+            if (userId !== userName) return;
```
SlimaneAmar, line 253: Perhaps test if `eventData.headers[USER_HEADER] ` is not null ?

src/components/study-container.js
```diff
@@ -182,6 +184,7 @@ export function StudyContainer({ view, onChangeTab }) {
     const studyParentDirectoriesUuidsRef = useRef([]);
 
     const network = useSelector((state) => state.network);
+    const userName = useSelector((state) => state.user.profile.name);
```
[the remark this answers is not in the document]
flomillot: Ah maybe, what do you use for auth, can I have some documentation or origin code ?
In this case it's sub ?

src/components/study-container.js
flomillot: Hmm why, in this case it will just satisfy the condition and return directly no ?

src/components/study-container.js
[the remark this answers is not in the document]
SlimaneAmar: yes

src/components/study-container.js
SlimaneAmar: ok

PR #969, opened 2022-12-23
Fix us 1630

FranckLecuyer:


src/components/diagrams/singleLineDiagram/single-line-diagram.js
```diff
@@ -403,7 +416,15 @@ const SingleLineDiagram = forwardRef((props, ref) => {
...
-    }, [props.svgUrl, forceState, snackError, intlRef, sldId]);
+    }, [
+        props.svgUrl,
+        forceState,
+        snackError,
+        intlRef,
+        sldId,
+        currentNode,
+        notificationIdList,
+    ]);
```
EstherDarkish, line 427: If we use isNodeInNotificationList as a dependency instead of notificationIdList, could this will prevent a reload of all the SLD each time someone else works on another unrelated node ?

FranckLecuyer: Done

EstherDarkish (approved):

PR #975, opened 2022-12-28
WIP Update the startup section of the generator creation dialog.

flomillot:
#1474


src/components/dialogs/generator-creation-dialog.js
```diff
@@ -194,7 +198,6 @@ const GeneratorCreationDialog = ({
         label: 'MaximumActivePowerText',
         validation: {
             isFieldRequired: true,
-            isFieldNumeric: true,
```
Meklo, line 197: You removed isFieldNumeric from the validation of these fields because the useDoubleValue already prevent text prompts ?

Meklo:
When I open the generator creation dialog I get multiple missing translations error fort the reactive capability curve table fields  like the following one : 
`Error: [@formatjs/intl Error MISSING_TRANSLATION] Missing message: "Pmin" for locale "en", using id as fallback.`
It doesn't occur on main, do you get those too ?

Meklo:
If you get a notification of a comment I left about missing translations errors you can ignore it, it does occur on main in fact

SlimaneAmar (approved):

src/components/dialogs/generator-creation-dialog.js
flomillot: Because it's already added by the hook useDoubleValue

Meklo (approved):

PR #986, opened 2023-01-05
fix map display on root node at startup

Walid-Sahnoun:
Signed-off-by: walid-sahnoun <walid.sahnoun@rte-france.com>

### DIFF
--- a/src/components/network-map-tab.js
+++ b/src/components/network-map-tab.js
@@ -35,7 +35,7 @@ import {
     isNodeRenamed,
 } from './graph/util/model-functions';
 import { RunningStatus } from './util/running-status';
-import { mapEquipmentsCreated, resetMapReloaded } from '../redux/actions';
+import { resetMapReloaded } from '../redux/actions';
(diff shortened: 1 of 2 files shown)


Meklo:
The bug was occuring not because of the value of lineFullPath itself but its underlying implications.
When lineFullPath is enabled, the map needs ~25MB of geo data dedicated to the lines in order to load properly, which leaves ample time for mapEquipements cache to load its data.
In consequence when it's not enabled, the map, and its layers, can load way quicker (since geo-data and network-map-data are loaded asynchroneously) so mapEquipments data was not ready to use yet.

In order for the layers to update its render when needed, mapEquipments has to notify when its data has finished to load which was not the case before because mapEquipmentsCreated was called not in sync with it.
This change fixes that

Meklo:
NB : we don't need a variable to store `new MapEquipments` in network-map-tab because `this.dispatch(mapEquipmentsCreated(this))` does the same job

Meklo (approved):

FranckLecuyer (approved):

PR #987, opened 2023-01-05
Change node layout during cut and paste process

dbraquart:


ayolab (commented):
Tests OK
Code seems OK

ayolab (approved):

src/components/graph/menus/create-node-menu.js
```diff
@@ -88,8 +92,18 @@ const CreateNodeMenu = ({
...
+            (selectedNodeForCopy.nodeId !== activeNode.id ||
+                selectedNodeForCopy.copyType !== CopyType.CUT)
+        );
+    }
+
+    function isAlreadySelectedForCut() {
+        return (
+            selectedNodeForCopy &&
+            selectedNodeForCopy.nodeId === activeNode.id &&
+            selectedNodeForCopy.copyType === CopyType.CUT
```
archendel, line 106:
```suggestion
            selectedNodeForCopy?.nodeId === activeNode.id &&
            selectedNodeForCopy?.copyType === CopyType.CUT
```

src/components/graph/menus/create-node-menu.js
```diff
@@ -125,8 +139,13 @@ const CreateNodeMenu = ({
...
             onRoot: false,
-            action: () => cutNetworkModificationNode(),
-            id: 'cutNetworkModificationNode',
+            action: () =>
+                isAlreadySelectedForCut()
+                    ? cancelCutNetworkModificationNode()
+                    : cutNetworkModificationNode(),
+            id: isAlreadySelectedForCut()
+                ? 'cancelCutNetworkModificationNode'
+                : 'cutNetworkModificationNode',
```
archendel, line 148: I think it would be cleaner to create a new menu entry instead

src/components/graph/nodes/network-modification-node.js
```diff
@@ -90,12 +93,30 @@ const NetworkModificationNode = (props) => {
     const classes = useStyles();
 
     const currentNode = useSelector((state) => state.currentTreeNode);
+    const selectedNodeForCopy = useSelector(
+        (state) => state.selectedNodeForCopy
+    );
```
archendel, line 98: I am wondering is it is really needed to use redux to do this, maybe we could just pass a prop to the right child

src/components/network-modification-tree-pane.js
```diff
@@ -87,10 +90,9 @@ export const NetworkModificationTreePane = ({
     const DownloadIframe = 'downloadIframe';
 
     const [activeNode, setActiveNode] = useState(null);
-    const [selectedNodeIdForCopy, setSelectedNodeIdForCopy] = useState(null);
-    const [copyType, setCopyType] = useState(null);
-    const selectedNodeIdForCopyRef = useRef();
-    selectedNodeIdForCopyRef.current = selectedNodeIdForCopy;
+    const [nodeForCopy, setNodeForCopy] = useState(noSelectionForCopy);
```
archendel, line 93: Since we now have this info in the redux, do we still need this state ?

archendel (changes requested):

src/components/graph/menus/create-node-menu.js
dbraquart: we want a dynamic menu entry here (to avoid 2 entries that could be enable / disable).
So I think this is the easier way to do it

src/components/graph/nodes/network-modification-node.js
dbraquart: Seems tricky to update a given node (in the treeNodes model) from network-modification-tree-pane.js, to pass this data as a prop.
I'd rather use dispatch and redux.

src/components/network-modification-tree-pane.js
dbraquart: good point!
if the state is in redux, it is useless here.

PR #989, opened 2023-01-06
Feat add substation properties fix2

laurgarn:


src/components/dialogs/inputs/use-autocomplete-field.js
```diff
@@ -109,20 +109,18 @@ export const useAutocompleteField = ({
...
+            const res = validateField(getLabel(value), validationRef.current);
             setError(res?.errorMsgId);
             return !res.error;
         }
 
-        if (inputForm) {
-            inputForm.addValidation(id ? id : label, validate);
-        }
-    }, [label, validation, inputForm, value, selectedValue, id]);
+        inputForm?.addValidation(id ? id : label, validate);
```
anistouri, line 117: you can remplace this `id ? id : label` by `id ?? label`

laurgarn: It would not be the same if id === ''
But id || label would do.

PR #992, opened 2023-01-09
add 2wt regulated terminal

AbdelHedhili:


AbdelHedhili:
There are a lot a warning triggered in the console when handling the creation form, any idea where does it come from ? I don't have a clue.

bendaamerahm (approved):
code review ok

src/translations/en.json
```diff
@@ -432,6 +432,7 @@
 
     "CreateLine": "Create line",
     "Characteristics": "Characteristics",
+    "RegulatedSide": "regulated side",
```
archendel, line 435:
```suggestion
    "RegulatedSide": "Regulated side",
```

src/translations/fr.json
```diff
@@ -432,6 +432,7 @@
 
     "CreateLine": "Créer une ligne",
     "Characteristics": "Caractéristiques",
+    "RegulatedSide": "côté reglé",
```
archendel, line 435:
```suggestion
    "RegulatedSide": "Côté réglé",
```

archendel (changes requested):
between "ratio tap changer" tab and "phase tap changer" tab, there is a small difference

when Voltage Regulation is unchecked, "regulated side" is marked (optional) in "phase tap changer" tab, but not in the other one

now testing full workflow

PR #993, opened 2023-01-09
Line menu: disable a menu item if we already are in the target state (instead of highlighting)

dbraquart:


FranckLecuyer (approved):

dbraquart:
With commit [8586cc3](https://github.com/gridsuite/gridstudy-app/pull/993/commits/8586cc362e56103449c2d0721cac2c3e255347d6) we disable a line menu item if the line already fits the target state. We use the same conditions previously used the mark the menu items as 'selected'.

FranckLecuyer (approved):

laurgarn (approved):

PR #997, opened 2023-01-10
fix nad zoom

YenguiSeddik:


EstherDarkish (changes requested):
If we unzoom a bit too much, the NAD totally disparears and do not come back by zooming again.

EstherDarkish (approved):

bendaamerahm (approved):

PR #1013, opened 2023-01-16
Separate the renders of the map

flomillot:
TS 1691


src/components/network/geo-data.js
```diff
@@ -115,6 +115,7 @@ export default class GeoData {
             ];
         }
 
+        // why it's not at the start of the function ?
```
etiennehomer, line 118: Yes it can be refacto with the test at the start

src/components/study-pane.js
```diff
@@ -262,6 +262,7 @@ const StudyPane = ({
                                 {/* TODO do not display if study does not exists or do not fetch geoData if study does not exists */}
                                 <NetworkMapTab
                                     /* TODO do we move redux param to container */
+                                    /* why passing params as props if it's stored in redux and only used by this component ? */
```
etiennehomer, line 265: I think there's no reason

src/components/network/geo-data.js
flomillot: Ok I'll do it in an other PR

src/components/network/network-map.js
```diff
@@ -445,9 +451,10 @@ const NetworkMap = (props) => {
 };
 
 NetworkMap.defaultProps = {
-    mapEquipments: null,
+    network: null,
+    substations: [],
+    lines: [],
```
etiennehomer, line 456: You can remove substations and lines nope ?

jonenst (approved):

etiennehomer (approved):

PR #1014, opened 2023-01-16
Feat add substation properties fix3

laurgarn:


src/utils/rest-api.js
```diff
@@ -1887,7 +1887,7 @@ export function createSubstation(
         substationCountry: substationCountry === '' ? null : substationCountry,
         properties: asObj,
     });
-    console.debug('createSubstation body', properties, body);
+    console.debug({ body });
```
AbdelHedhili, line 1890: a little message can be nice to understand better what we see in the console

AbdelHedhili (approved):
code ; ok 
not tested

src/utils/rest-api.js
laurgarn: There is `console.info('Creating substation creation');` some lines above, so I thought it could suffice.

AbdelHedhili: I was talking about the fact that we don't know that it's the body, see the previous message that you changed it was better

AbdelHedhili: I was talking about the fact that we don't know that it's the body, see the previous message that you changed it was better

dbraquart (approved):

PR #1018, opened 2023-01-17
Fixes a fullscreen bug when opening another SLD

EstherDarkish:

### DIFF
--- a/src/redux/reducer.js
+++ b/src/redux/reducer.js
@@ -587,6 +587,11 @@ export const reducer = createReducer(initialState, {
             });
         }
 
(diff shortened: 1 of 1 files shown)


laurgarn (approved):
Tested OK

etiennehomer (approved):

FranckLecuyer (changes requested):
Tests: we cannot return to a previously pinned sld from another sld in full screen mode

EstherDarkish:
Nice find. Here's a fix.

FranckLecuyer (approved):
Tests: OK

PR #1022, opened 2023-01-18
Switch click no full network reload

etiennehomer:
- fix reloadImpactedSubstationsEquipments() call - parameters were wrong
- refine the conditions for calling loadNetwork()
- remove useless loadNetworkRef.current();
- Remove redundant initalization of currentNodeRef.current
- Warning: as is, loadNetwork(isUpdate) is never called with isUpdate=false. It will be used again in the future when no collection of the network will be required to open the app.


src/components/study-container.js
```diff
@@ -644,8 +654,6 @@ export function StudyContainer({ view, onChangeTab }) {
             const ws = connectNotifications(studyUuid);
             const wsDirectory = connectDeletedStudyNotifications(studyUuid);
 
-            loadNetworkRef.current();
```
etiennehomer, line 647: This loadNetwork() is executed when the current node is not built. So it does nothing.

src/components/study-container.js
```diff
@@ -510,7 +508,11 @@ export function StudyContainer({ view, onChangeTab }) {
                 // updating data related to impacted substations
                 if (substationsIds?.length > 0) {
                     console.info('Reload network equipments');
-                    network.reloadImpactedSubstationsEquipments(substationsIds);
+                    network.reloadImpactedSubstationsEquipments(
+                        studyUuid,
+                        currentNode,
```
sBouzols, line 513: Use `currentNodeRef.current` to avoid to proc this effect when currentNode changes ?
`currentNodeRef.current` is updated in another Effect, maybe we should order our effects then.

etiennehomer: Yes it limits the number of times this effect is executed , I changed it.
But I don't think we need to change the order

Meklo (approved):

src/components/study-container.js
```diff
@@ -625,8 +626,14 @@ export function StudyContainer({ view, onChangeTab }) {
         // if only node renaming, do not reload network
         if (isNodeRenamed(previousCurrentNode, currentNode)) return;
         if (!isNodeBuilt(currentNode)) return;
-        loadNetwork(true);
-    }, [loadNetwork, currentNode, wsConnected]);
+        if (
+            !network ||
```
sBouzols, line 630: If first time and network is undefined then why not calling `loadNetwork(false)`
There is no case where we call this ( `loadNetwork(false)` or `loadNetwork()`), maybe there is something we are missing...

src/components/study-container.js
sBouzols: Here we want to force the network reload when we get the websocket connection back after a network problem maybe ?

src/components/study-container.js
etiennehomer: I think we moved away from the parameter name and comment of loadNetwork function.
Currently, we do the network initialization with a loadNetwork(isUpdate=true).
A clean up can be done in this PR, we'll talk about it

src/components/study-container.js
jonenst: No this code is not executed in this case. It maybe just be cruft that accumulated over time and can ben safely removed

sBouzols (approved):
Code Reviw OK
Tests OK
Console warning check OK

src/components/study-container.js
```diff
@@ -502,7 +501,6 @@ export function StudyContainer({ view, onChangeTab }) {
                         (node) => node.id === firstSelectedNode.id
                     ),
                 };
-                currentNodeRef.current = ModelFirstSelectedNode;
```
etiennehomer, line 505: @jonenst @sBouzols @Meklo ok with that ?
currentNodeRef.current is initialized at row 634.
Thus, at study opening, I will have previousCurrentNode=undefined. And the first and only execution of the network load useEffect will work fine.

jonenst (approved):

PR #1023, opened 2023-01-18
[WIP] Fix overloaded lines table and overloaded lines on map in loadflow active mode

FranckLecuyer:


src/components/network/line-layer.js
```diff
@@ -64,34 +65,57 @@ export const LineLoadingZone = {
 };
 
 export function getLineLoadingZoneOfSide(
+    line,
```
bendaamerahm, line 68: 'line' is declared but its value is never read

src/components/network/line-layer.js
```diff
@@ -64,34 +65,57 @@ export const LineLoadingZone = {
 };
 
 export function getLineLoadingZoneOfSide(
```
bendaamerahm, line 67: i suggest to optimise the code of this function to be like this: 
```
if (limit === undefined || (intensity === undefined && activeFlow === undefined) || (intensity === 0 && activeFlow === 0)) {
    return LineLoadingZone.UNKNOWN;
} 
const valIntensity = intensity !== undefined ? intensity : activeFlow * 1000 / (voltagLevel.nominalVoltage * Math.sqrt(3) * DEFAULT_COSPHI);
const threshold = (lineFlowAlertThreshold * limit) / 100;
return valIntensity > 0 && valIntensity < threshold ? LineLoadingZone.SAFE : valIntensity >= threshold && valIntensity < limit ? LineLoadingZone.WARNING : LineLoadingZone.OVERLOAD;
```

- you can remove the first if statement and combine the conditions to check for undefined or zero values into a single ternary operator.

- instead of checking for intensity !== undefined, you can use the ternary operator to assign a default value of (activeFlow * 1000) / (voltagLevel.nominalVoltage * Math.sqrt(3) * DEFAULT_COSPHI) when intensity is undefined.

- you can use a single return statement with a ternary operator to determine the line loading zone based on the value of valIntensity.

src/components/network/line-layer.js
```diff
@@ -64,34 +65,57 @@ export const LineLoadingZone = {
...
+        } else if (valIntensity >= threshold && valIntensity < limit) {
             return LineLoadingZone.WARNING;
         } else {
             return LineLoadingZone.OVERLOAD;
         }
     }
 }
 
-export function getLineLoadingZone(line, lineFlowAlertThreshold) {
+export function getLineLoadingZone(
```
bendaamerahm, line 98: i have a suggestion here , we can eliminate the second call to getLineLoadingZoneOfSide and uses a ternary operator to check the value of zone1 and zone2 and return the appropriate zone.
something like that:

```
export function getLineLoadingZone(
    line,
    lineFlowAlertThreshold,
    voltageLevel1,
    voltageLevel2
) {
    const zone1 = getLineLoadingZoneOfSide(
        line,
        line.permanentLimit1,
        line.p1,
        line.i1,
        lineFlowAlertThreshold,
        voltageLevel1
    );
    return zone1 >=
        getLineLoadingZoneOfSide(
            line,
            line.permanentLimit2,
            line.p2,
            line.i2,
            lineFlowAlertThreshold,
            voltageLevel2
        )
        ? zone1
        : LineLoadingZone.OVERLOAD;
}
```

bendaamerahm (commented):
i have some suggestions but not mandatories

geofjamg (changes requested):
The limit calculation has to be done on server side

dbraquart:
no computation in Front-end, we can cancel this PR

PR #1031, opened 2023-01-19
fetch map substations and lines in parallel

AbdelHedhili:


src/utils/rest-api.js
```diff
@@ -2407,9 +2408,17 @@ export function fetchMapEquipments(
...
 
-    let fetchEquipmentsUrl =
-        getStudyUrlWithNodeUuid(studyUuid, currentNodeUuid) +
-        '/network-map/map-equipments';
+    let fetchEquipmentsUrl;
+    if (equipmentType === 'substations') {
+        fetchEquipmentsUrl =
+            getStudyUrlWithNodeUuid(studyUuid, currentNodeUuid) +
+            '/network-map/map-substations';
+    } else {
```
flomillot: It's not safer to check if the equipmentType is 'line' otherwise do a console.error ?

flomillot (approved):

src/utils/rest-api.js
AbdelHedhili: I split it into 2 functions following @jonenst advice

flomillot: Ok

flomillot:
Tested with my PR #1032 and it seems to work fine

src/utils/rest-api.js
flomillot: Even if you split in 2 functions, I think both of these function should call the same function inside rest-api.js in fine. Because here all the code is duplicate and only one parameter change.

laurgarn (approved):

PR #1032, opened 2023-01-19
Parallelize map equipment and geo data fetch at initialization

flomillot:


src/components/network-map-tab.js
```diff
@@ -712,7 +710,7 @@ export const NetworkMapTab = ({
             mapEquipments={mapEquipments}
             updatedLines={updatedLines}
             geoData={geoData}
-            waitingLoadGeoData={waitingLoadGeoData}
+            displayOverlayLoader={!basicDataReady}
```
jonenst, line 713: need to add && waiting, otherwise if there is an error the loader is displayed indefinitely

src/components/network-map-tab.js
```diff
@@ -532,62 +525,65 @@ export const NetworkMapTab = ({
...
     useEffect(() => {
-        if (studyUpdatedForce.eventData.headers) {
+        if (isInitialized && studyUpdatedForce.eventData.headers) {
             if (
                 studyUpdatedForce.eventData.headers[UPDATE_TYPE_HEADER] ===
                 'loadflow'
             ) {
                 //TODO reload data more intelligently
-                loadMapEquipments();
+                updateMapEquipmentsAndGeoData();
```
etiennehomer, line 583: We need to update geo data after a loadflow ?

flomillot: I just kept the old behaviour (after updating map equipment we always called 'loadGeoData'), so do I change this ?

src/components/network-map-tab.js
```diff
@@ -497,13 +491,12 @@ export const NetworkMapTab = ({
 
         Promise.all([substationPositionsDone, linePositionsDone])
             .then(() => {
-                setWaitingFullLoadGeoData(false);
+                setWaitingLoadData(false);
```
sBouzols, line 496: To avoid duplication of this line, we could move it in a `Promise.finally` ?

src/components/network-map-tab.js
```diff
@@ -435,11 +431,11 @@ export const NetworkMapTab = ({
                         );
                         setGeoData(newGeoData);
                     }
-                    setWaitingLoadTemporaryGeoData(false);
+                    setWaitingLoadData(false);
```
sBouzols, line 436: To avoid duplication of this line, we could move it in a `Promise.finally` ?

src/components/network-map-tab.js
etiennehomer: Yes it's useless to avoid calculating that no geodata is missing here and only load map equipments

src/components/network-map-tab.js
flomillot: Yes totally, I do it

src/components/network-map-tab.js
flomillot: I refactor a bit

src/components/network-map-tab.js
flomillot: Done

src/components/network-map-tab.js
```diff
@@ -532,62 +525,68 @@ export const NetworkMapTab = ({
...
+            intlRef
+        );
+        dispatch(resetMapReloaded());
+    }, [currentNode, dispatch, intlRef, setErrorMessage, studyUuid]);
+
+    const updateMapEquipments = useCallback(() => {
+        if (!isNodeBuilt(currentNode) || !studyUuid || !mapEquipments) {
+            return Promise.reject();
+        }
+        console.info('Reload map equipments');
```
etiennehomer, line 542: Change log for "Upload map equipents" ?

src/components/network-map-tab.js
```diff
@@ -103,17 +103,16 @@ export const NetworkMapTab = ({
...
-    const [waitingFullLoadGeoData, setWaitingFullLoadGeoData] = useState(true);
-    const [waitingLoadTemporaryGeoData, setWaitingLoadTemporaryGeoData] =
-        useState(false);
+    const [waitingLoadData, setWaitingLoadData] = useState(true);
 
     const [geoData, setGeoData] = useState();
     const geoDataRef = useRef();
     geoDataRef.current = geoData;
 
+    const basicDataReady = mapEquipments && geoData;
```
etiennehomer, line 112: initialDataReady maybe

src/components/network-map-tab.js
```diff
@@ -712,7 +723,7 @@ export const NetworkMapTab = ({
             mapEquipments={mapEquipments}
             updatedLines={updatedLines}
             geoData={geoData}
-            waitingLoadGeoData={waitingLoadGeoData}
+            displayOverlayLoader={!basicDataReady && waitingLoadData}
```
etiennehomer, line 726: displayOverlayLoader={!basicDataReady}
it's enough nope ?

src/components/network-map-tab.js
```diff
@@ -532,62 +525,68 @@ export const NetworkMapTab = ({
...
     useEffect(() => {
-        if (studyUpdatedForce.eventData.headers) {
+        if (isInitialized && studyUpdatedForce.eventData.headers) {
             if (
                 studyUpdatedForce.eventData.headers[UPDATE_TYPE_HEADER] ===
                 'loadflow'
             ) {
-                //TODO reload data more intelligently
-                loadMapEquipments();
+                updateMapEquipments();
```
etiennehomer, line 586: Do we want to do loadMapEquipments() or updateMapEquipments() here ? It's  a bit tricky
What's the difference between new MapEquipments() and mapEquipments.reloadImpactedSubstationsEquipments() with empty updatedSubstationsToSend ?

src/components/network-map-tab.js
flomillot: I changed it after the comment of @jonenst here https://github.com/gridsuite/gridstudy-app/pull/1032#discussion_r1081579427
But actually if there an error all the screen get black with just the error message so yeah in practice it's enough I think.

flomillot: But maybe one day only the map screen will display the error ? I don't know.

src/components/network-map-tab.js
sBouzols: you mean "Update map equipments" ?

flomillot: I changed into update indeed

src/components/network-map-tab.js
```diff
@@ -615,41 +614,53 @@ export const NetworkMapTab = ({
...
     useEffect(() => {
-        if (!mapEquipments) return;
-        loadGeoData();
-    }, [loadGeoData, mapEquipments]);
+        const prevLineFullPath = lineFullPathRef.current;
+        lineFullPathRef.current = lineFullPath;
+        if (isInitialized && lineFullPath && !prevLineFullPath) {
+            loadGeoData();
+        }
+    }, [isInitialized, lineFullPath, loadGeoData]);
```
sBouzols, line 644: `isInitialized` change will proc this effect. Do we really need it ?
It is set to true just after a load or an update, then there is a case where `loadGeoData` will be called twice ?

src/components/network-map-tab.js
```diff
@@ -532,62 +525,68 @@ export const NetworkMapTab = ({
...
                 studyUpdatedForce.eventData.headers[UPDATE_TYPE_HEADER] ===
                 'loadflow'
             ) {
-                //TODO reload data more intelligently
-                loadMapEquipments();
+                updateMapEquipments();
             }
         }
-    }, [studyUpdatedForce, loadMapEquipments]);
+    }, [isInitialized, studyUpdatedForce, updateMapEquipments]);
```
sBouzols, line 589: `isInitialized` and `updateMapEquipments` change will proc this effect. Do we really need it ?
It' a sepcial effect of notification consumption here. We want to avoid process multiple time an identic notification.

src/components/network-map-tab.js
etiennehomer: Yes Update... And equipments* :)

src/components/network-map-tab.js
flomillot: If we don't check with isIinitialized we fetch geo data 2 times at initialization if lineFullPath is true.
We could fix it by setting useRef(true) but using this boolean seems to be safer and clearer to me. Should I change it ?

src/components/network-map-tab.js
flomillot: I added isInitialized by security but I can remove it. But updateMapEquipments I can't remove it of the deps, no ?

src/components/network-map-tab.js
etiennehomer: Actually, updateMapEquipments is fine

src/components/network-map-tab.js
etiennehomer: Ok to keep it

src/components/network-map-tab.js
```diff
@@ -615,41 +614,53 @@ export const NetworkMapTab = ({
...
-                         2) we keep the manual mode => we need to fix the "bug" and do only one reload
      */
     useEffect(() => {
-        let previousNodeStatus = isCurrentNodeBuiltRef.current;
-        isCurrentNodeBuiltRef.current = isNodeBuilt(currentNode);
-
         // when we build node we want the map to be up to date
-        if (!previousNodeStatus && isCurrentNodeBuiltRef.current) {
-            loadMapEquipments();
+        if (
```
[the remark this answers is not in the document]
flomillot: I changed it back

etiennehomer (approved):

sBouzols (approved):
Code review OK
Tests OK
Console Warning check OK

src/components/network-map-tab.js
jonenst: it's the wrong way to do it, but it was done like this in the previous code so I suggest to keep it for now

src/components/network-map-tab.js
jonenst: In this case, this is handled by explicitly checking for lineFullPath && !prevLineFullPath. The effect is triggered but it does nothing in other cases

src/components/network-map-tab.js
jonenst: yes it's good practice to never display a spinning loader "by default" (!data). Only display a loader when we know that there is an actual http request that we are waiting for (waiting && !data). Otherwise after the code is changed a few times and there are errors, the users may see infinite spinners because we're not actually waiting for anything

jonenst (approved):

Meklo (approved):

PR #1036, opened 2023-01-19
Fix equipment deletion bugs

ayolab:


src/components/network-map-tab.js
```diff
@@ -597,13 +598,15 @@ export const NetworkMapTab = ({
         if (!mapEquipments || refIsMapManualRefreshEnabled.current) {
             return;
         }
-        if (deletedEquipment) {
-            mapEquipments?.removeEquipment(
-                deletedEquipment?.type,
-                deletedEquipment?.id
-            );
+        if (deletedEquipments?.length > 0) {
```
laurgarn, line 601: May be test non-undefinition of mapEquipments out of the loop ?

src/components/network/network.js
```diff
@@ -607,15 +607,18 @@ export default class Network {
...
-
-                this.substationsById.get(substationId).voltageLevels =
-                    voltageLevelsOfSubstation;
+                    this.voltageLevelsById.get(equipmentId)?.substationId;
+                if (this.substationsById.get(equipmentId)) {
+                    let voltageLevelsOfSubstation =
+                        this.substationsById.get(substationId).voltageLevels;
+                    voltageLevelsOfSubstation =
+                        voltageLevelsOfSubstation.filter(
+                            (l) => l.id !== equipmentId
```
laurgarn, line 616: I would prefer `vl`, instead of `l`, as identifier for a voltage level.

src/components/network/network.js
```diff
@@ -607,15 +607,18 @@ export default class Network {
...
-                    this.voltageLevelsById.get(equipmentId).substationId;
-                let voltageLevelsOfSubstation =
-                    this.substationsById.get(substationId).voltageLevels;
-                voltageLevelsOfSubstation = voltageLevelsOfSubstation.filter(
-                    (l) => l.id !== equipmentId
-                );
-
-                this.substationsById.get(substationId).voltageLevels =
-                    voltageLevelsOfSubstation;
+                    this.voltageLevelsById.get(equipmentId)?.substationId;
```
laurgarn, line 610: It seems substationId is only needed inside the if.
I would prefer defining `const voltageLevel = this.voltageLevelsById.get(equipmentId)`
and testing its voidness as the condition of the the block.
The same with the substation : `const substation = this.substationsById.get(substationId)`

laurgarn:
Tested OK

src/components/network/network.js
```diff
@@ -607,15 +607,17 @@ export default class Network {
...
+                const substation = this.substationsById.get(substationId);
+                if (substation) {
```
etiennehomer, line 612: It could be:
if (substation) {
                    substation.voltageLevels = substation.voltageLevels.filter(
                        (vl) => vl.id !== equipmentId
                    );
                }

src/components/study-container.js
```diff
@@ -561,31 +551,33 @@ export function StudyContainer({ view, onChangeTab }) {
...
+                                deletedEquipment?.type,
+                                ' from the network'
+                            );
+                            network.removeEquipment(
+                                deletedEquipment?.type,
+                                deletedEquipment?.id
+                            );
+                        }
+                    });
+                    dispatch(setDeletedEquipments(deletedEquipments));
```
etiennehomer, line 573: Do we use the redux to dispatch the deleted equipments or we catch notifications the same way as here in the nework-map-tab ?

etiennehomer: To be done later

etiennehomer (approved):

PR #1041, opened 2023-01-23
Generator creation form regression

EstherDarkish:
Fixes a stuck validation form in the generator creation dialog if the reactive power or the qpercent is incorrectly set and then not used.


laurgarn (approved):
Tests OK

ayolab (approved):

PR #1048, opened 2023-01-24
Fix generator modification minMaxReactiveLimits validation

ayolab:


FranckLecuyer (approved):
Review: OK
Tests: OK

FranckLecuyer:
When we change a generator from 'By diagram' to 'By range' without setting the minimum and maximum reactive power, log shows something like :

Reactive limits By range modified :
    Minimum reactive power : NaN -> -1.7976931348623157E308
    Maximum reactive power : NaN -> 1.7976931348623157E308

ayolab:
> When we change a generator from 'By diagram' to 'By range' without setting the minimum and maximum reactive power, log shows something like :
> 
> Reactive limits By range modified : Minimum reactive power : NaN -> -1.7976931348623157E308 Maximum reactive power : NaN -> 1.7976931348623157E308

since the fields are optional, if the user doesn't set their values we set them to min and max value of Double and we add a functional log for that so that user can see what happened after the modification is applied
I already talked with Christelle about this

FranckLecuyer:
> > When we change a generator from 'By diagram' to 'By range' without setting the minimum and maximum reactive power, log shows something like :
> > Reactive limits By range modified : Minimum reactive power : NaN -> -1.7976931348623157E308 Maximum reactive power : NaN -> 1.7976931348623157E308
> 
> since the fields are optional, if the user doesn't set their values we set them to min and max value of Double and we add a functional log for that so that user can see what happened after the modification is applied I already talked with Christelle about this

ok

src/components/dialogs/generator-modification-dialog.js
```diff
@@ -294,6 +301,17 @@ const GeneratorModificationDialog = ({
...
     });
 
+    useEffect(() => {
+        setReactivePowerRequired(
+            (minimumReactivePower !== '' &&
+                !generatorInfos?.minMaxReactiveLimits?.minimumReactivePower) ||
+                (maximumReactivePower !== '' &&
+                    !generatorInfos?.minMaxReactiveLimits?.maximumReactivePower)
+                ? true
+                : undefined // if the field is not required then we set "reactivePowerRequired" to undefined so that the optional label is not displayed
```
etiennehomer, line 311: That's tricky. We used this in other places ?
But maybe it's not a big deal as we'll refacto forms.

EstherDarkish (approved):

etiennehomer (approved):

PR #1063, opened 2023-01-30
[WIP] Fix geo data retrieval error

niconoir:


src/components/network-map-tab.js
```diff
@@ -664,7 +688,7 @@ export const NetworkMapTab = ({
             !previousNodeStatus &&
             isCurrentNodeBuiltRef.current
         ) {
-            updateMapEquipmentsAndGeoData();
+            updateMapEquipmentsAndGeoData(currentNode);
```
etiennehomer, line 703: We don't need the Object.Assign() here ?

PR #1065, opened 2023-01-30
Fixes the height of the diagram and SVG

EstherDarkish:
When resizing a diagram and moving its SVG, the SVG can sometime disapear if the new diagram's height is bigger than every other diagram's maximum height.


flomillot (approved):
Tested

etiennehomer:
Do we want to zoom when we increase the height ? Or we just want to make the pane bigger and not zoom in ?

EstherDarkish:
The zoom issue is to be fixed in another US

jonenst (approved):

PR #1066, opened 2023-01-31
A NAD could be displayed on the left after window resizing

EstherDarkish:
Fixes a bug which causes the right displayed diagrams to be displayed on the left of the pane after a window resizing.


jonenst (approved):

flomillot (approved):
Tested

PR #1069, opened 2023-02-01
Fix GeoData errors

ayolab:

### DIFF
--- a/src/components/network-map-tab.js
+++ b/src/components/network-map-tab.js
@@ -561,13 +561,28 @@ export const NetworkMapTab = ({
 
         dispatch(resetMapReloaded());
 
+        const previousCurrentNode = currentNode;
+
         return mapEquipments
             .reloadImpactedSubstationsEquipments(
                 studyUuid,
                 currentNode,
-                updatedSubstationsToSend,
-                setUpdatedLines
+                updatedSubstationsToSend
             )
+            .then((results) => {
+                const isFullReload = updatedSubstationsToSend ? false : true;
+                if (previousCurrentNode === currentNodeRef.current) {
+                    mapEquipments.updateSubstations(
+                        mapEquipments.checkAndGetValues(results[0]),
+                        isFullReload
+                    );
+                    mapEquipments.updateLines(
+                        mapEquipments.checkAndGetValues(results[1]),
+                        isFullReload
+                    );
+                    setUpdatedLines(results[1]);
+                }
+            })
             .finally(() => {
                 setWaitingLoadData(false);
             });
@@ -581,8 +596,13 @@ export const NetworkMapTab = ({
     ]);
 
     const updateMapEquipmentsAndGeoData = useCallback(() => {
-        updateMapEquipments().then(loadGeoData);
-    }, [updateMapEquipments, loadGeoData]);
+        const previousCurrentNode = currentNode;
+        updateMapEquipments().then(() => {
+            if (previousCurrentNode === currentNodeRef.current) {
+                loadGeoData();
+            }
+        });
+    }, [currentNode, updateMapEquipments, loadGeoData]);
 
     useEffect(() => {
         if (isInitialized && studyUpdatedForce.eventData.headers) {
--- a/src/components/network/map-equipments.js
+++ b/src/components/network/map-equipments.js
@@ -81,8 +81,7 @@ export default class MapEquipments {
     reloadImpactedSubstationsEquipments(
         studyUuid,
         currentNode,
-        substationsIds,
-        handleUpdatedLines
+        substationsIds
     ) {
         const updatedSubstations = fetchMapSubstations(
             studyUuid,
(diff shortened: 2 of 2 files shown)


src/components/study-container.js
```diff
@@ -594,17 +594,28 @@ export function StudyContainer({ view, onChangeTab }) {
...
+                                dispatch(
+                                    setUpdatedSubstationsIds(substationsIds)
+                                );
+                            }
+                        });
                 }
             }
         }
-    }, [studyUpdatedForce, network, studyUuid, dispatch]);
+    }, [studyUpdatedForce, network, studyUuid, dispatch, currentNode]);
```
sBouzols, line 618: to avoid adding a new useEffect dependency you could use `currentNodeRef.current` instead of `currentNode`
in
```
const previousCurrentNode = currentNodeRef.current;
```
then it is also miss-named, maybe something like `currentNodeAtRelaodCalling` is a better name

src/components/network-map-tab.js
```diff
@@ -581,8 +594,13 @@ export const NetworkMapTab = ({
     ]);
 
     const updateMapEquipmentsAndGeoData = useCallback(() => {
-        updateMapEquipments().then(loadGeoData);
-    }, [updateMapEquipments, loadGeoData]);
+        const previousCurrentNode = currentNode;
```
sBouzols, line 597: to avoid adding a new usecallback dependency (which is a dependency in updateMapEquipments anyway) you could use `currentNodeRef.current` instead of `currentNode`
in
```
const previousCurrentNode = currentNodeRef.current;
```
then it is also miss-named, maybe something like `currentNodeAtRelaodCalling` is a better name
To mark that we check at the end of an async call (fetch) that the context (node) is still the same.

src/components/network-map-tab.js
```diff
@@ -561,15 +561,28 @@ export const NetworkMapTab = ({
 
         dispatch(resetMapReloaded());
 
+        const previousCurrentNode = currentNode;
```
sBouzols, line 564: same comment here

src/components/network-map-tab.js
```diff
@@ -561,15 +561,28 @@ export const NetworkMapTab = ({
...
                 studyUuid,
                 currentNode,
-                updatedSubstationsToSend,
-                setUpdatedLines
+                updatedSubstationsToSend
             )
-            .finally(() => {
-                setWaitingLoadData(false);
+            .then((results) => {
+                const isFullReload = updatedSubstationsToSend ? false : true;
```
etiennehomer, line 573: You can move this inside the if

src/components/study-container.js
ayolab: Done

src/components/network-map-tab.js
ayolab: Done

src/components/network-map-tab.js
ayolab: Done

etiennehomer: you could use : 
mapEquipments.updateSubstations(
                        mapEquipments.checkAndGetValues(results[0]),
                        !!updatedSubstationsToSend
                    );

src/components/network-map-tab.js
```diff
@@ -561,15 +561,30 @@ export const NetworkMapTab = ({
...
             .reloadImpactedSubstationsEquipments(
                 studyUuid,
                 currentNode,
-                updatedSubstationsToSend,
-                setUpdatedLines
+                updatedSubstationsToSend
             )
-            .finally(() => {
-                setWaitingLoadData(false);
+            .then((results) => {
```
jonenst, line 572: this wait for both to be done, we want to update things as soon as the request finished

src/components/network-map-tab.js
```diff
@@ -561,15 +561,30 @@ export const NetworkMapTab = ({
...
+                    mapEquipments.updateSubstations(
+                        mapEquipments.checkAndGetValues(results[0]),
+                        isFullReload
+                    );
+                    mapEquipments.updateLines(
+                        mapEquipments.checkAndGetValues(results[1]),
+                        isFullReload
+                    );
+                    setUpdatedLines(results[1]);
+                    setWaitingLoadData(false);
```
niconoir, line 586: Why did you remove the `.finally(() => {setWaitingLoadData(false);}` and put                     `setWaitingLoadData(false);` in the `.then` ?

src/components/network/map-equipments.js
```diff
@@ -95,13 +96,14 @@ export default class MapEquipments {
...
         const isFullReload = substationsIds ? false : true;
-
         updatedSubstations
             .then((values) => {
-                this.updateSubstations(
-                    this.checkAndGetValues(values),
-                    isFullReload
-                );
+                if (currentNodeAtReloadCalling?.id === currentNode?.id) {
+                    this.updateSubstations(
```
etiennehomer, line 102: Maybe we could avoid this update. It looks like currentNodeAtReloadCalling is updated too soon so the test returns true and the treatment is not cancelled.

etiennehomer (approved):

etiennehomer (approved):

niconoir (approved):

PR #1071, opened 2023-02-02
SLD stops changing width on un-built node

EstherDarkish:
Stops the diagrams from taking all the pane's width when switching from a built node to a not built one.


laurgarn (approved):
Works

etiennehomer (approved):

PR #1074, opened 2023-02-02
Synchronization plots for dynamic simulation results by using useRef and useImperativeHandle

thangqp:
Using useRef and useImperativeHandle instead of eventemiiter for communicating between sibling plots

TODO

- Something wrong when deactivate/activate button sync, but some plots sync, others KO.  WIP


thangqp:
Expired

PR #1079, opened 2023-02-06
Optimize item list in case user clicks so quickly

thangqp:
- use debounce technique with relay = 500 ms when user clicks quickly for choosing series items

**Other improvements:**
- add scroll bar to the graph zone
- work around grid layout to make plots responsible to a layout change event (using ResizeObserver)


thangqp:
image

jonenst (approved):

PR #1082, opened 2023-02-07
Fix generator modification maxreactivepower validation

ayolab:


FranckLecuyer (approved):

src/components/dialogs/generator-modification-dialog.js
```diff
@@ -271,11 +277,17 @@ const GeneratorModificationDialog = ({
         previousValues: generatorInfos?.reactiveCapabilityCurvePoints,
     });
 
+    const minimumReactivePowerRef = useRef();
```
EstherDarkish, line 280: Can you add a comment here saying something like 
`// We need this here because we can't access lexical declaration 'minimumReactivePower' before initialization`
?

EstherDarkish (approved):

src/components/dialogs/generator-modification-dialog.js
ayolab: Done

PR #1088, opened 2023-02-09
fix:show a lighter loader when loading sensitivity results

laurgarn:


FranckLecuyer (approved):
Code review: OK
Tests: OK

etiennehomer (changes requested):
When there's no sensi result, the timer makes the results to be refetched all the time

laurgarn:
> When there's no sensi result, the timer makes the results to be refetched all the time

Tang brings the solution in its PR below :
https://github.com/gridsuite/gridstudy-app/pull/1089

src/components/sensitivity-analysis-result-tab.js
```diff
@@ -165,7 +183,13 @@ function PagedSensitivityResult({
         next.isFetchNeedy = prev.isFetchNeedy;
         next.askedFilterVersion = prev.askedFilterVersion;
     } else if (!prev.fetched) {
-        // probably fetch failed, worth retrying
+        // probably fetch failed, worth retrying ?
+        if (retryCount >= RETRY_LIMIT_COUNT) {
```
etiennehomer, line 187: You don't need this retryCount. 
When there's no result, the behavior should be the same as when there's some results and should not refetch the result.
Removing setAllowsShowingLoader(false) can help

etiennehomer (changes requested):

src/components/sensitivity-analysis-result-tab.js
laurgarn: It will not re-fetch if the HTTP 204 is well handled beneath, as it s now Tang has corrected the fetch in rest-api.js.
The case it would try some more times is when a 4xx or 5xx is received.
The setAllowsShowingLoader(false) at row 264 is the point of the PR : prevent the loader to show too early.

etiennehomer: Yes but it's prevented as allowShowingLoader is initiated to false

laurgarn: The first time the table show : yes. Next times, when filtering/sorting/paging parameters change, it has to be set back to false in order to not have the loader appears too early.

src/components/sensitivity-analysis-result-tab.js
```diff
@@ -277,15 +308,18 @@ function PagedSensitivityResult({
...
     next.isLoading = isLoading;
 
     synthRef.current = next;
 
+    const showsLoading = isLoading && allowsShowingLoader;
+
     return (
         <>
-            <WaitingLoader message={'LoadingRemoteData'} loading={isLoading}>
+            <WaitingLoader message={'LoadingRemoteData'} loading={showsLoading}>
```
etiennehomer, line 322: You should rather use LoaderWithOverlay. WaitingLoader is on full page.

etiennehomer (approved):

PR #1089, opened 2023-02-09
Dynamic simulation failed run: fetch json must take into account no content http return code 204 

thangqp:
Http code no content 204 is without body.. reponse.json() can not read the null content.. must check this case to return directly null value


laurgarn (approved):
Code:OK
Test: correction applied in another context (sensi empty result) brings the solution.

PR #1104, opened 2023-02-16
Dynamic simulation: Support fit screen mode for individual plot and all plots together

thangqp:


thangqp:
fit_screen_mode_2

jonenst (approved):

thangqp:
Jon's comments: 

- a separated PR for the bug lack of handleLayoutChange
 
- h = 10 for fullscreen but computeRowHeight = height/12 does overflow in case of small page

- well formula is (h-constant)/10 for fullscreen or (h-rows*constant)/10 ?

- margin value is 10 by default, actually h=5 for normal mode and h=10 for fullscreen thus (h-1*10): 40 et 90

PR #1118, opened 2023-02-21
Useless fetch on hidden components

laurgarn:


src/components/study-pane.js
```diff
@@ -367,6 +367,7 @@ const StudyPane = ({
                     loadFlowInfos={loadFlowInfos}
                     network={network}
                     openVoltageLevelDiagram={openVoltageLevelDiagram}
+                    notTheContent={props.view !== StudyView.RESULTS}
```
etiennehomer, line 370: Rather name the props isActive or isOpen with opposite boolean value ?

etiennehomer, line 370: Rename the prop to visible as it is done for other components

etiennehomer (changes requested):

etiennehomer:
Code ok and tests ok. Just the rename to be done

src/components/result-view-tab.js
```diff
@@ -171,11 +174,11 @@ export const ResultViewTab = ({
                 </Tabs>
                 {disabled && <AlertInvalidNode />}
             </div>
-            {tabIndex === 0 && !disabled && renderLoadFlowResult()}
-            {tabIndex === 1 && !disabled && renderSecurityAnalysisResult()}
-            {tabIndex === 2 && !disabled && renderShortCircuitAnalysisResult()}
-            {tabIndex === 3 && !disabled && renderSensitivityAnalysisResult()}
-            {tabIndex === 4 && !disabled && renderDynamicSimulationResult()}
+            {tabIndex === 0 && showsSubs && renderLoadFlowResult()}
```
etiennehomer, line 177: The showTabs boolean could be used at the Paper level and we would have about the same behavior.

A smarter behavior would be : 1) not destroying the results tab component when it is not visible, so that we can have a very quick display without download when we click back on the results tab. 
2) but not downloading the results when the results tab is not visible and we change the current node for example.

src/components/result-view-tab.js
```diff
@@ -59,69 +94,61 @@ export const ResultViewTab = ({
...
+    const makeTabPaneJsx = (CompFunc, needsDeveloperMode, selected) => {
         return (
-            <Paper className={classes.analysisResult}>
-                <DynamicSimulationResultTab
-                    studyUuid={studyUuid}
-                    nodeUuid={currentNode?.id}
-                />
-            </Paper>
+            (!needsDeveloperMode || enableDeveloperMode) && (
+                <TabPanelLazy
```
jonenst, line 132: missing key

laurgarn: Done

laurgarn:
Superseded by #1194 (without the factorizations)

PR #1131, opened 2023-02-27
Dynamic Simulation - Config parameters startTime, stopTime, mapping, solver and provider 

thangqp:
Front-end: 

- Provide three sub tabs: 'Time Delay', 'Solver' and 'Mapping' in a new Dynamic Simulation Tab

Back-end:

- study-server persists solves in Json in a column of the new table DynamicSimulationParameters
- study-server sends a DTO {startTime, stopTime, mapping, solvers} to dynamic-simulation-server
- dynamic-simulation-server uses the received DTO (solver included) then exports to solver.par file by a xml exporter
- dynawaltz takes always files as inputs for the computation

parameters_solver

Limitation:

- When click 'Default Values', all parameters in different sub-tabs 'Time Delay', 'Solver' and 'Mapping' will be restore at the same time. This limitation should be improved later..


thangqp:
parameter_solver

src/components/dialogs/parameters/common/wrapper-input.js
```diff
@@ -0,0 +1,56 @@
...
+
+import { useStyles } from '../parameters';
+import * as yup from 'yup';
+import { FormProvider, useForm } from 'react-hook-form';
+import { yupResolver } from '@hookform/resolvers/yup/dist/yup';
+import React, { useEffect } from 'react';
+import { Box, Grid, Typography } from '@mui/material';
+import { FormattedMessage } from 'react-intl';
+
+const WrapperInput = ({ value, label, callback, validator, children }) => {
```
jonenst, line 16: why do we need a separate system based on RHF ? Can we have the code look like other forms ?

src/components/dialogs/parameters/common/wrapper-input.js
```diff
@@ -0,0 +1,56 @@
...
+            value: value,
+        },
+        resolver: yupResolver(schema),
+    });
+
+    const { handleSubmit, watch } = methods;
+
+    // to commit by onChange rather than onSubmit
+    useEffect(() => {
+        const subscription = watch(handleSubmit(callback));
```
jonenst, line 34: not sure why but this useeffect looks suspicious

src/translations/en.json
```diff
@@ -160,6 +159,27 @@
...
+    "DynamicSimulationIDASolverInitStep": "Initial step",
+    "DynamicSimulationIDASolverMinStep": "Minimum step",
+    "DynamicSimulationIDASolverMaxStep": "Maximum step",
+    "DynamicSimulationIDASolverAbsAccuracy": "Absolute accuracy",
+    "DynamicSimulationIDASolverRelAccuracy": "Relative accuracy",
+    "DynamicSimulationSimplifiedSolverHMin": "Minimum step",
+    "DynamicSimulationSimplifiedSolverHMax": "Maximum step",
+    "DynamicSimulationSimplifiedSolverKReduceStep": "Reduction factor of step",
+    "DynamicSimulationSimplifiedSolverNEff": "Maximum number of iterations with h too small",
+    "DynamicSimulationSimplifiedSolverNDeadband": " acceptable H minimum",
```
jonenst, line 179: typo whitespace " acceptable

jonenst (approved):

src/components/dialogs/parameters/common/wrapper-input.js
thangqp: To discuss later..

src/components/dialogs/parameters/common/wrapper-input.js
thangqp: Similar to other tabs, what do we expect is to commit a value input by user on the fly. When user begin typing, every change is captured then is passed to handleSubmit() which will invoke a callback later if input value is validated (by resolver) 

The solution is proposed following to the discussion at:

https://stackoverflow.com/questions/63466463/how-to-submit-react-form-fields-on-onchange-rather-than-on-submit-using-react-ho

src/translations/en.json
thangqp: Done!

PR #1136, opened 2023-03-01
ConnectionName is now saved as null instead of empty string

archendel:


flomillot (approved):

FranckLecuyer (approved):
Tests: OK
Code review: OK

PR #1139, opened 2023-03-02
migrate form delete a voltage level on a line (react hook form)

dbraquart:


src/components/refactor/dialogs/delete-voltage-level-on-line/delete-voltage-level-on-line-dialog.js
```diff
@@ -0,0 +1,133 @@
...
+import { FormProvider, useForm } from 'react-hook-form';
+import { deleteVoltageLevelOnLine } from '../../../../utils/rest-api';
+import { sanitizeString } from '../../../dialogs/dialogUtils';
+import yup from '../../utils/yup-config';
+import ModificationDialog from '../commons/modificationDialog';
+import DeleteVoltageLevelOnLineForm from './delete-voltage-level-on-line-form';
+
+const emptyFormData = {
+    [LINE_TO_ATTACH_TO_1_ID]: '',
+    [LINE_TO_ATTACH_TO_2_ID]: '',
```
archendel, line 26: use null default value when dealing with selectInput / Autocomplete
make those fields nullable in yup as well

currently, if you clear the field, you will get a wrong error message

archendel (changes requested):

src/components/refactor/dialogs/delete-voltage-level-on-line/delete-voltage-level-on-line-dialog.js
dbraquart: yes, I currently get "bad format" with an empty field.
with null / nullable, I got "this field must be set".

archendel (approved):

ayolab (approved):

PR #1143, opened 2023-03-03
LineAttachToVoltageLevel migration to react hook form and yup

ayolab:


src/components/refactor/dialogs/line-creation/line-creation-dialog.js
```diff
@@ -240,10 +240,10 @@ const LineCreationDialog = ({
                 line[CURRENT_LIMITS_2]?.[PERMANENT_LIMIT],
                 editData ? true : false,
                 editData ? editData.uuid : undefined,
-                sanitizeString(line[CONNECTIVITY_1]?.[CONNECTION_NAME]),
+                sanitizeString(line[CONNECTIVITY_1]?.[CONNECTION_NAME] ?? ''),
```
laurgarn, line 243: It bogs me we have to write this second protection.
Maybe we should have
```
function trimmedToNull(val) {
  if (!val) return null;
  const trimmed = val.trim();
  return trimmed || null;
}
```
or even `val?.trim() || null`

laurgarn (approved):
Test OK
Code OK with suggestions

archendel (approved):

PR #1148, opened 2023-03-06
Fix sensi paging change and loader

laurgarn:


ayolab (approved):
Tests OK

src/components/sensitivity-analysis-result-tab.js
```diff
@@ -19,6 +25,8 @@ import LinearProgress from '@mui/material/LinearProgress';
 
 const sensitivityAnalysisResultInvalidations = ['sensitivityAnalysisResult'];
 
+const loaderTimeoutMillis = 800;
```
etiennehomer, line 28: We keep 800ms or a bit less ? To validate with PO

etiennehomer (approved):

PR #1158, opened 2023-03-10
Fix some bugs ont the DndTable

flomillot:
Fix the bottom buttons positions of the DndTable when they are used independently.
Fix the form creation for steps to remove warnings and the validate button which were activated at opening.

The warning : MUI: A component is changing the uncontrolled checked state of SwitchBase to be controlled.
Elements should not switch from uncontrolled to controlled (or vice versa).
Decide between using a controlled or uncontrolled SwitchBase element for the lifetime of the component.
The nature of the state is determined during the first render. It's considered controlled if the value is not `undefined`.


dbraquart (changes requested):
Tests : all good
Review: maybe just a component to rename

dbraquart (approved):

PR #1164, opened 2023-03-14
Fix network update when switching between nodes

YenguiSeddik:


etiennehomer:
To validate with PO that we want a loader for those updates. Maybe it's ok without it

ayolab (approved):
code OK
tests OK

PR #1170, opened 2023-03-16
Form loading optimization

anistouri:


src/components/refactor/dialogs/load-modification/load-modification-dialog.js
```diff
@@ -69,7 +71,11 @@ const LoadModificationDialog = ({
         resolver: yupResolver(schema),
     });
 
-    const { reset } = methods;
+    const { reset, control } = methods;
+    const loadId = useWatch({
+        name: `${EQUIPMENT_ID}`,
```
archendel, line 76:
```suggestion
        name: EQUIPMENT_ID,
```

src/components/refactor/dialogs/load-modification/load-modification-dialog.js
```diff
@@ -50,8 +51,10 @@ const LoadModificationDialog = ({
     studyUuid,
     ...dialogProps
 }) => {
+    const [loadInfos, setLoadInfos] = useState(null);
     const currentNodeUuid = currentNode?.id;
     const { snackError } = useSnackMessage();
+    const isOpen = useOpenOnMount(editData, loadInfos, 200);
```
jonenst, line 57: use same name as creation
const open = useOpenOnMount(editData, generatorToModify, 200);

src/components/refactor/dialogs/commons/handle-modification-form.js
```diff
@@ -0,0 +1,16 @@
+import { useState, useEffect } from 'react';
+
+export const useOpenOnMount = (editData, infos, delay = 0) => {
```
jonenst, line 3: Simpler API ? Something like

```
export const useShortWaitFetching = (open, fetching, duration)

if (!open && !fetching) => return false
if (!open && fetching && "fetching vient de passer à true") => return false  (ie avant delay)
if (!open && fetching && "fetching est passé à true depuis longtemps") => return true (ie après delay)
if (open) => return true
```

PR #1179, opened 2023-03-23
don't invalidate copy on node rename/status update

AbdelHedhili:


src/components/network-modification-tree-pane.js
```diff
@@ -260,6 +260,25 @@ export const NetworkModificationTreePane = ({
...
+                updateNodes(studyUpdatedForce.eventData.headers['nodes']);
+                if (
+                    studyUpdatedForce.eventData.headers['nodes'].some(
+                        (nodeId) => nodeId === currentNodeRef.current?.id
+                    )
+                ) {
+                    dispatch(
+                        removeNotificationByNode([currentNodeRef.current?.id])
+                    );
+                }
```
Walid-Sahnoun, line 281: this method (useEffect) is becoming more and more complicated and contains a lot of if/else blocks (100 lines of code)
I think we should think about refactoring this method

SlimaneAmar: Yes
I have already noted this point in the  notification refactoring us

Walid-Sahnoun (approved):

SlimaneAmar (approved):

PR #1180, opened 2023-03-23
fix:prevent too early map redraw that hinder modification list to be shown early enough

laurgarn:


src/components/network/map-equipments.js
```diff
@@ -222,6 +226,7 @@ export default class MapEquipments {
     }
 
     updateLines(lines, fullReload) {
+        this.waitingForUpdate = false;
```
Meklo, line 229: Why does waitingForUpdate is set to false in updateLines and not in updateSubstations ?

laurgarn: Because lines are needed too and they are asked for later than substations (and are heavier to load), so I considered that enough data has come to start the redraw at that point.

anistouri:
Code OK
Test OK

anistouri (approved):

src/components/network/map-equipments.js
Meklo: Alright, can you add a comment to contextualize please ?

Meklo (approved):
Approved if that the brief unloading of the map when switching from one built node to another is not a deal breaker for the POs

PR #1181, opened 2023-03-24
Useless fetch on hidden components (useNodeData)

thangqp:
The logic of 'fetch or not' is implemented in useNodeData#isWorthUpdate().. so I push down the dormant state to this validator which decides to fetch or not


laurgarn:
I am not keen on bring the burden of dormancy deeply to every children downunder

PR #1193, opened 2023-03-29
Update the UI of the 2WT form

flomillot:
Update the UI of the 2WT form (in particular ratio and phase tap changer form).
We added a common header with name and ID.
The 'regulating' switch disapeared.
We added a regulation mode for ratio tap changer to replace it.
US1890


AbdelHedhili (changes requested):
why are the first 2 tabs (characteristics and limits) not increased in size also ? it's weird having the width changing on a tab click. Also it crushes the header since it doesn't have enough space.

AbdelHedhili (approved):
code: ok
test: ok
console checks: ok

PR #1195, opened 2023-03-29
Dynamic Simulation - load series on demand (no longer load on demand correctly)

thangqp:
By an incident, load on demand no longer works correctly, it performs the load overall series.


src/utils/rest-api.js
```diff
@@ -1254,7 +1254,7 @@ export function fetchDynamicSimulationResultTimeSeries(
         timeSeriesNames.length > 0 &&
         '?' + getQueryParamsList(timeSeriesNames, 'timeSeriesNames');
 
-    url += paramsList && '';
+    url += paramsList ?? '';
```
EstherDarkish, line 1257: paramsList cannot be undefined or null, so it is possible to remove entirely the "` ?? ''; `"

thangqp: In fact, the back end accepts also request without param. In this case back end will return all series.
This is an API so every cases must be covered.

EstherDarkish: Ok, my comment was incorrect, I thought these two evaluations would both return false : 

```
false && true && '?' + 'hello'
=> Returns false

undefined && true && '?' + 'hello'
=> Returns undefined
```

jonenst (approved):

PR #1205, opened 2023-04-04
Reopen sld at last position when it is minimized

YenguiSeddik:


EstherDarkish (changes requested):
Does not work. The state's order should be done when opening a diagram, not when minimizing it.

We can reproduce a non working scenario like this :

- Open A, pin A, open B, pin B, open C, pin C.
- Minimise B, minimise C.
- Open C, pin C, open B => B is between A and C, but should be on the right instead.

src/redux/reducer.js
```diff
@@ -683,8 +683,14 @@ export const reducer = createReducer(initialState, {
                             diagram.state = ViewState.MINIMIZED;
                         }
                     });
+                    const diagramToOpen = diagramStates[diagramToOpenIndex];
+
+                    // we push the diagram to the last position, so when we reopen it, it opens at the last position available.
```
EstherDarkish, line 688: The comment should be updated to :
`// We open and push the SLD to the last position in the array so it is displayed at the right of the others`

and the `diagramToOpen.state = ViewState.OPENED;` should be on the next line, before the splice, to make the code easier to read.

src/redux/reducer.js
```diff
@@ -683,8 +683,14 @@ export const reducer = createReducer(initialState, {
...
+                    diagramStates.splice(diagramToOpenIndex, 1);
+                    diagramStates.push(diagramToOpen);
+
                     // And update the one to open.
-                    diagramStates[diagramToOpenIndex].state = ViewState.OPENED;
+                    diagramToOpen.state = ViewState.OPENED;
```
EstherDarkish, line 693: This line should be moved (see comment above) and the comment should be removed.

EstherDarkish (approved):

PR #1206, opened 2023-04-04
Tweak AG Grid spreadsheet implementation

Meklo:


src/components/spreadsheet/equipment-table.js
```diff
@@ -95,8 +107,11 @@ export const EquipmentTable = ({
             wrapHeaderText: true,
             autoHeaderHeight: true,
             suppressKeyboardEvent: (params) => suppressKeyEvent(params),
+            icons: {
+                menu: filterIcon,
+            },
```
YenguiSeddik, line 112: i think we can directly set :
icons: {
                menu: '<span class="ag-icon ag-icon-filter"/>',
            },
and remove filterIcon const

YenguiSeddik (approved):

src/components/spreadsheet/equipment-table.js
Meklo: Yes it seems a bit overkill, i removed it

PR #1214, opened 2023-04-05
fix update status when we change node and back ko

anistouri:
When we change the node from a failed node, we need to update the status and display the new status accordingly.


laurgarn (approved):
OK

YenguiSeddik (approved):

PR #1220, opened 2023-04-12
Refactor SVG fetching

flomillot:
Before, the SVG was fetched and updated inside the final components (SingleLineDiagramContent or NetworkAreaDiagramContent).
It was managed inside DiagramPane by using a ref to these two components.
It has several advantages (no need to update manually the diagram views, the diagram view list was recreated from scratch everytime, and the update was still smooth thanks to React), but had few drawbacks : 
- we didn't have access to metadata of the SVG higher in the hierarchy where it was needed
- when a diagram was minimized / opened, we unmount and mount the diagram, so we lose all the data and we had to fetch again the SVG
- use of ref which is to avoid according to React documentation
- presence of few bugs, for exemple, fetch 6 times the same SVG when we change the current node

Now the data updating, fetching and storing is at the same level in DiagramPane.
The next step is to manage the fusion of the different NAD inside redux and we will be able to remove half of the code of this PR.


src/components/diagrams/diagram-pane.js
```diff
@@ -130,83 +138,134 @@ const useDisplayView = (network, studyUuid, currentNode) => {
...
-                let displayedVoltageLevels = [];
-                let nadTitle = '';
-                if (ids) {
-                    ids.forEach((id) =>
-                        displayedVoltageLevels.push(network.getVoltageLevel(id))
-                    );
+                if (ids?.length) {
+                    const svgUrl = checkAndGetNetworkAreaDiagramUrl(ids, depth);
+                    return fetchSvgData(svgUrl).then((svg) => {
+                        let nadTitle = '';
```
laurgarn, line 237: The nadTitle could/should be crafted in its own function

flomillot: I just kept the previous code

src/components/diagrams/diagram-pane.js
```diff
@@ -130,83 +138,134 @@ const useDisplayView = (network, studyUuid, currentNode) => {
...
+                        svg.additionalMetadata?.voltageLevels.forEach(
+                            (voltageLevel) => {
+                                const name = getNameOrId(voltageLevel);
+                                if (name !== null) {
+                                    nadTitle =
```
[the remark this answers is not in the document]
flomillot: As above I just kept the previous code

src/components/diagrams/diagram-pane.js
```diff
@@ -130,83 +138,134 @@ const useDisplayView = (network, studyUuid, currentNode) => {
         [studyUuid, currentNode]
     );
 
+    // this callback return a promise
+    const fetchSvgData = useCallback(
+        (svgUrl) => {
+            if (svgUrl) {
```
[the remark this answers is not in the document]
flomillot: I kept the previous code design

src/components/diagrams/diagram-pane.js
```diff
@@ -952,6 +961,13 @@ export function DiagramPane({
         ]
     );
 
+    const handleWarningToDisplay = (diagramView) => {
```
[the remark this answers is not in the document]
flomillot: I didn't move it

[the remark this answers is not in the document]
flomillot: Done

src/components/diagrams/diagram-pane.js
```diff
@@ -286,95 +362,314 @@ export function DiagramPane({
...
+            const networkAreaIds = [];
+            let networkAreaViewState = ViewState.OPENED;
+            diagramStates.forEach((diagramState) => {
+                if (diagramState.svgType === DiagramType.NETWORK_AREA_DIAGRAM) {
+                    networkAreaIds.push(diagramState.id);
+                    networkAreaViewState = diagramState.state; // They should all be the same value
+                }
+            });
+            if (networkAreaIds.length > 0) {
+                const sameNadAlreadyPresentInViews = viewsRef.current.find(
```
EstherDarkish, line 557: booleans should be names isXXX

src/components/diagrams/diagram-pane.js
```diff
@@ -286,95 +362,314 @@ export function DiagramPane({
...
+                    (diagramView) =>
+                        diagramView.svgType ===
+                            DiagramType.NETWORK_AREA_DIAGRAM &&
+                        diagramView.ids?.toString() ===
+                            networkAreaIds.toString() &&
+                        diagramView.depth === networkAreaDiagramDepth
+                );
+                if (!sameNadAlreadyPresentInViews) {
+                    addOrReplaceNAD(
```
EstherDarkish, line 566: The name of variable or functions do not make sense : if the NAD is not already in the views, why should we use a "xxxReplace" function? If the NAD **is** already present in the views, we **should** use a xxxReplace function, therefor why do we test `!sameNadAlreadyPresentInViews` ? Shouldn't we call the function below whether it's true or false ?

src/components/diagrams/diagram-pane.js
```diff
@@ -286,95 +362,314 @@ export function DiagramPane({
...
+    // Check if we need to remove or add SLDs
+    const updateSLDs = useCallback(
+        (diagramStates) => {
+            removeSLDs(diagramStates);
+            addSLDs(diagramStates);
+        },
+        [removeSLDs, addSLDs]
+    );
+
+    const addOrReplaceNAD = useCallback(
```
EstherDarkish, line 481: A simple comment is welcome here to help understand the goal of this function

src/components/diagrams/diagram-pane.js
```diff
@@ -286,95 +362,314 @@ export function DiagramPane({
...
+            ) {
+                // no more NAD, remove it from the views
+                removeNAD();
             }
+        },
+        [addOrReplaceNAD, removeNAD, networkAreaDiagramDepth]
+    );
+
+    // Update only the state of the diagrams (opened, minimized, etc)
+    const updateDiagramStates = useCallback((diagramStates) => {
```
EstherDarkish, line 586: Update this comment to say that this function's goal is to update 'views'

src/components/diagrams/diagram-pane.js
```diff
@@ -286,95 +362,314 @@ export function DiagramPane({
...
+                        });
                     });
+                });
+            }
+        },
+        [createView, intl]
+    );
+
+    // Check if we need to remove old SLDs from the 'views' and remove them if necessary
+    const removeSLDs = useCallback((diagramStates) => {
```
EstherDarkish, line 442: Maybe rename this function to "removeObsoleteSLDs" ?

src/components/diagrams/diagram-pane.js
```diff
@@ -286,95 +362,314 @@ export function DiagramPane({
...
-                        ...diagramState,
-                        name: diagramState?.id,
-                        // this variable is used to show a warning message inside the SLD
-                        warningMessage:
-                            DiagramType.VOLTAGE_LEVEL === diagramState.svgType
-                                ? 'VoltageLevelNotFound'
-                                : 'SubstationNotFound',
-                    };
+    // Check if we need to add new SLDs in the 'views' and add them if necessary
+    const addSLDs = useCallback(
```
EstherDarkish, line 379: Maybe rename this function to "addMissingSLDs" ?

src/components/diagrams/diagram-pane.js
```diff
@@ -286,95 +362,314 @@ export function DiagramPane({
...
+                ) {
+                    diagramsToUpdate.push({
+                        index: diagramIndex,
+                        state: diagramState.state,
+                    });
+                }
+            }
+        });
+
+        // If we found diagrams to update, we update them
```
EstherDarkish, line 633: // If we found diagrams to update, we update their states in "views"

src/components/diagrams/diagram-pane.js
```diff
@@ -409,27 +704,55 @@ export function DiagramPane({
...
-            viewsRef.current.forEach((diagramView) => {
-                // We search an instance of the current node's ID inside the diagram's URL to determine if the diagram should be updated
-                if (
-                    diagramView.svgUrl &&
-                    diagramView.svgUrl.indexOf(currentNodeRef.current?.id) !==
-                        -1
-                ) {
-                    diagramView.ref?.current?.reloadSvg();
+    // Updates particular diagrams or every diagram for the current node
+    const updateDiagrams = useCallback((ids) => {
```
EstherDarkish, line 708: If this function is rewriten, maybe that's a good time to change it so it is clearer that updateDiagrams(undefined) will update all diagrams, whereas updateDiagrams(something) will only update something. I would use two different functions to clarify this.

src/components/diagrams/diagram-pane.js
```diff
@@ -460,14 +783,26 @@ export function DiagramPane({
...
-                        const vl = network.getVoltageLevel(v.id);
-                        if (vl && substationsIds.includes(vl.substationId)) {
-                            updateDiagram(vl.id);
-                        }
-                    });
+                    let diagramIds = [];
+                    viewsRef.current
+                        .filter(
+                            (diagramView) =>
+                                diagramView.svgType !== DiagramType.SUBSTATION
```
EstherDarkish, line 790: I don't get why we only update SUBSTATIONs now and before we also updated VOLTAGE_LEVELs ?

src/components/diagrams/diagram-pane.js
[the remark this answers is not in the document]
EstherDarkish: Fixed in commit e30f64dd03d

src/components/diagrams/diagram-pane.js
```diff
@@ -286,95 +362,314 @@ export function DiagramPane({
...
+    viewsRef.current = views;
     const classes = useStyles();
-    const [warnings, setWarnings] = useState(new Map());
 
     /**
      * BUILDS THE DIAGRAMS LIST
      */
 
-    // Here, the goal is to build a list of view, each view corresponding to a diagram.
+    // Here (for the next 300 lines), the goal is to build a list of view, each view corresponding to a diagram.
```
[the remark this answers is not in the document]
flomillot: I agree, I change it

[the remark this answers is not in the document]
flomillot: I replace diagram data by diagram list, data comes later IMO when we fetch SVG

src/components/diagrams/diagram-pane.js
flomillot: Ok done

src/components/diagrams/diagram-pane.js
flomillot: Because a NAD, but different one (less or more VL inside, or different depth) can be present. In this case we replace it.
The boolean is here to check if we need to update (by adding or replacing) the NAD view.

src/components/diagrams/diagram-pane.js
flomillot: Ok done

src/components/diagrams/diagram-pane.js
flomillot: Then I should rename "addSLDs" in "addNewSLDs" ?

src/components/diagrams/diagram-pane.js
flomillot: Ok indeed clearer

src/components/diagrams/diagram-pane.js
flomillot: It will be duplicate with the comment above

src/components/diagrams/diagram-pane.js
flomillot: I kept the previous design but I agree with you

src/components/diagrams/diagram-pane.js
flomillot: We discussed about

flomillot:
TODO : do the same for PositionDiagram too

src/components/diagrams/diagram-pane.js
```diff
@@ -1062,7 +1062,7 @@ export function DiagramPane({
      */
 
     const handleWarningToDisplay = (diagramView) => {
-        // First, check if the node is built(the highest priority) then do the warning checks..
+        // First, we check if the node is built (the highest priority), so when disabled is true
```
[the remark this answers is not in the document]
EstherDarkish: Yes, the whole comment is useless and worse, confusing.

EstherDarkish (approved):
OK

SlimaneAmar (approved):

PR #1221, opened 2023-04-12
Pass missing translation intl errors to warning level

sBouzols:
anyway it will only appears in dev mode not in production mode

see : https://formatjs.io/docs/react-intl/components/#formattedmessage
and : https://github.com/formatjs/formatjs/issues/465

A Warning will remain.


jonenst:
Not sure about this... What is the usecase for making the message less visible ? Don't we always want to add the missing translation ?

PR #1222, opened 2023-04-12
Dynamic simulation - lack of translation for fetching errors

thangqp:
Correction after merge PR : https://github.com/gridsuite/gridstudy-app/pull/1131
In case of error on back, the title of snack bar is not translated as bellow

image


thangqp:
After corrected 
image

PR #1231, opened 2023-04-18
Exclude network redux from some components

flomillot:
Remove the usage of the network stored in redux from the maximum number of components (except tables).
We replaced the usage of network by http calls to the servers.
To do this, we had to add some endpoints and / or more information inside the DTOs from the back-end.
Some work still need to be done to optimize the number of calls.


jonenst (approved):

SlimaneAmar:
I'm going to create an US to reduce the number of queries by enriching the result of other queries

PR #1237, opened 2023-04-21
Add nested menus in tree node menu

ghazwarhili:
To fix the issue of **SyntaxError: Cannot use import statement outside a module** 
we need to update the mui-nested-men dependencie version
the link below gives details of the issue and the fix https://github.com/steviebaa/mui-nested-menu/issues/20


src/components/graph/menus/create-node-menu.js
```diff
@@ -193,26 +215,112 @@ const CreateNodeMenu = ({
             onClose={handleClose}
         >
             {Object.values(NODE_MENU_ITEMS).map((item) => {
+                const isCreate = item.id === 'createNetworkModificationNode';
```
sBouzols, line 218: This is strnage...
Maybe you should have a NODE_MENU_ITEMS definition containing some nested Menu
ex:
```
const NODE_MENU_ITEMS = {
        MENU_ITEM_A: {
            onRoot: true,
            action: () => foo(),
            id: 'myIdA',
            subMenuItems: {
                MENU_ITEM_B: {
                    id: tutu,
                    ...
                }
            }
        },
```
Then you could create a recursive function, which will render the corresponding JSX Menu with `NestedMenuItem` when needed (i.e when subMenuItems !== undefined and has at least one menu item...)

src/translations/en.json
```diff
@@ -756,14 +757,14 @@
...
+    "insertNetworkModificationNodeBelow": "below in the current branch",
     "copyNetworkModificationNode": "Copy node",
     "cutNetworkModificationNode": "Cut node",
     "cancelCutNetworkModificationNode": "Cancel cut node",
-    "pasteNetworkModificationNodeAbove": "Paste node above in the current branch",
-    "pasteNetworkModificationNodeBelow": "Paste node below in the current branch",
-    "pasteNetworkModificationNodeOnNewBranch": "Paste node in a new branch",
+    "pasteNetworkModificationNodeAbove": "above in the current branch",
+    "pasteNetworkModificationNodeBelow": "below in the current branch",
+    "pasteNetworkModificationNodeOnNewBranch": "Paste node",
```
sBouzols, line 767: fix key -> pasteNetworkModificationNode~~OnNewBranch~~

package.json
```diff
@@ -26,15 +26,15 @@
         "localized-countries": "^2.0.0",
         "lucene-escape-query": "^1.0.1",
         "mjolnir.js": "^2.4.0",
-        "mui-nested-menu": "^3.0.0",
+        "mui-nested-menu": "^3.2.1",
```
sBouzols, line 29: please keep the form ^3.0.0 in here, even if you should update the dep in package-lock

package.json
```diff
@@ -26,15 +26,15 @@
...
         "notistack": "^3.0.0",
         "plotly.js-basic-dist-min": "^2.17.0",
         "qs": "^6.9.1",
         "react": "^18.0.0",
         "react-beautiful-dnd": "^13.1.0",
         "react-csv-downloader": "^2.7.1",
         "react-dom": "^18.0.0",
-        "react-flow-renderer": "^10.0.0",
+        "react-flow-renderer": "10.3.17",
```
sBouzols, line 37: revert this change

package-lock.json
```diff
@@ -10147,9 +10147,9 @@
             "integrity": "sha512-sGkPx+VjMtmA6MX27oA4FBFELFCZZ4S4XqeGOXCv68tT+jb3vk/RyaKWP0PTKyWtmLSM0b+adUTEvbs1PEaH2w=="
         },
         "mui-nested-menu": {
-            "version": "3.2.0",
-            "resolved": "https://registry.npmjs.org/mui-nested-menu/-/mui-nested-menu-3.2.0.tgz",
-            "integrity": "sha512-3js6P2RHkexTZPRFYY5VnkVAncmAWVcZrVX2lqcWRF2X+LjF7v6JmXf3tZHl6w7A2DFbJW5vU1EaUOa48W5Rlg=="
+            "version": "3.2.1",
```
sBouzols, line 10150: If you need to upgrade to this version, please note in comment of the PR why.

sBouzols (changes requested):

src/components/graph/menus/create-node-menu.js
```diff
@@ -193,26 +234,27 @@ const CreateNodeMenu = ({
...
+                            {!hasSubMenuItem && <SubMenuItem item={item} />}
+
+                            {hasSubMenuItem && (
+                                <NestedMenuItem
+                                    label={intl.formatMessage({
+                                        id: item.id,
+                                    })}
+                                    parentMenuOpen={true}
+                                >
+                                    {Object.values(item.subMenuItems).map(
```
sBouzols, line 250: This is not recursive. If we add `subMenuItems` in an item which is already a sub menu item then it won't work.

src/components/graph/menus/create-node-menu.js
```diff
@@ -17,8 +17,14 @@ import PropTypes from 'prop-types';
 import { useIsAnyNodeBuilding } from '../../util/is-any-node-building-hook';
 import { useSelector } from 'react-redux';
 import { CopyType } from '../../network-modification-tree-pane';
+import { NestedMenuItem } from 'mui-nested-menu';
 
 const useStyles = makeStyles((theme) => ({
+    menu: {
+        minWidth: 300,
```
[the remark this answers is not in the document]
sBouzols: removed OK

dbraquart (changes requested):
code review: ok. It does not support recursive nested menus, but this is acceptable for now

tests: works fine, but 1 remark:
1) If clipboard is empty, maybe "Paste node" whole menu should be disabled ?

console:
1) "Warning: Each child in a list should have a unique "key" prop."
-> This is an important React warning that must be adressed
2) "MUI: The Menu component doesn't accept a Fragment as a child. Consider providing an array instead."

dbraquart (commented):
code: ok
test: ok

dbraquart (approved):

src/components/graph/menus/create-node-menu.js
```diff
@@ -255,6 +269,31 @@ const CreateNodeMenu = ({
         },
     };
 
+    const SubMenuItem = ({ item }) => {
```
sBouzols, line 272: It is forbidden to create nested component in another component. Anti pattern, performance issue.
Move this component definintion in another file or outside of the CreateNodeMenu component.
Or create a render function instead.
Also `SubMenuItem` is misnamed. Something like `NodeMenuItem` is better

src/components/graph/menus/create-node-menu.js
```diff
@@ -268,26 +307,33 @@ const CreateNodeMenu = ({
...
+                            {hasSubMenuItem && (
+                                <NestedMenuItem
+                                    label={intl.formatMessage({
+                                        id: item.id,
+                                    })}
+                                    parentMenuOpen={true}
+                                    disabled={item.disabled}
+                                >
+                                    {Object.values(item.subMenuItems).map(
+                                        (item) => {
```
sBouzols, line 325: use another name here to avoid confusion with the existing `item` variable.
Something like `(subItem)` should be ok

sBouzols (changes requested):
Anti Pattern : Nested component

src/components/graph/menus/create-node-menu.js
sBouzols: done

sBouzols (approved):
Code Review OK
Tests OK
Console Warning Check OK

PR #1239, opened 2023-04-21
fix(parameters): Avoid multiple providers fetch calls

sBouzols:
Move states where needed
Add some useRef to control effect processing
Rename AdvancedParameterButton -> SubgroupParametersButton

### DIFF
--- a/src/components/dialogs/parameters/load-flow-parameters.js
+++ b/src/components/dialogs/parameters/load-flow-parameters.js
@@ -362,22 +362,6 @@ const SpecificLoadFlowParameters = ({
         [commitLFParameter, currentProvider, defaultValues, lfParams]
     );
 
-    const paramsComponent = useMemo(() => {
-        return (
-            <FlatParameters
-                className={classes.parameterName}
-                paramsAsArray={specificParamsDescription}
-                initValues={currentParameters}
-                onChange={onChange}
-            />
-        );
-    }, [
-        specificParamsDescription,
-        currentParameters,
-        onChange,
-        classes.parameterName,
-    ]);
-
     return (
         <>
             <SubgroupParametersButton
@@ -385,7 +369,14 @@ const SpecificLoadFlowParameters = ({
                 label={'showSpecificParameters'}
                 onClick={() => setShowSpecificLfParams(!showSpecificLfParams)}
             />
-            {showSpecificLfParams && paramsComponent}
+            {showSpecificLfParams && (
+                <FlatParameters
+                    className={classes.parameterName}
+                    paramsAsArray={specificParamsDescription}
+                    initValues={currentParameters}
+                    onChange={onChange}
+                />
+            )}
         </>
     );
 };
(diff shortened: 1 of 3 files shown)


src/components/dialogs/parameters/parameters.js
```diff
@@ -333,6 +309,27 @@ export const useParametersBackend = (
         ]
     );
 
+    useEffect(() => {
+        if (user !== null) {
+            backendFetchProviders()
+                .then((providers) => {
+                    // we can consider the provider get from back will be also used as
```
laurgarn, line 316: get -> gotten

laurgarn (approved):
Code : OK
Test : OK
console : no worse

src/components/dialogs/parameters/load-flow-parameters.js
```diff
@@ -253,12 +253,9 @@ const AdvancedParameterButton = ({ showOpenIcon, label, callback }) => {
...
 };
 
-const AdvancedLoadFlowParameters = ({
-    lfParams,
-    commitLFParameter,
-    showAdvancedLfParams,
-    setShowAdvancedLfParams,
-}) => {
+const AdvancedLoadFlowParameters = ({ lfParams, commitLFParameter }) => {
+    const [showAdvancedLfParams, setShowAdvancedLfParams] = useState(false);
```
dbraquart, line 257: moving these states into inner components has a little behaviour impact:
when we expand "advanced parameters" and/or "specific parameters", and then we visit others tabs from the Parameters dialog, and go back to LF parameters,  then "advanced parameters" and/or "specific parameters" are not expanded anymore.

This is a small acceptable change (to me), but I wanted to report it.

src/components/dialogs/parameters/parameters.js
```diff
@@ -265,14 +248,7 @@ export const useParametersBackend = (
...
             });
-    }, [
-        type,
-        backendFetchDefaultProvider,
-        providers,
-        provider,
-        updateProvider,
-        snackError,
-    ]);
+    }, [type, backendFetchDefaultProvider, updateProvider, snackError]);
```
dbraquart, line 251: yes, removing 'providers' is  a good point (seems to avoid multiple effects)

dbraquart (approved):

src/components/dialogs/parameters/parameters.js
sBouzols: fixed even if not my text

src/components/dialogs/parameters/load-flow-parameters.js
sBouzols: You are right ! It's a regression.
Fixed, tell we if it's ok now ?

src/components/dialogs/parameters/parameters.js
sBouzols: Did you see why this effect was even calling `resetProvider` sometimes...
this was random, depending on two fetch performance... amazing...

dbraquart: non did not see that

src/components/dialogs/parameters/load-flow-parameters.js
dbraquart: good!

dbraquart (approved):

PR #1244, opened 2023-04-25
Fix error when validating generator modifcation after editing

YenguiSeddik:


src/components/dialogs/generator/modification/generator-modification-dialog.js
```diff
@@ -337,29 +335,34 @@ const GeneratorModificationDialog = ({
                                 );
                             } else {
                                 // on first render, we need to adjust the UI for the reactive capability curve table
-                                const currentReactiveCapabilityCurveTable =
-                                    getValues(REACTIVE_CAPABILITY_CURVE_TABLE);
-                                const sizeDiff =
-                                    previousReactiveCapabilityCurveTable.length -
-                                    currentReactiveCapabilityCurveTable.length;
+                                if (previousReactiveCapabilityCurveTable) {
```
archendel, line 338: improve comment to explain why we are checking previousReactiveCapabilityCurveTable value here

archendel (approved):

PR #1253, opened 2023-04-27
Fix Control the effect in SpecificLoadFlowParameters 

sBouzols:
to update parameters when `provider` and `defaultValues` changes but not when `lfParams` changes otherwise you enter a loop of `setCurrentParameters` calls

### DIFF
--- a/src/components/dialogs/parameters/load-flow-parameters.js
+++ b/src/components/dialogs/parameters/load-flow-parameters.js
@@ -344,7 +344,7 @@ const SpecificLoadFlowParameters = ({
             ...defaultValues,
             ...lfParamsRef.current,
         });
-    }, [currentProvider, defaultValues]);
+    }, [defaultValues]);
 
     const onChange = useCallback(
         (paramName, value, isEdit) => {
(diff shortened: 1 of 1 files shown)


etiennehomer (approved):

laurgarn (approved):

src/components/dialogs/parameters/load-flow-parameters.js
```diff
@@ -328,12 +334,17 @@ const SpecificLoadFlowParameters = ({
...
+
+    // When provider changes or defaultValues then we must reset currentParameters state
     useEffect(() => {
         setCurrentParameters({
             ...defaultValues,
-            ...lfParams?.specificParametersPerProvider?.[currentProvider],
+            ...lfParamsRef.current,
         });
-    }, [lfParams, currentProvider, defaultValues]);
+    }, [currentProvider, defaultValues]);
```
dbraquart, line 347: it looks like we dont need currentProvider as dep anymore.
eslint does not say anything, curious.

dbraquart (approved):

src/components/dialogs/parameters/load-flow-parameters.js
sBouzols: done

PR #1254, opened 2023-04-27
Add delay for upateParameters call

sBouzols:


Meklo (approved):

src/components/dialogs/parameters/parameters.js
```diff
@@ -272,16 +277,21 @@ export const useParametersBackend = (
 
     const updateParameter = useCallback(
```
SlimaneAmar, line 278: Maybe rename by updateParameter**s**

PR #1261, opened 2023-05-02
Add node subtree actions

Meklo:


ayolab (approved):

src/components/graph/util/model-functions.js
```diff
@@ -104,3 +104,15 @@ export function isNodeInNotificationList(node, notificationIdList) {
...
+export function getAllChildren(elements, nodeId) {
+    const selectedNode = elements.treeNodes.find((node) => node.id === nodeId);
+    const directChildren = elements.treeNodes.filter(
+        (node) => node.data.parentNodeUuid === selectedNode.id
+    );
+    let allChildren = [...directChildren];
+    directChildren.forEach((child) => {
+        allChildren = allChildren.concat(getAllChildren(elements, child.id));
+    });
+    return allChildren;
```
YenguiSeddik, line 117: Maybe here, we can do instead:
return [...directChildren, ...directChildren.map((child) => getAllChildren(elements, child.id))];

YenguiSeddik (approved):

src/components/graph/util/model-functions.js
Meklo: Unfortunaly it doesn't do the trick, this way of doing populates subarrays which breaks the system

PR #1264, opened 2023-05-03
fix vl creation when copy vl of kind busbreaker

Walid-Sahnoun:


ayolab (approved):

src/components/graph/menus/modification-list-item.js
```diff
@@ -67,7 +67,7 @@ export const ModificationListItem = ({
     const classes = useStyles();
     const studyUuid = useSelector((state) => state.studyUuid);
     const currentNode = useSelector((state) => state.currentTreeNode);
-    const [computedValues, setComputedValues] = useState({});
+    const [computedValues, setComputedValues] = useState();
```
FranckLecuyer, line 70: why this change ? is there no risk of regression ?

FranckLecuyer (changes requested):

src/components/graph/menus/modification-list-item.js
Walid-Sahnoun: this change because there is an error of translation showed in console
like this, i fixed and there is no regression,
i tested multiple modifications and there is no problem

FranckLecuyer (approved):

PR #1273, opened 2023-05-09
Generation dispatch : reducing generator maxP with forced and planned outage rate startup extension

FranckLecuyer:


jonenst (approved):

src/components/dialogs/generation-dispatch/generation-dispatch-form.js
```diff
@@ -4,48 +4,96 @@
...
-        const lossCoefficientValue = formatPercentageValue(value);
-        setValue(LOSS_COEFFICIENT, lossCoefficientValue, {
+    const handleCoefficientValueChange = (id, value) => {
+        const coefficientValue = formatPercentageValue(value);
+        setValue(id, coefficientValue, {
             shouldValidate: true,
         });
-        return lossCoefficientValue;
+        return coefficientValue;
     };
```
archendel, line 46: why do we need setValue in outputTransform ?
it means we are calling "setValue" right before setting the value in the FloatInput
I don't think it's necessary

archendel (changes requested):
Test: OK
Small code change

src/components/dialogs/generation-dispatch/generation-dispatch-form.js
FranckLecuyer: You're right
I have removed this unneccessary code

PR #1274, opened 2023-05-10
WIP add backend perf custom hook

ghazwarhili:


src/components/utils/inputs/input-hooks.js
```diff
@@ -490,6 +490,10 @@ export const useValidNodeName = ({ studyUuid, defaultValue, triggerReset }) => {
         },
         [studyUuid, intl, defaultValue, snackError]
     );
+    const debouncedValidName = useMemo(
```
sBouzols, line 493: useCallback

sBouzols: You must take the name parameter here as well

src/components/utils/search-matching-equipments.js
```diff
@@ -69,9 +60,14 @@ export const useSearchMatchingEquipments = (
         ]
     );
 
+    const delayedSearchMatchingEquipments = useMemo(
```
sBouzols, line 63: useCallback

src/components/utils/search-matching-equipments.js
```diff
@@ -21,41 +22,31 @@ export const useSearchMatchingEquipments = (
...
-                        .catch((error) => {
-                            snackError({
-                                messageTxt: error.message,
-                                headerId: 'equipmentsSearchingError',
-                            });
-                        });
-                },
-                sooner ? 10 : SEARCH_FETCH_TIMEOUT_MILLIS
-            );
+        (searchTerm) => {
```
sBouzols, line 29: how did you replaced the `sooner = false` parameter usage ?

src/components/dialogs/parameters/parameters.js
```diff
@@ -274,6 +281,11 @@ export const useParametersBackend = (
         ]
     );
 
+    const delayedUpdateParameter = useMemo(
```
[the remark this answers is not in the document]
ghazwarhili: I have doubt that `useCallback` is less performant than `useMemo` 
For reference : https://dmitripavlutin.com/react-throttle-debounce/#3-debouncing-a-callback-second-attempt

src/components/utils/inputs/input-hooks.js
ghazwarhili: Can you clear this point ? `validName` is not enough ?

src/components/utils/search-matching-equipments.js
ghazwarhili: I can't see the utility for the `sooner` param after `debounce` impl for this component

src/utils/use-debounce.js
```diff
@@ -0,0 +1,11 @@
+import { useMemo } from 'react';
```
jonenst, line 1: file header

src/utils/use-debounce.js
```diff
@@ -0,0 +1,11 @@
...
+import { debounce } from '@mui/material/utils';
+
+const useDebounce = (updateParameter, delay) => {
```
sBouzols, line 4: rename `updateParameter` to `func` because it's very generic.
It is as much generic that you should add this custom hook in commons-ui I think

src/components/dialogs/parameters/parameters.js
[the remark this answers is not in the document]
sBouzols: you are totally right nice catch ! thanks for this feeback !

src/components/utils/search-matching-equipments.js
sBouzols: nope useMemo is better you are right

src/components/utils/inputs/input-hooks.js
sBouzols: my bad and useMemo is better you are right

src/components/utils/search-matching-equipments.js
sBouzols: ok then

jonenst (approved):

PR #1277, opened 2023-05-11
add sld line popover

AbdelHedhili:


src/components/tooltips/line-popover.js
```diff
@@ -0,0 +1,290 @@
...
+        setLineInfo(null);
+    };
+
+    const checkValue = (value) => {
+        if (value) {
+            return value;
+        } else {
+            return '';
+        }
+    };
```
Walid-Sahnoun, line 53: you can change it to (remove else)
const checkValue = (value) => {
        if (value) {
            return value;
        } 
        return '';
    };

Walid-Sahnoun (approved):

src/components/tooltips/line-popover.js
```diff
@@ -0,0 +1,290 @@
...
+                                            <TableCell>
+                                                {intl.formatMessage({
+                                                    id: 'I_(A)',
+                                                })}
+                                            </TableCell>
+                                            <TableCell>
+                                                {loadFlowStatus ===
+                                                RunningStatus.SUCCEED
+                                                    ? checkValue(
+                                                          Math.round(
```
EstherDarkish, line 214: If the rounded value is zero, then there is no displayed value at all, because checkValue filters it.

src/components/tooltips/line-popover.js
```diff
@@ -0,0 +1,290 @@
...
+                        </Grid>
+
+                        <Grid item>
+                            <TableContainer component={Paper}>
+                                <Table size={'small'}>
+                                    <TableHead>
+                                        <TableRow>
+                                            <TableCell>
+                                                {intl.formatMessage({
+                                                    id: 'Intensity',
```
EstherDarkish, line 188: We need :
Intensité (fr) = Current (en)

Correct key : "CURRENT"

src/components/tooltips/line-popover.js
```diff
@@ -0,0 +1,290 @@
...
+                        </Grid>
+
+                        <Grid item>
+                            <TableContainer component={Paper}>
+                                <Table size={'small'}>
+                                    <TableHead>
+                                        <TableRow>
+                                            <TableCell>
+                                                {intl.formatMessage({
+                                                    id: 'LimitType',
```
EstherDarkish, line 243: We need :
Seuil (fr) = Limit name (en) 

A new key is needed.

src/components/tooltips/line-popover.js
```diff
@@ -0,0 +1,290 @@
...
+                                                })}
+                                            </TableCell>
+                                            <TableCell>
+                                                {intl.formatMessage({
+                                                    id: 'LimitLabel',
+                                                })}
+                                            </TableCell>
+                                            <TableCell>
+                                                {intl.formatMessage({
+                                                    id: 'Loading',
```
EstherDarkish, line 253: We need "Load (%)" instead of "Loading (%)"

EstherDarkish (changes requested):
When a popover is visible, if we right clic on the line's name in the SLD, the popover is moved to the top left of the screen and do not go away. This is problematic if we do something like "View in spreadsheet" : the screen changes, we cannot remove the popover, and it is blocking a part of the screen.

[EDIT] => Fixed. Thanks

src/components/tooltips/line-popover.js
AbdelHedhili: fixed

src/components/tooltips/line-popover.js
AbdelHedhili: fixed

src/components/tooltips/line-popover.js
AbdelHedhili: fixed

src/components/tooltips/line-popover.js
AbdelHedhili: fixed

EstherDarkish (approved):

Walid-Sahnoun (approved):

ayolab (approved):

PR #1279, opened 2023-05-12
Generation dispatch define and use non adjustable generators

FranckLecuyer:


archendel (approved):
same remark than the previous PR for handleCoefficientValueChange
but it should be treated in the other branch

Test: OK

PR #1281, opened 2023-05-15
Separate network elements info types second part

AAJELLAL:


FranckLecuyer (approved):

Walid-Sahnoun (approved):

src/components/dialogs/generator/modification/generator-modification-dialog.js
```diff
@@ -308,10 +312,11 @@ const GeneratorModificationDialog = ({
         (equipmentId) => {
             if (equipmentId) {
                 setDataFetchStatus(FetchStatus.RUNNING);
-                fetchEquipmentInfos(
+                fetchNetworkElementInfos(
                     studyUuid,
                     currentNodeUuid,
-                    'generators',
+                    EQUIPMENT_TYPES.GENERATOR.type,
```
flomillot, line 318: @SlimaneAmar if we rename "equipment" into "network element", we should rename it everywhere, including the constants no ?

PR #1283, opened 2023-05-15
Fix for substation update modification

dbraquart:
- correct display when we edit a modification in a node not built
- deletion mark must dirty the form
- when we edit a form, "validate" button must be disabled


ayolab (approved):

FranckLecuyer (approved):
Code review: OK
Tests: OK (except that an extra rendering occurs when displaying the old properties values)

PR #1291, opened 2023-05-16
Add popover when onhover line

souissimai:


ayolab (approved):

Walid-Sahnoun (approved):

src/components/tooltips/line-popover.js
```diff
@@ -36,22 +36,28 @@ const useStyles = makeStyles((theme) => ({
...
     useEffect(() => {
-        if (lineId !== '') {
+        if (lineInfos) {
+            setLineInfo(lineInfos);
+        } else if (lineId && lineId !== '') {
             fetchLine(studyUuid, currentNode.id, lineId).then((value) => {
                 setLineInfo(value);
             });
-        } else {
-            setLineInfo(null);
```
AbdelHedhili, line 52: why did you remove that part ? I think it's good practice to empty the state when the data is no longer relevant so that we don't accidentally print old data for a few frames (even if there are others controls that prevents it)

AbdelHedhili (changes requested):

AbdelHedhili (approved):

PR #1293, opened 2023-05-17
Restore node clipboard invalidation scenarios

Meklo:


src/components/network-modification-tree-pane.js
```diff
@@ -182,6 +182,31 @@ export const NetworkModificationTreePane = ({
         [studyUuid, dispatch]
     );
 
+    const isSubtreeImpacted = useCallback(
+        (headerName) => {
+            const nodesImpacted = Array.isArray(
```
Meklo, line 187: Depending on the notification header containing node information we have to format the variable to test its value, its either an array or a single value

sBouzols: Maybe `isSubtreeImpacted` should accept an array of nodes as parameter no matters what.
It should be independant of the notification specific state `studyUpdatedForce.eventData.headers` I think. 

Then when consuming notification you pass either `isSubtreeImpacted([studyUpdatedForce.eventData.headers['parentNode']])` or
`isSubtreeImpacted(studyUpdatedForce.eventData.headers['nodes'])` 
(see brackets)

src/components/network-modification-tree-pane.js
```diff
@@ -342,6 +350,23 @@ export const NetworkModificationTreePane = ({
...
+                    studyUpdatedForce.eventData.headers['updateType']
+                )
+            ) {
+                if (
+                    (true === isInitiatingCopyTab.current &&
+                        studyUpdatedForce.eventData.headers['parentNode'] ===
+                            selectionForCopyRef.current.nodeId) ||
+                    isSubtreeImpacted('parentNode')
+                ) {
+                    dispatch(setSelectionForCopy(noSelectionForCopy));
```
sBouzols, line 365: It could be nice to place this code in a function (something like `checkClipboard` or something) to avoid reset ( `dispatch(setSelectionForCopy(noSelectionForCopy))`) and snackbar code duplication everywhere.

sBouzols: And do we need to broadcast here (`broadcastChannel.postMessage(noSelectionForCopy);`) ?

sBouzols (changes requested):

sBouzols (approved):
Code Review OK
Console Warning check OK
Tests OK
* copy/cut/remove of subtrees
* Tests with multiple cases of selected nodes in subtree or not.

PR #1294, opened 2023-05-17
improve popover style

AbdelHedhili:


Walid-Sahnoun (approved):
review ok 
tests ok
note : cannot select the text inside the popup because it disappear quickly

PR #1297, opened 2023-05-19
fix control on minActivePower field

Walid-Sahnoun:


EstherDarkish (changes requested):
This fixes the original bug, but introduces another bug : now we do not test if the minimal value is inferior to the maximal value, when the max value is not explicitly set.

souissimai (approved):

PR #1305, opened 2023-05-23
Dynamic Simulation - Avoid crash in tab Dynamic Simulation of parameters while dynamic mapping server is not available

thangqp:
Use nullish operator ?? when de-structuring a possible nullish value.


dbraquart (approved):
code: ok
tests: ok

PR #1320, opened 2023-05-26
[WIP] refacto_gridStudy_rest_api

ghelloumi:


AAJELLAL (changes requested):
As discussed before, it's better to put all the endpoints related with study-server in a single file.

src/services/config-notification.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright (c) 2022, RTE (http://www.rte-france.com)
```
flomillot, line 2: Should replace all the 2022 by 2023 (we all forget it :) )

src/components/utils/equipment-types.js
```diff
@@ -25,66 +15,50 @@ export const EQUIPMENT_INFOS_TYPES = {
 export const EQUIPMENT_TYPES = {
     SUBSTATION: {
         type: 'SUBSTATION',
```
flomillot, line 17: Maybe we can get rid of this nested field now, and just use EQUIPMENT_TYPES.SUBSTATION ?

ghelloumi: It would be preferable to engage another technical story instead, as we currently have numerous conflicts to resolve with this particular US. Therefore, it is crucial to avoid further complicating the situation.

thank you :)

src/services/config-notification.js
ghelloumi: done, thank you

ghelloumi:
> As discussed before, it's better to put all the endpoints related with study-server in a single file.

@AAJELLAL done, thank you

PR #1322, opened 2023-05-26
Recursive version to build node nested contextual menu

sBouzols:
Remove useless savage CSS classes
Remove useless <div>
Use FormattedMessage component when possible instead of intl

Tests done : 
Contextual Nested Menu in tree
Contextual Nested Menu in Map


dbraquart (changes requested):
code: ok, just a log to remove
tests: ok. I even tried a 5-level submenu

dbraquart (approved):
code ok
tests: not done on last commit (remove log)

PR #1324, opened 2023-05-29
 show line separator for specefic params

souissimai:


src/components/dialogs/parameters/load-flow-parameters.js
```diff
@@ -383,6 +383,7 @@ const SpecificLoadFlowParameters = ({
                             currentProvider
                         ],
                     }}
+                    showSeparator={true}
```
AAJELLAL, line 386: showSeparator is enough.

souissimai: done

AAJELLAL (commented):
Test ok but when the dropdown  field is empty it takes up a very large space. Is  this the desired behavior?

AAJELLAL (approved):
Test ok.

Walid-Sahnoun (approved):

PR #1330, opened 2023-05-31
Two windings transformer modification

flomillot:
Add the two windings transformer modification as network modification


FranckLecuyer (approved):

FranckLecuyer (approved):
Code review: OK
Tests: OK

ayolab (approved):
Tests OK
Code OK

dbraquart (commented):
code: ok
console: ok
Tests:
- when we edit ID field and there is a match, all data is updated. When there is no more match,
characteristics tab becomes empty, IST in limit tab as well, but not limit tables. It may justify a change request (to be validated with PO ?)

- test case: edit 2WT GUPIEY631, and change IST on side1 (dont change the limit table).
Result: the last limit IT1 with MAX value is sent to the back/db with "MODIFIED" mark, which seems to be an error.
It does not create any error or functional issue, but it is strange and could create future issues ?

- when I change an empty Rated nominal power (cf like with GUPIEY631), I got this LOG: "Rated nominal power : NaN -> 225.0"
I wonder if the PO will agree with "NaN" (to be asked ?).

PR #1334, opened 2023-06-02
Fix line modification temporary limits

ayolab:
- When we go back on the original ID of the line modification, we get back the previous limits values
- Fix temporaryLimits modificationType sent to backend


flomillot (approved):

dbraquart (approved):
code: ok
tests: ok

PR #1345, opened 2023-06-07
Separate network elements info types second part

AAJELLAL:


src/utils/rest-api.js
```diff
@@ -486,12 +486,13 @@ export function fetchTwoWindingsTransformers(
...
-    return fetchEquipments(
+    return fetchNetworkElementsInfos(
         studyUuid,
         currentNodeUuid,
         substationsIds,
-        'Two windings transformers',
-        '2-windings-transformers'
+        EQUIPMENT_TYPES.TWO_WINDINGS_TRANSFORMER.type,
+        EQUIPMENT_INFOS_TYPES.TAB.type,
+        true
```
archendel, line 495: why do all "inUpstreamBuiltParentNode" are now passed as "true" ? They were passed as "false" before if I'm not wrong

src/utils/rest-api.js
```diff
@@ -486,12 +486,13 @@ export function fetchTwoWindingsTransformers(
     currentNodeUuid,
     substationsIds
 ) {
-    return fetchEquipments(
+    return fetchNetworkElementsInfos(
```
archendel, line 489: I think we could improve readability by changing all "fetch{equipmen}" to "fetch{equipment}{infoType}" or by passing "infoType" as argument

Here we don't really know which info type we are fetching without taking a look at the function implementation

archendel (commented):
some comments, will test soon

SlimaneAmar (approved):

src/utils/rest-api.js
archendel: yes but for all fetchGenerators, fetchBatteries, ... we don't know which infoType we are fetching

archendel (approved):

PR #1347, opened 2023-06-08
Tooltip visibility when onHovering line depends on zoom level

souissimai:


src/components/network/network-map.js
```diff
@@ -505,6 +506,7 @@ NetworkMap.defaultProps = {
     filteredNominalVoltages: null,
     labelsZoomThreshold: 9,
     arrowsZoomThreshold: 7,
+    tooltipZoomThreshold: 5,
```
EstherDarkish, line 509: We tried different values instead of 5 and 7 seems better, can you make the change ?

souissimai: It was validated by the PO

EstherDarkish (approved):

PR #1351, opened 2023-06-12
Add shunt compensator modification

YenguiSeddik:


src/components/dialogs/network-modifications/shunt-compensator/characteristics-pane/characteristics-form-utils.js
```diff
@@ -15,17 +15,21 @@ import {
...
         .number()
         .nullable()
         .when([CHARACTERISTICS_CHOICE], {
-            is: CHARACTERISTICS_CHOICES.SUSCEPTANCE.id,
+            is: (characteristicsChoice) =>
+                characteristicsChoice ===
+                    CHARACTERISTICS_CHOICES.SUSCEPTANCE.id && !isModification,
             then: (schema) => schema.required(),
         }),
     [SHUNT_COMPENSATOR_TYPE]: yup.string().when([CHARACTERISTICS_CHOICE], {
```
Meklo, line 29: Shouldn't this field be nullable it's a modification situation ?

dbraquart (changes requested):
code: ok
console: no warning
tests:
1) Issues:
- "Name", "susceptance" and "A at nominal voltage" fields must be nullable (have a cross to empty the field)
- when we create a modification, and just select an Id, we can validate the form wihout any modification.
In this case we create a modification in db with a change on "voltage_level_id".
We can realize the modification, but there is nothing to do (no logs).

2) remarks:
- we should use the same label in creation and modification sub-menus (same in EN but different in FR)

dbraquart (approved):
code: ok
console: ok
tests: almost ok
- we still can validate an "empty" modification, but I think it can be resolve later by an existing US.

sBouzols (approved):
Code Review OK
Tests OK
Console Warning Check OK

EstherDarkish (commented):
Found an edge case that is maybe a bug

EstherDarkish (approved):

PR #1353, opened 2023-06-12
access to transformers modification dialog from SLD

souissimai:


AbdelHedhili (changes requested):
`Warning: React does not recognize the `previousValue` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `previousvalue` instead. If you accidentally passed it from a parent component, remove it from the DOM element.` in the console the first time you open the dialog

AbdelHedhili:
Also to be discussed with the POs but it's weird having the cross to empty the ID field because you specifically want to edit the 2WT you clicked on + If you click on it, the id is coming back.

AbdelHedhili:
> Also to be discussed with the POs but it's weird having the cross to empty the ID field because you specifically want to edit the 2WT you clicked on + If you click on it, the id is coming back.

answer: It's ok for now it'll be fixed later so there is only the warning issue

souissimai:
> `Warning: React does not recognize the `previousValue`prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase`previousvalue` instead. If you accidentally passed it from a parent component, remove it from the DOM element.` in the console the first time you open the dialog

it's an existing warning , I fixed it

Walid-Sahnoun (approved):
review ok
tests ok

AbdelHedhili (approved):

PR #1369, opened 2023-06-15
Fix NetworkModificationNodeEditor: cleanClipboard, remove abusive rerender case

sBouzols:
fix(NetworkModificationNodeEditor): cleanClipboard do not setCopiedModifications to new empty array, if already empty


src/components/graph/menus/network-modification-node-editor.js
```diff
@@ -176,13 +176,14 @@ const NetworkModificationNodeEditor = () => {
 
     const cleanClipboard = useCallback(() => {
         setCopyInfos(null);
-        setCopiedModifications((old) => {
-            if (old.length > 0) {
+        setCopiedModifications((oldCopiedModifications) => {
+            if (oldCopiedModifications.length > 0) {
```
ghelloumi, line 180: you can simplify to 
`if (oldCopiedModifications.length) {`

sBouzols: fixed

jonenst (approved):

PR #1373, opened 2023-06-16
Add access to modification line from SLD and Network

souissimai:


ayolab (approved):

Walid-Sahnoun (approved):
code ok
test ok but the validate button is activated when the form opened

PR #1374, opened 2023-06-16
Hvdc deletion including shunt compensators

dbraquart:


FranckLecuyer (changes requested):
Tests: 
HVDC deletion option not available in voltage level, substation sld, network map ...

When there is a shunt deletion modification in a node not built, before an hvdc deletion, and we specify in this hvdc deletion to delete this shunt, an error is generated (shunt is not found), which is normal, but maybe this should not stop the treatment and maybe we should be able to delete the hvdc ...

Log of hvdc deletion is generated twice : 
   Hvdc line L has been removed
   equipment of type=HVDC_LINE and id=L deleted

src/components/dialogs/network-modifications/equipment-deletion/equipment-deletion-dialog.js
```diff
@@ -7,7 +7,15 @@
 
 import { yupResolver } from '@hookform/resolvers/yup';
 import yup from 'components/utils/yup-config';
-import { TYPE, EQUIPMENT_ID } from '../../../utils/field-constants';
+import {
+    TYPE,
+    EQUIPMENT_ID,
+    SHUNT_COMPENSATOR_SIDE_1,
+    SHUNT_COMPENSATOR_SIDE_2,
+    HVDC_WITH_LCC,
```
FranckLecuyer, line 15: I don't think hvdc_with_lcc info is usefull ...

See remarks in following PRs: 
https://github.com/gridsuite/network-modification-server/pull/272#discussion_r1239823115
https://github.com/gridsuite/network-map-server/pull/122#discussion_r1239777488

FranckLecuyer (changes requested):

src/utils/rest-api.js
```diff
@@ -436,6 +436,40 @@ export function fetchHvdcLines(studyUuid, currentNodeUuid, substationsIds) {
     );
 }
 
+export function fetchHvdcLinesWithShuntCompensators(
```
SlimaneAmar, line 439: Remove this unused call

src/components/dialogs/network-modifications/equipment-deletion/equipment-deletion-dialog.js
dbraquart: yes/ maybe I can just use mcsOnSideX

dbraquart: done

FranckLecuyer (changes requested):

src/components/utils/field-constants.js
```diff
@@ -227,3 +227,11 @@ export const VARIABLE_SHUNT_COMPENSATORS = 'variableShuntCompensators';
 export const PRIORITY = 'priority';
 export const FILTER_ID = 'filterId';
 export const FILTER_NAME = 'filterName';
+
+// HVDC deletion
+export const DELETION_SPECIFIC_DATA = 'specificData';
+export const DELETION_SPECIFIC_TYPE = 'specificType';
+export const HVDC_LINE_LCC_DELETION_SPECIFIC_TYPE = 'HVDC_LINE_WITH_LCC';
```
SlimaneAmar, line 234: Move to `equipment-types.js `file ?

src/components/dialogs/network-modifications/equipment-deletion/equipment-deletion-dialog.js
```diff
@@ -65,6 +72,18 @@ const EquipmentDeletionDialog = ({
...
+                [DELETION_SPECIFIC_DATA]: editData.specificData,
+            });
+        },
+        [reset]
+    );
+
+    const fromMenuDataValues = useCallback(
+        (menuSelectId) => {
+            reset({
+                [TYPE]: EQUIPMENT_TYPES.HVDC_LINE,
```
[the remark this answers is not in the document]
dbraquart: There is a deletion dialog only for HVDC_LINE (having a LCC), so it is hard-coded here.
Maybe we could have an additional component parameter for the type, like:
```
    defaultIdValue, // Used to pre-select an equipmentId when calling this dialog from the SLD/map
    defaultTypeValue, // Used to pre-select an equipment type when calling this dialog from the SLD/map
```
Do you want me to add it ?

src/components/utils/field-constants.js
dbraquart: It is not a real type. It's used as a specific "sub-type" in deletion modification.
let's keep it here ?

FranckLecuyer (approved):

src/components/utils/field-constants.js
SlimaneAmar: yes i think

SlimaneAmar (approved):

PR #1388, opened 2023-06-26
fix temporary limits modification

ayolab:


FranckLecuyer (approved):
Code review: OK
Tests: OK

Walid-Sahnoun (changes requested):
there is small problem at validation in limits table,
when you add a existing limit and validate, the error message appear but i cannot update the added line to fix it

Walid-Sahnoun (approved):
code ok
test ok

PR #1393, opened 2023-06-27
Adjust AgGrid results for LF and AS

souissimai:


trikimahdi (approved):
tests ok
code ok

src/components/security-analysis-result.js
```diff
@@ -412,57 +419,57 @@ const SecurityAnalysisResult = ({ onClickNmKConstraint, result }) => {
     const nmKConstraintsColumns = useMemo(() => {
         return [
             {
-                width: 200,
+                flex: 1,
```
ghelloumi, line 422: you can add flex: 1 only on columnDefs
for both renderTableNmKConstraints and renderTableNmKContingencies (then, you can remove width for columnsNmKContingencies)

archendel (changes requested):
Load sorting is not sorting correctly
It should sort data as numbers but sorts them alphabetically

ghelloumi (approved):
tests ok
code ok

archendel (approved):

PR #1396, opened 2023-06-28
Pass SHUNT_COMPENSATOR_MODIFICATION into developer mode

sBouzols:
I chose the spread way to Initialize an Array with Conditional Elements
cf https://codeburst.io/3-ways-to-initialize-an-array-with-conditional-elements-in-javascript-c95397615a7e

### DIFF
--- a/src/components/graph/menus/create-child-menu-item.js
+++ b/src/components/graph/menus/create-child-menu-item.js
@@ -22,7 +22,7 @@ const useStyles = makeStyles(() => ({
     },
 }));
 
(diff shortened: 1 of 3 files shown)


EstherDarkish (approved):

EstherDarkish (changes requested):
When trying to edit a shunt modification after creating it in dev mode and then going back to normal mode, there is a crash

EstherDarkish (approved):

FranckLecuyer (approved):
Code review:OK
Tests: OK

PR #1398, opened 2023-06-28
When editing a voltage level modification hypothesis, the substation'…

EstherDarkish:
…s default value will not trigger react hook form's dirty flag


sBouzols (approved):
Code Review OK
Tests OK
Console Warning Check OK

flomillot (approved):

PR #1401, opened 2023-06-28
Fix short circuit current units, rename ICC max column to IMACC and rename few variables

etiennehomer:


sBouzols (approved):
Code Review OK

EstherDarkish (approved):

PR #1408, opened 2023-07-04
access to hvdc lines in spreedsheet than vsc converter station

Walid-Sahnoun:


EstherDarkish (changes requested):
There seems to be a bug in the backend : the retrieved type is VSC_CONVERTER_STATION but the ID is from the HVDC. To fix this, we should not change anything in the front, but fix the type in the backend.

If this is really the case, then this PR should be closed.

FranckLecuyer (approved):

EstherDarkish (approved):

PR #1410, opened 2023-07-04
Disable the modification of a shunt compensator with multiple sections.

FranckLecuyer:
To be merged together with PR : https://github.com/gridsuite/network-modification-server/pull/280


Walid-Sahnoun (changes requested):
when the node is not build
when we choose a shunt compensator with multi section
the error message is showed

FranckLecuyer:
> when the node is not build when we choose a shunt compensator with multi section the error message is showed

Walid-Sahnoun (approved):
code ok
test ok

FranckLecuyer:
> > when the node is not build when we choose a shunt compensator with multi section the error message is showed



> When trying to enter a non existant Id, we display an empty dialog: shunt

Right !
I fixed it

PR #1425, opened 2023-07-20
Update librairies and add missing TS types

Tristan-WorkGH:
An update of librairies using `npm update` to have latest definitions of types in Typescript.
Also add some missing `@types/*` to have all `.d.ts` definitions, in TS.


Tristan-WorkGH:
Did we upgrade these libraries while we're at it?

image

Tristan-WorkGH:
will be done later on all projects at once

PR #1427, opened 2023-07-24
Bump TypeScript to 5.1 & add missing TS types

Tristan-WorkGH:
Little PR to:
  - bump `typescript` from `5.0.4` to latest `5.1.6`, with `@types/react`
    - `react` minor update to be sure it match `@types/react` definitions
    - main reason is to have fix to some definition, like `FunctionComponent` having `ReactNode` instead of `ReactElement`
  - fix TSlint error in `src/components/custom-aggrid/custom-aggrid.tsx`
    ```
    TS7006: Parameter 'props' implicitly has an 'any' type.
    ```
  - update `react-scripts` to support typescript 5.1 (warning during build)
  - add devDeps as a warning during build said
    ```
    One of your dependencies, babel-preset-react-app, is importing the "@babel/plugin-proposal-private-property-in-object" package without declaring it in its dependencies. This is currently working because "@babel/plugin-proposal-private-property-in-object" is already in your node_modules folder for unrelated reasons, but it may break at any time.

    babel-preset-react-app is part of the create-react-app project, which is not maintianed anymore. It is thus unlikely that this bug will ever be fixed. Add "@babel/plugin-proposal-private-property-in-object" to your devDependencies to work around this error. This will make this message go away.
    ```

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
         "@svgdotjs/svg.js": "^3.0.12",
         "@svgdotjs/svg.panzoom.js": "^2.1.2",
         "@types/node": "^20.2.3",
-        "@types/react": "^18.2.7",
+        "@types/react": "^18.2.17",
         "@types/react-dom": "^18.2.4",
         "ag-grid-community": "^29.0.0",
         "ag-grid-react": "^29.0.0",
@@ -35,7 +35,7 @@
         "notistack": "^3.0.0",
         "plotly.js-basic-dist-min": "^2.17.0",
         "qs": "^6.9.1",
-        "react": "^18.0.0",
+        "react": "^18.2.0",
         "react-beautiful-dnd": "^13.1.0",
         "react-csv-downloader": "^2.7.1",
         "react-dom": "^18.0.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -16,9 +16,23 @@
...
         "@svgdotjs/svg.js": "^3.0.12",
         "@svgdotjs/svg.panzoom.js": "^2.1.2",
+        "@types/core-js": "^2.5.5",
+        "@types/dagre": "^0.7.48",
+        "@types/localized-countries": "^2.0.0",
+        "@types/lucene-escape-query": "^1.0.0",
         "@types/node": "^20.2.3",
+        "@types/plotly.js-basic-dist-min": "^2.12.1",
+        "@types/qs": "^6.9.7",
         "@types/react": "^18.2.7",
```
archendel, line 26: need to update this to at least 18.2.8
this whole update was meant to make FunctionComponent work without any bug
we need at least typescript 5.1.* and @types/react 18.2.8+

src/components/custom-aggrid/custom-aggrid.tsx
```diff
@@ -43,7 +43,7 @@ const useStyles = makeStyles<GridStudyTheme, CustomAGGGridStyleProps>(
 
             // hides right border for header of "Edit" column due to column being pinned
             '& .ag-pinned-left-header': {
-                borderRight: (props) =>
+                borderRight: (props: Record<string, any>) =>
```
archendel, line 46: why do we need to change this ? It was already well typed

package.json
archendel, line 1: should we include all types now, or add them one by one when we will need them ?

archendel (changes requested):

src/components/custom-aggrid/custom-aggrid.tsx
Tristan-WorkGH: thank @archendel for the pinpoint

package.json
Tristan-WorkGH: yes I think, because it seems the majority is already used (I haven't check precisely), and we avoid for future pr that the definitions was forgotten.

Tristan-WorkGH:
I have split this PR in smaller ones.
The description of this one is updated.

PR #1428, opened 2023-07-24
Stops the silent renew from hiding the parameters dialog

EstherDarkish:
Stops the silent renew from hiding the parameters dialog, and hide the parameters dialog when the user is logged out.


jonenst (approved):

flomillot (approved):
Just code

PR #1430, opened 2023-07-25
Remove study type from short circuit params

AAJELLAL:


src/components/dialogs/parameters/short-circuit-parameters.js
```diff
@@ -145,14 +145,6 @@ const BasicShortCircuitParameters = ({
             type: TYPES.bool,
             description: 'descWithFeederResult',
         },
-        studyType: {
-            type: TYPES.enum,
-            description: 'descStudyType',
-            values: {
-                TRANSIENT: 'descTransient',
-                SUB_TRANSIENT: 'descSubTransient',
```
trikimahdi, line 153: Suggestion: you can remove 'descTransient' and  'descSubTransient' from en.json and fr.json since we no longer use these variables in the project

trikimahdi (changes requested):

dbraquart (approved):
code: ok (but unused translations to be removed)
tests: ok

PR #1434, opened 2023-07-26
add multiselect to sensi filters + default sort on sensibility for N-K

ghelloumi:
**BEFORE:**
image

**AFTER:**
image


EstherDarkish (changes requested):
When trying to consult the results, there is an error message :
`HttpResponseError : 500 Internal Server Error, message : Cannot invoke "java.util.Comparator.reversed()" because "direct" is null`

It could be because AGGrid tries to sort by default with a non existant column.

src/components/custom-aggrid/custom-aggrid-header.js
```diff
@@ -131,9 +131,10 @@ const CustomHeaderComponent = ({
                 }}
             >
                 <Autocomplete
-                    value={filterSelectedOption || ''}
+                    multiple
+                    value={filterSelectedOptions || []}
```
flomillot, line 135: How can it be falsy ?

ghelloumi:
> When trying to consult the results, there is an error message : `HttpResponseError : 500 Internal Server Error, message : Cannot invoke "java.util.Comparator.reversed()" because "direct" is null`
> 
> It could be because AGGrid tries to sort by default with a non existant column.

done, thank you.

src/components/custom-aggrid/custom-aggrid-header.js
ghelloumi: right, it can't be falsy. thank you

EstherDarkish (approved):
Code OK, tests OK

src/components/custom-aggrid/custom-aggrid-header.js
```diff
@@ -131,9 +131,10 @@ const CustomHeaderComponent = ({
                 }}
             >
                 <Autocomplete
-                    value={filterSelectedOption || ''}
+                    multiple
+                    value={filterSelectedOptions}
                     isOptionEqualToValue={(option, value) =>
-                        value === '' || option === value
+                        value === [] || option === value
```
sBouzols, line 137: if value ===[] then it matches every options ?

src/components/results/sensitivity-analysis/sensitivity-analysis-result-tab.js
```diff
@@ -23,35 +23,40 @@ const SensitivityAnalysisResultTab = ({ studyUuid, nodeUuid }) => {
...
+        initTable(nOrNkIndex);
+        setSensiKindIndex(newSensiKindIndex);
     };
 
-    const handleSensiNOrNkIndexChange = (_, newTabIndexes) => {
-        setNOrNkIndex(newTabIndexes);
-        initTable();
+    const handleSensiNOrNkIndexChange = (event, newNOrNKIndex) => {
+        initTable(newNOrNKIndex);
+        setNOrNkIndex(newNOrNKIndex);
```
sBouzols, line 59: this code looks like an effect which need to proc when `nOrNkIndex` or `sensiKindIndex` states changes and which call a `initTable` considering those states changes no ?

sBouzols (commented):
some comments will test

src/components/results/sensitivity-analysis/sensitivity-analysis-result-tab.js
```diff
@@ -23,35 +23,40 @@ const SensitivityAnalysisResultTab = ({ studyUuid, nodeUuid }) => {
...
         DATA_KEY_TO_SORT_KEY,
-        { colKey: defaultSortColumn, sortWay: defaultSortOrder }
+        {
+            colKey: defaultSortColumn,
+            sortWay: defaultSortOrder,
+        }
     );
 
-    const initTable = () => {
+    const initTable = (newNOrNKIndex) => {
```
flomillot, line 42:
```suggestion
    const initTable = (nOrNkIndex) => {
```
It's not always new, cf line 53

flomillot (approved):

src/components/custom-aggrid/custom-aggrid-header.js
```diff
@@ -27,7 +27,7 @@ const CustomHeaderComponent = ({
     sortConfig,
     onSortChanged,
     updateFilter,
-    filterSelectedOption,
+    filterSelectedOptions = [],
```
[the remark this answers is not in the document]
ghelloumi: it can be undefined or null but we have to add verifications in next lines like `!!filterSelectedOptions.length` on  line 74

src/components/custom-aggrid/custom-aggrid-header.js
ghelloumi: yes, it can be only an array. I also add [] as default value for filterSelectedOptions to avoid any problem

src/components/results/sensitivity-analysis/sensitivity-analysis-result-tab.js
ghelloumi: done, thank you

src/components/results/sensitivity-analysis/sensitivity-analysis-result-tab.js
ghelloumi: yes exactly, but if I use `useEffect` with deps `[nOrNkIndex, sensiKindIndex]`, then when nOrNkIndex or sensiKindIndex change we will have an api call  (fetchSensitivityAnalysisResult from PagedSensitivityAnalysisResult) with the old sortConfig and then another same api call with the new sortConfig.

So we have to change sortConfig, filterConfig, and [nOrNkIndex or sensiKindIndex] at the same time to have one single api call

src/components/custom-aggrid/custom-aggrid-header.js
ghelloumi: yes, it can be only an array. I also add [] as default value for filterSelectedOptions to avoid any problem

src/components/custom-aggrid/custom-aggrid-header.js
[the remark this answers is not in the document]
sBouzols: ok I see

src/components/results/sensitivity-analysis/sensitivity-analysis-result-tab.js
sBouzols: ok will see later

sBouzols (approved):
Code review OK
Tests OK
Console Warning check OK

PR #1435, opened 2023-07-26
fix sensi result table

anistouri:


src/components/results/sensitivity-analysis/sensitivity-analysis-result.js
```diff
@@ -119,6 +158,8 @@ const SensitivityAnalysisResult = ({
         () => ({
             suppressMovable: true,
             sortable: true,
+            resizable: true,
+            flex: 1,
```
ghelloumi, line 162: why flex: 1 ? when  we have   const onGridReady = useCallback((params) => {
        if (params.api) {
            params.api.sizeColumnsToFit();
        }
    }, []);
    
    do you think that it is necessary?

ghelloumi (commented):
- Even when titles are not truncated, I consistently encounter tooltips.

- On the HD 1366x768 resolution, none of the sort icons will be visible. Therefore, it would be preferable to reduce the width of the first three columns on the initial state (to confirm with the Product Owner).

image

src/components/results/sensitivity-analysis/sensitivity-analysis-result.js
anistouri: Done

ayolab (approved):

ghelloumi (approved):

PR #1444, opened 2023-08-01
Validate button disable

trikimahdi:


AAJELLAL (changes requested):
Code OK.
**Weird behavior:** 
when I open the dialog the validate button is disabled. But when I change tabs the button is clickable even if nothing has changed.

souissimai (approved):
Code Ok
Test Ok, Button validation is activated only if we detect changes

AAJELLAL (approved):
As seen together, at the moment we keep the current behavior . More investigation is needed to fix the behavior of "isDirty()".

PR #1445, opened 2023-08-02
Migrate the RHF inputs from gridstudy-app to commons-ui

flomillot:


archendel (approved):
we'll implement decs.d.ts when needed

flomillot:
> we'll implement decs.d.ts when needed

I'll had the one in gridexplore right before the merge

thangqp (approved):
OK

PR #1446, opened 2023-08-03
Fix when new Colum on Config table added for Spreadsheet

souissimai:


src/components/app.js
```diff
@@ -254,6 +255,23 @@ const App = () => {
                                     REORDERED_COLUMNS_PARAMETER_PREFIX_IN_DATABASE.length
                                 )
                             );
+                            const reordredParams = JSON.parse(param.value);
+                            const tabDefinition = TABLES_DEFINITION_INDEXES.get(
+                                index
+                            ).columns.map((item) => item.id);
+                            if (
+                                tabDefinition.length !== reordredParams.length
```
archendel, line 263:
```suggestion
                                tabDefinition.length > reordredParams.length
```

We only treat the "new fields" case here, not the "remove fields" one

src/components/app.js
```diff
@@ -269,6 +270,64 @@ const App = () => {
                 dispatch(changeLockedColumns(lockedColumnsParams));
             }
             if (dispatchReorderedColumns) {
+                for (let item of reorderedColumnsParams) {
```
SlimaneAmar, line 273: Create a function

SlimaneAmar (changes requested):

archendel (approved):
Test: OK
Code: OK

It might be nice one day to regroup those 3 config params into a single one to prevent desync bug

PR #1448, opened 2023-08-03
scroll to the error message when it's not visible

achour94:


souissimai (changes requested):
please fix prettiers

src/components/utils/rhf-inputs/error-inputs/error-input.js
```diff
@@ -16,6 +16,10 @@ const ErrorInput = ({ name, InputField }) => {
         name,
     });
 
+    const { isSubmitted } = useFormState();
```
flomillot, line 19: No need this, add 
`formState: { isSubmitted },`
inside the `useController`

src/components/utils/rhf-inputs/error-inputs/error-input.js
```diff
@@ -32,11 +36,19 @@ const ErrorInput = ({ name, InputField }) => {
         return {};
     };
 
+    useEffect(() => {
+        if (isSubmitted && error && errorRef?.current) {
```
flomillot, line 40: Why `?` on `errorRef`, how can it be null ?

src/components/utils/rhf-inputs/error-inputs/error-input.js
flomillot: And why using `isSubmitted` instad of `isSubmitting` ?
Then it focus everytime you submit

achour94: if you use isSubmitting you risk to trigger the effect before the validation is done, i added error to the useEffect dependencies so it scroll each time you submit

src/components/utils/rhf-inputs/error-inputs/error-input.js
achour94: because in the first render the errorRef will be null

flomillot: Current will be null, not errorRef

flomillot (approved):

souissimai (approved):

PR #1455, opened 2023-08-04
fix filter icon

anistouri:


ghelloumi (commented):
If I click on filter icon, an unecessary sort event will be here

src/components/custom-aggrid/custom-aggrid-header.js
```diff
@@ -87,14 +87,15 @@ const CustomHeaderComponent = ({
                 item
                 direction={'row'}
                 alignItems={'center'}
+                wrap={'nowrap'}
                 sx={{ height: '100%' }}
                 onClick={handleSortChange}
             >
-                <Grid item xs={11}>
+                <Grid item xs={12}>
```
ghelloumi, line 94: can you reduce the number of grid

src/components/custom-aggrid/custom-aggrid-header.js
```diff
@@ -87,14 +87,15 @@ const CustomHeaderComponent = ({
...
                     <Grid
                         container
                         alignItems={'center'}
-                        sx={{ height: '100%' }}
+                        sx={{ height: '100%', display: 'flex' }}
```
ghelloumi, line 98: I think that Grid container is flex displayed by default

ghelloumi, line 98: I think that Grid container is flex displayed by default

ghelloumi (approved):
test ok
code ok

PR #1463, opened 2023-08-09
Better handling aggrid when data is fetched.

AAJELLAL:


trikimahdi (approved):
Code ok
tests ok on different studies

anistouri (approved):

PR #1465, opened 2023-08-10
Fix hvdc with lcc deletion

dbraquart:
2 error cases fixed:
- when we create a lcc-hvdc deletion modification the MCS list is not saved to database
- in rare cases, when we edit an existing lcc-hvdc deletion modification, the form is updated several times (too many effects) and we dont display the actual content of the modification (the content coming from the back is overwritten with dynamic content).


FranckLecuyer (changes requested):
Tests: PB
On edition of an existing hvdc lcc deletion modification, if i switch to another equipment type (substation for example) and then return to hvdc line type and choose the previous hvdc lcc, the shunt compensator part is not displayed anymore ...
On creation of an hvdc lcc deletion modification, this problem doesn't occur

FranckLecuyer (approved):
Code Review: OK
Tests: OK

PR #1467, opened 2023-08-10
add Alias tooltip on AS result

anistouri:


src/components/security-analysis-result.js
```diff
@@ -211,11 +216,31 @@ const SecurityAnalysisResult = ({ onClickNmKConstraint, result }) => {
...
+            params.data?.contingencyEquipmentsIds
+        ) {
+            const values = Object.values(params.data?.contingencyEquipmentsIds);
+            const valuesToDisplay = values?.join(', ');
+
+            return {
+                title: null,
+                values: valuesToDisplay,
+            };
+        } else {
```
AAJELLAL, line 232: You can retun null without the "else"

src/components/security-analysis-result.js
```diff
@@ -211,11 +216,31 @@ const SecurityAnalysisResult = ({ onClickNmKConstraint, result }) => {
         },
         [onClickNmKConstraint, classes.button]
     );
+
+    const toolTipValueGetterValues = (params) => {
+        if (
+            params?.data?.contingencyId &&
```
AAJELLAL, line 222: Do you really need all the optional chaining?
If yes, you need to add "?" for : params.data?.contingencyEquipmentsIds

src/components/custom-aggrid/custom-tooltip-values.tsx
```diff
@@ -0,0 +1,38 @@
+import { Grid, Typography, Tooltip } from '@mui/material';
+import { ITooltipParams } from 'ag-grid-community';
+import { FunctionComponent } from 'react';
+
+const CustomTooltipValues: FunctionComponent<ITooltipParams> = (props) => {
+    const values: string[] = props?.value?.values;
+    return (
+        <>
+            {values && values.length && (
```
AAJELLAL, line 9: I think it'd better to use values?.length

dbraquart (approved):
code: ok

tests:
	- with explicit naming list: ok
	- with criteria-based list: it works too, but I think the PO does not want a tooltip in this case, because the element Id is always like the contengency Id. The thing is we dont know in which case we are when we look at the SA result => to be discussed with the PO.

src/components/security-analysis-result.js
anistouri: Done

src/components/custom-aggrid/custom-tooltip-values.tsx
anistouri: Done

AAJELLAL (approved):

PR #1469, opened 2023-08-11
Display the static var compensators and vsc converter stations modifications in voltage init modification

FranckLecuyer:


dbraquart (approved):
code: ok
tests: ok

Suggestion: To avoid a scrollbar in both the tab header and the grid, you should put
back the maxWidth="md" for the form dialog.
I tried, it is prettier.
To fit properly with "md", you have to use onRowDataUpdated rather than onGridReady.

ayolab (approved):

PR #1473, opened 2023-08-16
fix spreadsheet selection

dbraquart:
When we select 'view in spreadsheet', we switch to the spreadsheet tab and the target row is now selected again.

Another issue fixed: when we wanted to view an equipement, but its type is already displayed in the spreadsheet, nothing was happening. Now we scroll and select too.


Meklo (changes requested):
I reckon you should either fix the previous system which was working through the state `scrollToIndex` and the method `getRowStyle` in `EquipmentTable` or remove it in favor of what you implemented, AG grid's `setSelected` API call. Personally i'd favour getting rid of the previous system as it is unnecessarily complex

Additionally the other bug you mention in the description is still occurring in my client, the spreadsheet sometimes doesn't display the right line, sometimes goes blank if the same type of equipment as the one we try to view was already displayed

YenguiSeddik (approved):

Meklo (approved):

PR #1475, opened 2023-08-16
keep status and results after refresh

anistouri:


ayolab (approved):

dbraquart (changes requested):
code: your code does not solve the pb
tests:
- it does not work all the time when we re-open a study: we may still lose the computation status and results (it happens when we reset the provider, see below)
- with this PR comes new issues:
	* at study opening, there are no more Rest Get call to xxx-default-provider
	* when we try to change the provider in LF/AS parameters form, the combobox is not updated

With or Without your PR (main branch), I observe at study opening that:
- they are many (normal) calls to get providers, provider, default-provider, parameters ...
- AND some abnormal POST methods to update the provider for the study, for several computations (lf/as/sensi) => this is the root cause (cf we must not call resetProvider() at line 415)

src/components/dialogs/parameters/parameters.js
```diff
@@ -415,7 +410,7 @@ export const useParametersBackend = (
                 .then((provider) => {
                     // if provider is not defined or not among allowed values, it's set to default value
                     if (provider in providersRef.current) {
-                        setProvider(provider);
+                        providerRef.current = provider;
                     } else {
                         resetProvider();
```
dbraquart, line 415: we must not call resetProvider at study opening

src/components/dialogs/parameters/parameters.js
```diff
@@ -302,13 +306,13 @@ export const useParametersBackend = (
     const resetProvider = useCallback(() => {
         backendFetchDefaultProvider()
             .then((defaultProvider) => {
-                const providerNames = Object.keys(providersRef.current);
+                const providerNames = Object.keys(provider);
```
dbraquart, line 309: we have to compare with all providers, like before

PR #1476, opened 2023-08-17
auto focus on error in line dialog

achour94:
focus on the first error input when there is an error in line dialog

### DIFF
--- a/src/components/dialogs/network-modifications/line/creation/line-creation-dialog.js
+++ b/src/components/dialogs/network-modifications/line/creation/line-creation-dialog.js
@@ -144,7 +144,11 @@ const LineCreationDialog = ({
         resolver: yupResolver(formSchema),
     });
 
-    const { reset, setValue, formState : {errors, isSubmitting} } = formMethods;
+    const {
+        reset,
+        setValue,
+        formState: { errors, isSubmitting },
+    } = formMethods;
 
     const fromSearchCopyToFormValues = (line) => {
         reset(
--- a/src/components/dialogs/network-modifications/line/creation/line-creation-dialog.js
+++ b/src/components/dialogs/network-modifications/line/creation/line-creation-dialog.js
@@ -144,7 +144,7 @@ const LineCreationDialog = ({
         resolver: yupResolver(formSchema),
     });
 
-    const { reset, setValue } = formMethods;
+    const { reset, setValue, formState : {errors, isSubmitting} } = formMethods;
 
     const fromSearchCopyToFormValues = (line) => {
         reset(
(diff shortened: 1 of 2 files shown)


src/components/dialogs/network-modifications/line/creation/line-creation-dialog.js
```diff
@@ -349,13 +353,41 @@ const LineCreationDialog = ({
...
+                    let errorId = temporaryLimitsErrors.ref.name;
+                    let errorInput = document.getElementById(errorId);
+                    errorInput.scrollIntoView({
+                        behavior: 'smooth',
+                        block: 'nearest',
+                    });
+                }
+            }
+        }
+    }, [tabIndexesWithError, errors, isSubmitting, tabIndex]);
```
archendel, line 379: I think it would be cleaner to do this in "onValidationError"

"errors" object is updated each time the user will update field contents, this useEffect could execute more often than expected

src/components/dialogs/network-modifications/line/creation/line-creation-dialog.js
```diff
@@ -349,13 +353,41 @@ const LineCreationDialog = ({
...
+                    errors[LIMITS][firstArrayWithErrors]?.[TEMPORARY_LIMITS];
+                if (temporaryLimitsErrors?.ref) {
+                    //case where ErrorInput is rendered
+                    let errorId = temporaryLimitsErrors.ref.name;
+                    let errorInput = document.getElementById(errorId);
+                    errorInput.scrollIntoView({
+                        behavior: 'smooth',
+                        block: 'nearest',
+                    });
+                }
```
archendel, line 376: I don't think it would work as expected if there are other fields than those two arrays, it would scroll to arrays even if another field is in error

we can see it already if you add empty fields to 2nd table without typing any "ID"
If you press "Validate", it will scroll to the table error whereas the first error is for the "ID"

src/components/dialogs/network-modifications/line/creation/line-creation-dialog.js
```diff
@@ -144,7 +144,11 @@ const LineCreationDialog = ({
         resolver: yupResolver(formSchema),
     });
 
-    const { reset, setValue } = formMethods;
+    const {
+        reset,
+        setValue,
+        formState: { errors, isSubmitting },
```
archendel, line 150: be careful if you get values from react hook form in root component, it will rerender the whole dialog each time one of those values changes

archendel (changes requested):

src/components/dialogs/network-modifications/line/creation/line-creation-dialog.js
achour94: i did it in useEffect because i want to do the scroll AFTER the tab is changed
and it does not trigger the useeffect everytime 'errors' is changed, i tested it

flomillot (commented):
The point of this TS was to do something that will be used automatically by all forms, not only for one dialog. I gave it as an exemple, but it can happen everywhere. If prefer the previous code, at least it works for all components. And as suggested by the PO, we can do an other US for this case.

PR #1481, opened 2023-08-21
Create edit filter

Walid-Sahnoun:


flomillot:
We should not copy / paste a huge quantity of code like that.
I think an effort have to be done to mutualize the code and put it inside commons-ui.

src/components/utils/rhf-inputs/ag-grid-table-rhf/csv-uploader/csv-uploader.js
```diff
@@ -0,0 +1,260 @@
+/**
```
YenguiSeddik, line 1: we already have a component to uplode csv ( ImportRuleDialog ). can it be used in this case ?

src/components/utils/rhf-inputs/ag-grid-table-rhf/bottom-right-buttons.js
```diff
@@ -0,0 +1,96 @@
+/**
```
YenguiSeddik, line 1: we have a similar component used with dndTable : DndTableBottomRightButtons. can it be used here ?

YenguiSeddik (changes requested):
Some findings after testing

- When editing a filter, this error is show in the dialog: "Name Error while checking name validity: HttpResponseError : 400 Bad Request, message :" even if the name is valid
- When editing, the selection filter dialog is not closed after a filter is selected
- "Explicit naming" is selected by default (in grid explorer "Criteria based" is the default value)
- some missing translation for errors innaming explicit creation like: missingDistributionKeyError

src/components/utils/rhf-inputs/ag-grid-table-rhf/bottom-right-buttons.js
Walid-Sahnoun: yes, done

src/components/dialogs/filter/custom-mui-dialog.js
```diff
@@ -0,0 +1,113 @@
+/**
```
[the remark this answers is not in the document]
Walid-Sahnoun: yes, i used

FranckLecuyer (approved):
Remark: for a filter with identifier list, when we edit and change the identifiers order, this order is not kept when we reedit the filter ... (another US)

YenguiSeddik (approved):

src/components/dialogs/filter/name-wrapper.js
```diff
@@ -0,0 +1,157 @@
+/**
```
Meklo, line 1: I used this component for the development of voltage init parameter saving and moved the file to dialogs/commons, do you agree to move it there too so we don't duplicate it ?

PR #1482, opened 2023-08-21
change targetDeadband controle

anistouri:


FranckLecuyer (changes requested):
There are still occurrences of targetDeadBand (instead of targetDeadband) in ratio-tap-changer-pane.js and phase-tap-changer-pane.js ...

FranckLecuyer (approved):

Meklo (approved):

PR #1485, opened 2023-08-22
Disable subtree remove option if it has no child.

AAJELLAL:


Walid-Sahnoun (approved):
code ok
test ok

src/components/graph/menus/create-node-menu.js
```diff
@@ -174,7 +174,12 @@ const CreateNodeMenu = ({
             selectionForCopy?.copyType === CopyType.SUBTREE_CUT
         );
     }
-
+    function isSubtreeRemovingAllowed() {
```
trikimahdi, line 177: I suggest a name of function that has relation to check if the parentNode has children. Example: parentNodeHasChildren

trikimahdi (changes requested):

PR #1492, opened 2023-08-23
Improve loader display in sensi result

anistouri:


src/components/dialogs/commons/handle-loader.ts
```diff
@@ -0,0 +1,45 @@
...
+        if (!isLoading) {
+            setShouldOpen(false);
+        } else {
+            // Otherwise, wait for the specified delay before displaying the loader.
+            timeout = setTimeout(() => {
+                setShouldOpen(true);
+            }, delay);
+        }
+        // Return a cleanup function to cancel the timeout if loading finishes before the end of the delay.
+        return () => clearTimeout(timeout);
```
AAJELLAL, line 41: We already have **useDebounce** to handle this kind of timeout. It would be better to use it. What do you think?
Otherwise it's good for me.

AAJELLAL: Ok

AAJELLAL (approved):
Code ok.
Test ok.

PR #1499, opened 2023-08-25
Emotion migration, third batch.

EstherDarkish:


src/components/graph/menus/editable-title.js
```diff
@@ -30,28 +30,27 @@ const useStyles = makeStyles((theme) => ({
...
+        <Box sx={styles.header}>
             <IconButton
                 size={'small'}
                 onClick={() => setOpenEditTitle(true)}
                 disabled={onChange === undefined}
             >
                 <EditIcon />
             </IconButton>
-            <OverflowableText text={name} className={classes.nodeNameTitle} />
+            <OverflowableText text={name} sx={styles.nodeNameTitle} />
```
archendel, line 53: I'm not sure sx will work here
OverflowableText is defined this way :
```
                <Box
                    {...props}
                    ref={element}
                    children={children || text}
                    className={className}
                    sx={overflowStyle.overflow}
                />
```
Since sx is defined again, it might be overridden ?

src/components/graph/menus/modification-list-item.js
```diff
@@ -159,20 +152,16 @@ export const ModificationListItem = ({
...
                                 checked={checked}
                                 onClick={toggle}
                                 disableRipple
                             />
                         </ListItemIcon>
-                        <OverflowableText
-                            className={classes.label}
-                            text={getLabel()}
-                        />
+                        <OverflowableText sx={styles.label} text={getLabel()} />
```
archendel, line 164: same here for sx

src/components/graph/menus/editable-title.js
EstherDarkish: After further testing, it works

src/components/graph/menus/modification-list-item.js
EstherDarkish: After further testing, it works

Meklo (approved):

PR #1507, opened 2023-08-30
Move calculation button to topbar

anistouri:


src/components/app-top-bar.js
```diff
@@ -204,6 +215,7 @@ const AppTopBar = ({ user, tabIndex, onChangeTab, userManager }) => {
     const studyUuid = useSelector((state) => state.studyUuid);
 
     const currentNode = useSelector((state) => state.currentTreeNode);
+    const disabled = !isNodeBuilt(currentNode);
```
EstherDarkish, line 218: I don't like this new value for two reasons :
- "disabled" is too generic a name. It would mean that the topbar is disabled, when it is not (even if the current node is not built, the topbar's other buttons are still working)
- There is already disabled props in this file, and each time their values means something functionally, that's easy to understand.

I would remove this const and set the component's disabled props like this : 
`<RunButtonContainer
                            studyUuid={studyUuid}
                            currentNode={currentNode}
                            setIsComputationRunning={setIsComputationRunning}
                            disabled={!isNodeBuilt(currentNode) || isNodeReadOnly(currentNode)}
                        />`

src/components/app-top-bar.js
```diff
@@ -75,6 +76,10 @@ const styles = {
         margin: theme.spacing(1.5),
         fontWeight: 'bold',
     }),
+    divRunButton: {
+        marginRight: '100px',
```
EstherDarkish, line 80: That would be better with flexible design instead of hard values. With a smaller window, there is a big gap where space could be saved.

src/components/app-top-bar.js
```diff
@@ -170,7 +175,13 @@ const CustomSuffixRenderer = ({ props, element }) => {
     );
 };
 
-const AppTopBar = ({ user, tabIndex, onChangeTab, userManager }) => {
+const AppTopBar = ({
+    user,
+    tabIndex,
+    onChangeTab,
+    userManager,
+    setIsComputationRunning,
```
EstherDarkish, line 183: Instead of going back all the way to the top with this and pass it down only to use it in a SLD, could we move this status in the redux ?

src/components/app-top-bar.js
```diff
@@ -460,7 +472,18 @@ const AppTopBar = ({ user, tabIndex, onChangeTab, userManager }) => {
                         })}
                     </Tabs>
                 )}
+                {studyUuid && (
+                    <Box sx={styles.divRunButton}>
+                        <RunButtonContainer
+                            studyUuid={studyUuid}
+                            currentNode={currentNode}
+                            setIsComputationRunning={setIsComputationRunning}
+                            disabled={disabled || isNodeReadOnly(currentNode)}
```
EstherDarkish, line 481: See my comment above : 
`disabled={!isNodeBuilt(currentNode) || isNodeReadOnly(currentNode)}`

src/components/app.js
```diff
@@ -550,6 +553,9 @@ const App = () => {
                                     <StudyContainer
                                         view={STUDY_VIEWS[tabIndex]}
                                         onChangeTab={onChangeTab}
+                                        isComputationRunning={
+                                            isComputationRunning
```
EstherDarkish, line 557: See my redux remark

src/components/study-container.js
```diff
@@ -201,7 +201,7 @@ const USER_HEADER = 'userId';
 // the delay before we consider the WS truly connected
 const DELAY_BEFORE_WEBSOCKET_CONNECTED = 12000;
 
-export function StudyContainer({ view, onChangeTab }) {
+export function StudyContainer({ view, onChangeTab, isComputationRunning }) {
```
EstherDarkish, line 204: See my redux remark

EstherDarkish (changes requested):
Some structural remarks.

src/components/app-top-bar.js
```diff
@@ -460,7 +472,18 @@ const AppTopBar = ({ user, tabIndex, onChangeTab, userManager }) => {
                         })}
                     </Tabs>
                 )}
+                {studyUuid && (
+                    <Box sx={styles.divRunButton}>
```
ghelloumi, line 476: can you rename 'divRunButton' to 'runButtonContainerBox' for example (try to not use html tags names inside styles names)

src/components/utils/split-button.js
ghelloumi, line 1: I think that you have to change a z-index somewhere to not have this visual issue

image

src/components/utils/split-button.js
```diff
@@ -241,6 +241,9 @@ const SplitButton = ({
                 anchorEl={anchorRef.current}
                 role={undefined}
                 transition
+                style={{
+                    zIndex: 2,
```
ghelloumi, line 245: you can make it 5 for example to solve the problem above

and you can move it to the useStyles to make the migration to emotion easier.

src/components/app-top-bar.js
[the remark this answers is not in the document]
EstherDarkish: Or maybe remove this prop entirely and use the redux instead ?

src/components/app-top-bar.js
anistouri: DOne

src/components/app-top-bar.js
anistouri: Done

src/components/app-top-bar.js
anistouri: Done

src/components/app.js
anistouri: Done

src/components/study-container.js
anistouri: DOne

src/components/app-top-bar.js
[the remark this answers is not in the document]
anistouri: Done

src/components/app-top-bar.js
anistouri: Done

src/components/utils/split-button.js
```diff
@@ -29,14 +29,14 @@ import { RunningStatus } from './running-status';
 
 const useStyles = makeStyles((theme) => ({
     expand: {
-        transform: 'rotate(180deg)',
+        transform: 'rotate(0deg)',
```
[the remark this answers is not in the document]
anistouri: Why do you want to remove it? What do you suggest?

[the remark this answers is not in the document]
ghelloumi: you are rigth, we can't remove it, sorry

ghelloumi (approved):
code ok
test ok

src/components/utils/split-button.js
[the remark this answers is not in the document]
anistouri: Removed

PR #1515, opened 2023-09-01
Fix crash on equipment type selection in equipment deletion dialog

FranckLecuyer:


flomillot (approved):
I feel like we will have more bugs... thanks

PR #1525, opened 2023-09-05
add tooltip everywhere in aggrid

trikimahdi:


flomillot (changes requested):
A global tooltip is already defined here for spreadsheet : https://github.com/gridsuite/gridstudy-app/blob/e3a5afa02e4cde80ff26c89e707f6fda2ab1a6f2/src/components/spreadsheet/utils/cell-renderers.js#L107
So it lead to double tooltip.
Plus I'm not sure the style of the browser tooltip is what the PO wants (need to check with them).
Maybe we should generalize and use the same cell renderer everywhere : https://github.com/gridsuite/gridstudy-app/blob/d25dcaba93a99994cd452d60b41d4d4700298b4d/src/components/spreadsheet/utils/cell-renderers.js#L102
Then the work will have to been done only here, but I think it need to be discussed with a TL and/or @Meklo who setted up AGGrid in the project, because it can leads to performance drop.

trikimahdi:
Another PR is created to implement Tooltip in Aggrid

PR #1539, opened 2023-09-08
restore nodes

AbdelHedhili:


src/components/dialogs/restore-node-dialog.js
```diff
@@ -0,0 +1,146 @@
...
+                        isFixed={false}
+                        loadingMessageText={'LoadingRemoteData'}
+                    />
+                )}
+                {!isLoading && nodes && (
+                    <FormControl
+                        sx={{ paddingLeft: '20px' }}
+                        component="fieldset"
+                    >
+                        <RadioGroup name="nodes-to-restore-selection">
```
Meklo, line 101: On the feature mockup it looks like checkbox were initially intended, was this change intented ?

AbdelHedhili: yes it was discussed and a single choice is what is wanted -> radio

Meklo (approved):

souissimai (approved):

PR #1543, opened 2023-09-08
JSS to Emotion migration : focus on tabs outside of dialogs

EstherDarkish:
Batch 10, focused on tabs outside of dialogs.


Walid-Sahnoun (approved):
code ok
test ok

PR #1548, opened 2023-09-11
fix bug when editing 2TWT created from copy

souissimai:


AAJELLAL (approved):
Code and tests are OK.

PR #1565, opened 2023-09-18
Previsualize voltage init modification

dbraquart:


FranckLecuyer (approved):

src/components/dialogs/network-modifications/voltage-init-modification/voltage-init-modification-dialog.tsx
```diff
@@ -57,6 +57,10 @@ interface CloseFunction {
     (): void;
 }
 
+interface PreviewModeSubmitFunction {
```
ghazwarhili, line 60: IPreviewModeSubmitFunction ?

dbraquart: Do we have a convention for interface names ?
I'm not sure we have to do it, so I prefer not to change (otherwise I should rename all interfaces in this file).

ghazwarhili: okey

PR #1566, opened 2023-09-18
Create unit tests for services 

ghelloumi:
* add some jest global setups
* add jest unit tests for services utils
* add some unit test for some services: user-admin

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,12 +63,19 @@
         "start": "react-scripts start",
         "build": "react-scripts build",
         "test": "react-scripts test --watchAll=false --transformIgnorePatterns \"node_modules/(?!@gridsuite/commons-ui)/\"",
-        "eject": "react-scripts eject"
+        "eject": "react-scripts eject",
+        "jest_clear_cache": "jest --clearCache"
     },
     "eslintConfig": {
+        "plugins": [
+            "jest",
+            "testing-library"
+        ],
         "extends": [
             "react-app",
-            "plugin:prettier/recommended"
+            "plugin:prettier/recommended",
+            "plugin:prettier/recommended",
+            "plugin:jest/recommended"
         ],
         "rules": {
             "prettier/prettier": "warn",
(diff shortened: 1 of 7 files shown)


src/services/utils.test.ts
```diff
@@ -0,0 +1,427 @@
...
+            const result = prepareRequest(init);
+
+            expect(result.headers.get('Authorization')).toBe(
+                'Bearer mocked-token-returned-by-getToken'
+            );
+        });
+    });
+
+    describe('safeFetch', () => {
+        it('should return the response when response.ok is true (get successfully the response)', async () => {
```
trikimahdi, line 275: maybe we should remove the async await since we don't use in the project

trikimahdi (changes requested):

package.json
```diff
@@ -90,9 +96,11 @@
         ]
     },
     "devDependencies": {
-        "http-proxy-middleware": "^2.0.0",
```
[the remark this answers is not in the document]
ghelloumi: line just moved the position

trikimahdi (approved):

AAJELLAL (commented):
You tested only user-admin, what about the other services?

src/setupTests.js
```diff
@@ -5,6 +5,17 @@
...
+import { TextDecoder, TextEncoder } from 'util';
+
+jest.mock('uuid', () => ({
+    v1: () => 'xxx-xxx-xxx-xxx-xxx',
+}));
+
+console.info = jest.fn();
+console.debug = jest.fn();
+
+Object.assign(global, { TextDecoder, TextEncoder });
```
AAJELLAL, line 17: Could you use spread operator plz?

src/services/user-admin.test.ts
```diff
@@ -0,0 +1,78 @@
...
+    });
+
+    it('should reject with an error and not call backendFetch if user.profile.sub is missing', async () => {
+        await expect(fetchValidateUser(mockUser)).rejects.toThrow(
+            'Error : Fetching access for missing user.profile.sub'
+        );
+        expect(mockBackendFetch).not.toHaveBeenCalled();
+    });
+
+    it('should return true if the response status is 200', async () => {
```
AAJELLAL, line 43: It is better to avoid async to be consistent with the project.

AAJELLAL (commented):
You tested user admin service but not the other services.

package.json
```diff
@@ -63,12 +63,18 @@
...
     },
     "eslintConfig": {
+        "plugins": [
+            "jest"
+        ],
         "extends": [
             "react-app",
-            "plugin:prettier/recommended"
+            "plugin:prettier/recommended",
+            "plugin:prettier/recommended",
```
capyq, line 76: duplicate line

PR #1569, opened 2023-09-19
Reset substation spreadsheet if properties change

AAJELLAL:


ghelloumi (approved):

flomillot (commented):
So you delete all the substations and fetch them back just to update one substation ? Why don't you just use the updateEquipments action ?

PR #1570, opened 2023-09-19
Fix the editable cells of VL in the spreadsheet.

flomillot:
Make these fields optionals.
Add checks on ISC min and ISC max.


Meklo (approved):
It's all good on the validation side, but there's an issue on how the value 0 is handled : it's not showing up in the editor because of this snippet `value={value || ''}` but it is still accounted for validation so when iccmax is emptied, the field is marked as in error, because it reads iccmin as 0 although the editor is empty
I'll write it down as a comment in the ticket

achour94 (approved):

flomillot:
> It's all good on the validation side, but there's an issue on how the value 0 is handled : it's not showing up in the editor because of this snippet `value={value || ''}` but it is still accounted for the validation so when iccmax is emptied, the field is marked as in error, because it reads iccmin as 0 although its editor is empty I'll write it down as a comment in the ticket

I fixed this line which is not necessary I think, but an other PR will be necessary to manage correctly the value 0

PR #1577, opened 2023-09-21
Fix unit formatting in resources.

antoinebhs:


src/translations/fr.json
```diff
@@ -419,7 +419,7 @@
     "TargetV": "Consigne tension (kV)",
     "TargetDeadband": "Consigne bande morte",
     "VoltageSetpoint": "Consigne tension (kV)",
-    "ReactivePowerSetpoint": "Consigne puissance réactive (MVAR)",
+    "ReactivePowerSetpoint": "Consigne puissance réactive (MVar)",
```
FranckLecuyer, line 422: This translation should not add unit (see english translation where the unit is absent)
there is another key 'ReactivePowerSetpointMVAR' that adds the unit ...

FranckLecuyer (changes requested):

src/translations/fr.json
antoinebhs: @FranckLecuyer You're right. There was an inconsistent use of ReactivePowerSetpoint with and without unit. Same for VoltageSetpoint. I removed these two resources and replace their use with the one with units in the STATIC_VAR_COMPENSATORS table as we expect units there.

FranckLecuyer (approved):

PR #1578, opened 2023-09-21
Editable name and nominal voltage of VL in spreadsheet

flomillot:


src/components/spreadsheet/utils/config-tables.js
```diff
@@ -144,6 +145,17 @@ export const TABLES_DEFINITIONS = {
                 numeric: true,
                 filter: 'agNumberColumnFilter',
                 fractionDigits: 0,
+                editable: true,
+                cellEditor: NumericalField,
+                cellEditorParams: (params) => {
```
achour94, line 150: the nominal voltage is mandatory in the creation form, it should be mandatory here too i think

achour94 (changes requested):

src/components/spreadsheet/utils/config-tables.js
flomillot: It's fixed now normally

anistouri (approved):

PR #1587, opened 2023-09-22
Fixes a warning in the console related to kebab-case of CSS definition.

EstherDarkish:
Fixes a warning in the console related to kebab-case of CSS definition.


sBouzols (approved):
image

PR #1597, opened 2023-09-27
Dynamic Simulation: add scrollbar to curve selection dialog

thangqp:
**Bug found by Charly during the emotion migration** => when expand the variable tree box, a partial content at the bottom is hidden. Note that the bug is only produced if reducing the size of the global window, cf demo

curve_before_hidden_when_expand_model

**Solution**: add a scrollbar to the dialogue, see demo after fix

curve_after_scrollbar_

**Bonus**: In this PR, switch to CustomAGGrid for some tables in the curve configuration..


EstherDarkish (approved):
Code OK Test OK

PR #1600, opened 2023-09-28
fix: Enable ratio and phase tap changers 

achour94:
fix: Enable ratio and phase tap changers when modifications has been done and the node was unbuilt


src/components/dialogs/network-modifications/two-windings-transformer/modification/two-windings-transformer-modification-dialog.js
```diff
@@ -109,6 +109,7 @@ import {
     fetchVoltageLevelsListInfos,
 } from '../../../../../services/study/network';
 import { FetchStatus } from '../../../../../services/utils';
+import _ from 'lodash';
```
EstherDarkish, line 112: That's the first time we use lodash in gridsuite, maybe use something else ?

src/components/dialogs/network-modifications/two-windings-transformer/modification/two-windings-transformer-modification-dialog.js
```diff
@@ -226,7 +227,8 @@ const TwoWindingsTransformerModificationDialog = ({
                 ...getRatioTapChangerFormData({
                     enabled:
                         twt?.[RATIO_TAP_CHANGER]?.[ENABLED]?.value ??
-                        !!twtToModify?.ratioTapChanger,
+                        (!_.isEmpty(twt?.[RATIO_TAP_CHANGER]) ||
+                            !!twtToModify?.ratioTapChanger),
```
EstherDarkish, line 231: These three lines of code are unreadeable/unmaintainable. Could you do the logic somewhere else, with an easy to read and understand algorithm ?

EstherDarkish (changes requested):

src/components/dialogs/network-modifications/two-windings-transformer/modification/two-windings-transformer-modification-dialog.js
achour94: done

src/components/dialogs/network-modifications/two-windings-transformer/modification/two-windings-transformer-modification-dialog.js
achour94: done

EstherDarkish (approved):

src/components/dialogs/network-modifications/two-windings-transformer/modification/two-windings-transformer-modification-dialog.js
```diff
@@ -185,6 +185,36 @@ const TwoWindingsTransformerModificationDialog = ({
         }
     };
 
+    const isRatioTapChangerEnabled = useCallback(
+        (twtEditData) => {
+            const ratioTapEnabledInEditData =
```
Meklo, line 190: Is it possible to remove this one and test only `ratioTapFormHasBeenEdited || ratioTapEnabledInTwtToModify` ?
I don't see how ratioTapEnabledInEditData would give a different result than ratioTapFormHasBeenEdited since ratioTapFormHasBeenEdited is more broad

achour94: we can have the enabled value set to false then we don't have to check if there's modifications or not in the form

Meklo (approved):

PR #1605, opened 2023-09-29
Disable "Restore nodes" option if there is no node to restore.

AAJELLAL:


src/components/network-modification-tree-pane.js
```diff
@@ -349,6 +361,18 @@ export const NetworkModificationTreePane = ({
...
             }
+            if (
+                studyUpdatedForce.eventData.headers['updateType'] ===
+                    UpdateType.NODE_CREATED ||
+                UpdateType.SUBTREE_CREATED ||
+                UpdateType.NODE_MOVED ||
+                UpdateType.SUBTREE_MOVED ||
+                UpdateType.NODE_DELETED
+            ) {
+                fetchStashedNodes(studyUuid).then((res) => {
```
FranckLecuyer, line 372: Fetching stashed nodes should be done only if updateType is UpdateType.NODE_DELETED, no ?

FranckLecuyer (changes requested):
Tests: OK
Code review: just one remark

src/components/network-modification-tree-pane.js
AAJELLAL: > Fetching stashed nodes should be done only if updateType is UpdateType.NODE_DELETED, no ?

Yes you are right but fetching stashed node should be done if updateType is UpdateType.NODE_DELETED and UpdateType.NODE_CREATED. 
considere the falowing scenario:
You have only one node to restore and when you do it, you need to fetch stashed nodes otherwise the restoring option will be unabled but with no node to restore.

FranckLecuyer (approved):

souissimai (approved):

PR #1606, opened 2023-09-29
add  SA result refresh after starting analysis

anistouri:


src/components/results/sensitivity-analysis/paged-sensitivity-analysis-result.js
```diff
@@ -39,6 +42,9 @@ const PagedSensitivityAnalysisResult = ({
     const [count, setCount] = useState(0);
     const [result, setResult] = useState(null);
     const [isLoading, setIsLoading] = useState(false);
+    const sensiStatus = useSelector(
+        (state) => state.computingStatus[ComputingType.SENSITIVITY_ANALYSIS]
```
ghelloumi, line 46: you can simply destruct 'computingStatus
``const sensiStatus = useSelector(
({computingStatus}) => computingStatus[ComputingType.SENSITIVITY_ANALYSIS]
);``

src/components/results/sensitivity-analysis/paged-sensitivity-analysis-result.js
```diff
@@ -146,8 +152,13 @@ const PagedSensitivityAnalysisResult = ({
     ]);
 
     useEffect(() => {
-        fetchResult();
-    }, [fetchResult]);
+        if (sensiStatus === RunningStatus.RUNNING) {
```
AAJELLAL, line 155: Don't we need to handle FAILED status?

src/components/results/sensitivity-analysis/paged-sensitivity-analysis-result.js
AAJELLAL: > you can simply destruct 'computingStatus `const sensiStatus = useSelector( ({computingStatus}) => computingStatus[ComputingType.SENSITIVITY_ANALYSIS] );`

Personally I prefer the 1st way tin order o be consistent with the rest of the application.

anistouri: I prefer that it be consistent with what we currently have

src/components/results/sensitivity-analysis/paged-sensitivity-analysis-result.js
anistouri: We don't need to check this status because we have to go through the RUNNING status anyway, and we won't get any results when we are in FAILED

ghelloumi (approved):

PR #1609, opened 2023-10-02
change constraint on ipmax of voltage level

ghazwarhili:


AAJELLAL (approved):
Tests and code are OK.

souissimai (approved):

src/components/spreadsheet/utils/equipment-table-editors.js
```diff
@@ -100,7 +100,7 @@ export const NumericalField = forwardRef(
 
         const isValid = useCallback(
             (val, minVal, maxVal) => {
-                if (!val) {
+                if (val === undefined || val === null) {
```
anistouri, line 103: You can do that 
`if (val == null) {`
This will check for both undefined and null without needing two distinct checks

PR #1618, opened 2023-10-03
Remove optionInfos value control on showVoltageLevelDiagram

sBouzols:
Must update commons-ui version to get the fix.


Walid-Sahnoun (approved):
code and test ok

PR #1619, opened 2023-10-03
Switch to the tab with error when validating a modificattion dialog

YenguiSeddik:


src/components/dialogs/commons/modificationDialog.js
```diff
@@ -46,9 +46,24 @@ const ModificationDialog = ({
...
     };
 
-    const handleValidationError = (errors) =>
+    const handleValidationError = (errors) => {
         onValidationError && onValidationError(errors);
 
+        // When scrolling to the field with focus, you can end up with the label not completely displayed
+        // We ensure that field with focus is displayed in the middle of the dialog
+        // Delay focusing to ensure it happens after the validation. without timout, document.activeElement will return the validation button.
+        setTimeout(() => {
```
trikimahdi, line 55: you should use the setTimeout function in a useCallback or a useEffect. Don't forget the return to clear the timeout

src/components/dialogs/commons/modificationDialog.js
```diff
@@ -46,9 +46,24 @@ const ModificationDialog = ({
...
+            const focusedElement = document.activeElement;
```
trikimahdi, line 56: you should use a useRef instead of document

trikimahdi (changes requested):

src/components/dialogs/commons/modificationDialog.js
anistouri: I agree, it's better to use useRef to pick a specific element and avoid problems that document.activeElement can sometimes cause

anistouri (approved):

PR #1620, opened 2023-10-03
Externalize notification logic in custom hooks

trikimahdi:


src/components/app-top-bar.js
```diff
@@ -60,15 +46,15 @@ import {
...
-import { ComputingType } from './computing-status/computing-type';
-import { RunningStatus } from './utils/running-status';
 import { fetchNetworkElementInfos } from '../services/study/network';
 import {
     EQUIPMENT_INFOS_TYPES,
     EQUIPMENT_TYPES,
 } from './utils/equipment-types';
 import { fetchAppsAndUrls } from '../services/utils';
 import { RunButtonContainer } from './run-button-container';
+import { useResultNotificationCount } from '../hooks/use-result-notification-count';
```
ghelloumi, line 56: I think that it can be a return it useLaunchNotification hook

src/hooks/use-launch-notification.ts
```diff
@@ -0,0 +1,157 @@
...
+    const dynamicSimulationStatus = useSelector(
+        (state: ReduxState) =>
+            state.computingStatus[ComputingType.DYNAMIC_SIMULATION]
+    );
+
+    const voltageInitStatus = useSelector(
+        (state: ReduxState) => state.computingStatus[ComputingType.VOLTAGE_INIT]
+    );
+
+    useEffect(() => {
```
ghelloumi, line 71: you can extract logic in one function (reuse)

Meklo: I agree you could factorize the logic and use only one useeffect

ghelloumi (approved):

src/hooks/use-launch-notification.ts
```diff
@@ -0,0 +1,118 @@
...
+        if (isNodeBuilt(currentNode)) {
```
Meklo, line 72: You should also handle the case where the node is not built is order to reset all the notifications

src/hooks/use-launch-notification.ts
```diff
@@ -0,0 +1,118 @@
+/**
```
Meklo, line 1: I reckon use-computation-notification would be more appropriate to refer to loadflow, security analysis etc...

src/hooks/use-result-notification-count.ts
```diff
@@ -0,0 +1,44 @@
+/**
```
Meklo, line 1: And here it would be use-result-notification-count

PR #1624, opened 2023-10-04
Fixes the double call of reports when clicking in the LOGS tab.

EstherDarkish:


Tristan-WorkGH (approved):

src/components/report-viewer-tab.js
```diff
@@ -114,44 +114,20 @@ export const ReportViewerTab = ({
...
     );
 
+    // This useEffect is responsible for updating the reports when the user goes to the LOGS tab
+    // and when the application receives a notification.
     useEffect(() => {
+        // Visible and !disabled ensure that the user has the LOGS tab open and the current node is built.
         if (visible && !disabled) {
             fetchAndProcessReport(studyId, currentNode);
         }
+        // It is important to keep the notifications in the useEffect's dependencies (even if it is not
```
jonenst, line 124: why

EstherDarkish: Because without those dependencies, the notifications do not trigger a new fetch of the reports. If a user has the logs open and another user updates something (with a loadflow as an example), the first user should have his reports up to date.

I updated the comment to explain why. Is this updated comment clear enough?

anistouri (approved):

PR #1626, opened 2023-10-05
loading the results pages don't block the whole page

AbdelHedhili:


src/components/results/loadflow/load-flow-result.tsx
```diff
@@ -192,23 +234,35 @@ export const LoadFlowResult: FunctionComponent<LoadflowResultProps> = ({
...
-                rowData={rowsToShow}
-                defaultColDef={defaultColDef}
-                enableCellTextSelection={true}
-                columnDefs={loadFlowCurrentViolationsColumns}
-                onGridReady={onGridReady}
-                getRowStyle={getRowStyle}
-                overlayNoRowsTemplate={message}
-            />
+            <>
+                <Box sx={{ height: '4px' }}>
```
trikimahdi, line 253: I think you should include  the box in the condition because after the load, a little space is created between the tab and the table: {  openLoaderCurrentTab &&  <Box sx={{ height: '4px' }}> ...

trikimahdi (changes requested):

src/components/results/loadflow/load-flow-result.tsx
AbdelHedhili: the space is always here on purpose, this way when the loader appear the table doesn't "jump" down. It was requested this way at least.

trikimahdi: ok so nothing to do

trikimahdi (approved):
Code ok
Tests ok

ayolab (approved):

src/components/results/loadflow/load-flow-result.tsx
```diff
@@ -217,45 +271,64 @@ export const LoadFlowResult: FunctionComponent<LoadflowResultProps> = ({
...
         );
     };
 
     const renderLoadFlowResult = () => {
         const message = getNoRowsMessage(
             messages,
             result?.componentResults,
-            loadFlowStatus
+            loadFlowStatus,
+            result?.componentResults ? true : false
```
ayolab, line 313:
```suggestion
            !!result?.componentResults
```

ayolab (approved):

src/components/results/loadflow/load-flow-result.tsx
AbdelHedhili: yeah, I've seen this suggestion by the IDE but for me it's not so clear at first glance, I'd rather make it a few characters longer but human-readable

PR #1635, opened 2023-10-11
Prevent undefined value set to clipboard on cleanup

Meklo:
The clipboard cleanup callback may be called multiple times because of a useEffect, and previously, undefined was set to the copiedModifications array if the cleanup method was called although the array was already empty. copiedModifications is not supposed to be undefined otherwise it can cause issues


ayolab (approved):

sBouzols (approved):
nice catch

PR #1636, opened 2023-10-11
Fix: Do not launch reindex-all by notification

sBouzols:
But invite the user to refresh its web page to avoid re-indexation during index cleaning.


Meklo:
The warning snackbar is triggered numerous times, while this makes the point across to the user but it's a bit distracting
I'm not sure if it's the right PR to do the following move but we can prevent duplicate snackbars by enabling preventDuplicate in the SnackbarProvider and also adding the same option to the snackWarning call

sBouzols:
> The warning snackbar is triggered numerous times, while this makes the point across to the user it's a bit distracting I'm not sure if it's the right PR to do the following move but we can prevent duplicate snackbars by enabling preventDuplicate in the SnackbarProvider and also adding the same option to the snackWarning call

The reason of this numerous triggering was a bug in study-server. It's fixed in the other PR now.

Meklo (approved):

jonenst (approved):

PR #1638, opened 2023-10-11
Add pagination filtering and multi sorting for one bus SCA

flomillot:
Add the pagination and filtering on one bus SCA.
Rename some interfaces.
Some small fixes and code cleaning.


src/components/results/shortcircuit/shortcircuit-analysis-result-table.tsx
```diff
@@ -235,24 +226,36 @@ const ShortCircuitAnalysisResultTable: FunctionComponent<
...
         return current;
     };
 
-    const flattenResult = (shortCircuitAnalysisResult: SCAResultFault[]) => {
+    // When we filter / sort the 'current' column in one bus, it's actually the 'fortescueCurrent.positiveMagnitude' field in the back-end
+    const fromColumnToField = useCallback(
+        (column: string) => {
+            if (
+                analysisType === ShortCircuitAnalysisType.ONE_BUS &&
+                column === 'current'
```
sBouzols, line 244: Why not rename this column whit the correct name as in the back and remove this transcription method which is useless for all other columns which have the correct name ?

src/components/results/shortcircuit/shortcircuit-analysis-result-table.tsx
```diff
@@ -331,6 +335,68 @@ const ShortCircuitAnalysisResultTable: FunctionComponent<
         }
     }, []);
 
+    const onFilterChanged = useCallback(
+        (e: FilterChangedEvent) => {
+            // to see what contains filter model : https://www.ag-grid.com/javascript-data-grid/filter-api/
+            const formattedFilter = Object.entries(e.api.getFilterModel())
```
sBouzols, line 341: I understand this code, but it's a pity that we don't use the raw filterModel format from AGGrid I think. You flatten the filters as well but it remove filters composition rules no ? How do you know which order to follow then and all?

src/components/results/shortcircuit/shortcircuit-analysis-all-buses-result.tsx
```diff
@@ -0,0 +1,47 @@
...
+    );
+
+    const allBusesShortCircuitAnalysisStatus = useSelector(
+        (state: ReduxState) =>
+            state.computingStatus[ComputingType.ALL_BUSES_SHORTCIRCUIT_ANALYSIS]
+    );
+
+    const [result, setResult] = useState<SCAFaultResult[]>([]);
+
+    const updateResult = useCallback(
```
sBouzols, line 31: You don't have a useEffect which fetchShortCircuitAnalysisResult here ? for ALL_BUSES, mode NONE and all you know ?

src/components/directory-item-selector.js
```diff
@@ -26,11 +27,6 @@ const styles = {
     }),
 };
 
-// this method checks if an object (as map here) is empty or has at least one property
-function isObjectEmpty(obj) {
-    return obj == null || Object.keys(obj).length === 0;
-}
-
```
sBouzols, line 33: This is not in the scope of this PR isn't it ?

src/components/results/shortcircuit/shortcircuit-analysis-result-table.tsx
```diff
@@ -29,15 +31,14 @@ import {
...
 
 interface ShortCircuitAnalysisResultProps {
-    result: SCAResultFault[];
-    onSortChanged: (colKey: string, sortWay: number) => void;
-    sortConfig: ISortConfig;
-    analysisType: ShortcircuitAnalysisType;
+    result: SCAFaultResult[];
+    analysisType: ShortCircuitAnalysisType;
+    updateFilter: (filter: any) => void;
+    updateSort: (sort: any) => void;
```
anistouri, line 41: You can create objects for both filtering and sorting to replace the any type. This approach is clearer

src/components/results/shortcircuit/shortcircuit-analysis-result-table.tsx
flomillot: I wanted to make it a bit different of the filterModel, which can take different JSON form depending how many filters are activated by columns. Here you just send an array of filters, and the back manage it.
And yeah the drawback of this was no OR conditions was possible yet...
Anyway we won't use filterModel soon as discussed, and this will probably change...

src/components/results/shortcircuit/shortcircuit-analysis-all-buses-result.tsx
flomillot: Nope because it was not needed here, as it was done before my PR.
But as discussed maybe the best thing to do is call this endpoint for both one and all buses.

src/components/results/shortcircuit/shortcircuit-analysis-result-table.tsx
flomillot: Yes you're right I will add it

src/components/results/shortcircuit/shortcircuit-analysis-result-table.tsx
flomillot: Actually as we decided to use only one filter by column now this code is obsolete (and will be remove soon anyway), so I will just simplify this part.

PR #1645, opened 2023-10-13
Fix line tooltip

flo-dup:
When hovering over a line, if the line has a name and use name instead of id is enabled, there's no tooltip displayed but an infinite progression bar instead.


FranckLecuyer (changes requested):
Tests: With this modification, there's no tooltip displayed but an infinite progression bar in all cases ...

FranckLecuyer (approved):

PR #1664, opened 2023-11-03
Add colors to map legend

Tristan-WorkGH:
Add colors to the main map legend, when map is in "Nominal tension" coloration mode, with each tension colored by same color use by lines on the map.


antoinebhs:
Not required for now

PR #1667, opened 2023-11-03
Separate computation logs from modification logs

dbraquart:


src/components/results/common/computation-report-viewer.tsx
```diff
@@ -0,0 +1,113 @@
...
+    const [report, setReport] = useState(undefined);
+    const { snackError } = useSnackMessage();
+    const studyUuid = useSelector((state: ReduxState) => state.studyUuid);
+    const currentNode = useSelector(
+        (state: ReduxState) => state.currentTreeNode
+    );
+
+    const makeReport = useCallback(
+        (reportData: any) => {
+            const nodeName = currentNode?.data.label;
```
EstherDarkish, line 38:
```suggestion
            const nodeName = currentNode?.data?.label;
```

src/components/results/sensitivity-analysis/sensitivity-analysis-result-tab.js
```diff
@@ -69,23 +77,39 @@ const SensitivityAnalysisResultTab = ({ studyUuid, nodeUuid }) => {
...
+                        updateFilter={updateFilter}
+                        filterSelector={filterSelector}
+                        onSortChanged={onSortChanged}
+                        sortConfig={sortConfig}
+                        page={page}
+                        setPage={setPage}
+                    />
+                </>
+            )}
+            {sensiKindIndex === 3 &&
```
EstherDarkish, line 104: This whole "sensiKindIndex" doesn't work to track the tabIndex.
Maybe we should rename that to "tabIndex", and use the FUNCTION_TYPES instead of sensiKindIndex in the PagedSensitivityAnalysisResult component ?

src/components/report-viewer-tab.js
```diff
@@ -189,8 +197,10 @@ export const ReportViewerTab = ({
                                 inputProps={{
                                     'aria-label': 'primary checkbox',
                                 }}
-                                onChange={(e) => handleChangeValue(e)}
-                                disabled={disabled}
+                                onChange={(e) => handleChangeNodeOnlySwitch(e)}
+                                disabled={
+                                    disabled || rootNodeId === currentNode?.id
```
EstherDarkish, line 202: If we disable the switch on the root node, maybe we also should force the "checked" value to either true or false if the current node is root, because at the moment, it is possible to manipulate the `parent-nodes-report?nodeOnlyReport=` parameter in the REST query by setting the switch to true or false in another node.

That should not be possible, and the results seem to always be the same, but that should not be allowed anyway.

src/components/report-viewer-tab.js
```diff
@@ -63,7 +63,7 @@ export const ReportViewerTab = ({
     const { snackError } = useSnackMessage();
     const [nodeOnlyReport, setNodeOnlyReport] = useState(true);
 
-    const handleChangeValue = useCallback((event) => {
+    const handleChangeNodeOnlySwitch = useCallback((event) => {
```
EstherDarkish, line 66: nice :)

src/components/results/common/computation-report-viewer.tsx
```diff
@@ -0,0 +1,113 @@
...
+        <>
+            <Box sx={{ height: '4px' }}></Box>
+            {report && (
+                <ReportViewer
+                    jsonReportTree={report}
+                    subReportPromise={subReportPromise}
+                    nodeReportPromise={nodeReportPromise}
+                />
+            )}
+            ;
```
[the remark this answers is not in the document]
dbraquart: done

src/components/report-viewer-tab.js
dbraquart: I think it is not usefull

src/components/results/sensitivity-analysis/sensitivity-analysis-result-tab.js
dbraquart: sensiKindIndex is used as a tabIndex and works.
I prefer not to change existing code in this PR, at this time.

src/components/results/shortcircuit/shortcircuit-analysis-result-tab.tsx
```diff
@@ -55,11 +78,40 @@ export const ShortCircuitAnalysisResultTab: FunctionComponent<
...
-            )}
+
+            <Tabs value={resultOrLogIndex} onChange={handleSubTabChange}>
+                <Tab label={<FormattedMessage id={'Results'} />} />
+                <Tab
+                    label={<FormattedMessage id={'ComputationResultsLogs'} />}
+                />
+            </Tabs>
+
+            {resultOrLogIndex === 0 &&
```
SlimaneAmar, line 89: Simplification the subtab handliing ?

SlimaneAmar (approved):

PR #1669, opened 2023-11-06
Add sections to shunt compensator modification

ayolab:


Meklo:
I think we ought to remove Q_AT_NOMINAL_V and SUSCEPTANCE_PER_SECTION occurences from field-constants.js and characteristics-form-utils.js they are no longer necessarry
Similarly in characteristics-form-utils.js `getCharacteristicsFormDataFromSearchCopy` is not in use anymore too

Meklo (approved):

PR #1673, opened 2023-11-07
Fix: Avoid calling fetchStashedNodes for all notification update types

sBouzols:
Parenthesis were missing in the condition, then it was true for all types of notifications.
Moved the call in the corresponding types but then it could have hide a updatetype where it is necessary to call fetchStashedNodes.
We have to test it well !


archendel (approved):
Code: OK

AAJELLAL (approved):
Code OK and tests also.

PR #1681, opened 2023-11-08
Edit fields spreadsheet generators

souissimai:


src/components/spreadsheet/equipment-table.js
```diff
@@ -29,8 +35,74 @@ export const EquipmentTable = ({
...
+    const handleSavePopupGeneratorEditRegulatingTerminalFields = (
+        voltageRegulationGenerator
+    ) => {
+        setGeneratorId(clickedCellData.data.id);
+
+        const isDistantRegulation =
+            voltageRegulationGenerator.voltageRegulationType ===
+            REGULATION_TYPES.DISTANT.id;
+
+        modifyGenerator(
```
ghelloumi, line 60: I think it is a good opportunity to modify this function to make it receive an object instead of all those params

src/components/dialogs/network-modifications/generator/modification/regulating-terminal-modification-dialog.js
```diff
@@ -0,0 +1,172 @@
+import { SelectInput } from '@gridsuite/commons-ui';
```
ghelloumi, line 1: add RTE licence

src/components/dialogs/network-modifications/generator/modification/regulating-terminal-modification-dialog.js
```diff
@@ -0,0 +1,172 @@
...
+const RegulatingTerminalModificationDialog = ({
+    data,
+    currentNode,
+    studyUuid,
+    isUpdate,
+    editDataFetchStatus,
+    onModifyRegulatingTeminalGenerator,
+    ...dialogProps
+}) => {
+    const emptyFormData = useMemo(
```
ghelloumi, line 30: emptyFormData is a simple object, and its computation is not particularly complex. 
Therefore, the use of useMemo for emptyFormData might not provide a significant performance benefit.
as a good practice, useMemo is only used for heavy CPU using code blocks

src/components/dialogs/network-modifications/generator/modification/regulating-terminal-modification-dialog.js
```diff
@@ -0,0 +1,172 @@
...
+            [VOLTAGE_LEVEL]: null,
+            [EQUIPMENT]: null,
+        }),
+        []
+    );
+
+    const formSchema = yup
+        .object()
+        .shape({
+            [VOLTAGE_REGULATION_TYPE]: yup.string().nullable().required(),
```
ghelloumi, line 42: in this case, the [VOLTAGE_REGULATION_TYPE] field can either have a string value or be null, but it cannot be any other type..
if it is not the required validation, you have to remove 'nullable()'

src/components/dialogs/network-modifications/generator/modification/regulating-terminal-modification-dialog.js
```diff
@@ -0,0 +1,172 @@
...
+        (voltageRegulationGenerator) => {
+            onModifyRegulatingTeminalGenerator(voltageRegulationGenerator);
+        },
+        [onModifyRegulatingTeminalGenerator]
+    );
+    const intl = useIntl();
+
+    const regulatingTerminalField = (
+        <RegulatingTerminalForm
+            id={''}
```
ghelloumi, line 104: you have to set an id

src/components/dialogs/network-modifications/generator/modification/regulating-terminal-modification-dialog.js
```diff
@@ -0,0 +1,172 @@
...
+import { getTapChangerEquipmentSectionTypeValue } from 'components/utils/utils';
+import { fetchVoltageLevelsListInfos } from 'services/study/network';
+
+const RegulatingTerminalModificationDialog = ({
+    data,
+    currentNode,
+    studyUuid,
+    isUpdate,
+    editDataFetchStatus,
+    onModifyRegulatingTeminalGenerator,
```
ghelloumi, line 27:
```suggestion
    onModifyRegulatingTerminalGenerator,
```

src/components/spreadsheet/equipment-table.js
```diff
@@ -101,38 +173,68 @@ export const EquipmentTable = ({
...
-            onColumnMoved={handleColumnDrag}
-            suppressDragLeaveHidesColumns={true}
-            suppressColumnVirtualisation={true}
-            suppressClickEdit={true}
-            context={gridContext}
-            onGridReady={handleGridReady}
-            shouldHidePinnedHeaderRightBorder={
-                shouldHidePinnedHeaderRightBorder
+    const handleOnClickOnCell = (params) => {
+        if (onCellClicked != null) {
```
ghelloumi, line 177: why not strict equal?
otherwise I think you can use !!onCellClicked

src/components/spreadsheet/utils/config-tables.js
```diff
@@ -57,6 +62,33 @@ const propertiesGetter = (params) => {
     }
 };
 
+const RegulatingTerminalCellRenderer = (params) => {
```
ghelloumi, line 65:
```suggestion
const RegulatingTerminalCellRenderer = (params) => {
    const regulatingTerminalConnectableId =
        params.data?.regulatingTerminalConnectableId;
    if (
        params.data?.regulatingTerminalVlId ||
        regulatingTerminalConnectableId
    ) {
        return `${params.data.regulatingTerminalConnectableType} (${regulatingTerminalConnectableId} )`;
    } else {
        return null;
    }
};

```

ghelloumi: create a const for params.data.regulatingTerminalConnectableId as it used two times
using Template literals is more readable in this case

src/components/spreadsheet/utils/config-tables.js
```diff
@@ -95,6 +127,17 @@ const generateEditableNumericColumnDefinition = (
     };
 };
 
+const onCellGeneratorCellClicked = (event, callBack) => {
```
ghelloumi, line 130: why switch?
```suggestion
const onCellGeneratorCellClicked = (event, callBack) => {
    const { colDef, data } = event;
    if ([ 'RegulationTypeText', 'RegulatingTerminalGenerator'].includes(colDef.id)) {
        callBack(data);
    }
};
```

src/components/spreadsheet/equipment-table.js
souissimai: we should create a TS to change this function for all equipments

src/components/dialogs/network-modifications/generator/modification/regulating-terminal-modification-dialog.js
```diff
@@ -0,0 +1,172 @@
...
+        );
+    }, [previousRegulationType, voltageRegulationType]);
+
+    const onSubmit = useCallback(
+        (voltageRegulationGenerator) => {
+            onModifyRegulatingTeminalGenerator(voltageRegulationGenerator);
+        },
+        [onModifyRegulatingTeminalGenerator]
+    );
+    const intl = useIntl();
```
AAJELLAL, line 100: Move this to the top.

src/components/dialogs/network-modifications/generator/modification/regulating-terminal-modification-dialog.js
AAJELLAL, line 104: Why the id is empty?

src/components/dialogs/network-modifications/generator/modification/regulating-terminal-modification-dialog.js
```diff
@@ -0,0 +1,172 @@
...
+                data
+            )}
+        />
+    );
+
+    const clear = useCallback(() => {
+        reset(emptyFormData);
+    }, [emptyFormData, reset]);
+
+    const getVoltageRegulationTypeLabel = () => {
```
AAJELLAL, line 120: It would be good to use **useMemo** here

src/components/spreadsheet/equipment-table.js
```diff
@@ -29,8 +35,74 @@ export const EquipmentTable = ({
...
+    const [clickedCellData, setClickedCellData] = React.useState({});
+    const [generatorId, setGeneratorId] = React.useState('');
+    const [
+        popupGeneratorEditRegulatingTerminalFields,
+        setPopupGeneratorEditRegulatingTerminalFields,
+    ] = React.useState(false);
+    const onRegulatingTerminalPopupClose = () => {
+        setPopupGeneratorEditRegulatingTerminalFields(false);
+    };
+    const handleSavePopupGeneratorEditRegulatingTerminalFields = (
```
AAJELLAL, line 51: The same here For me the name is too long and confusing

src/components/spreadsheet/equipment-table.js
```diff
@@ -29,8 +35,74 @@ export const EquipmentTable = ({
...
     network,
     shouldHidePinnedHeaderRightBorder,
+    onCellClicked,
 }) => {
     const theme = useTheme();
+    const { snackError } = useSnackMessage();
+    const [clickedCellData, setClickedCellData] = React.useState({});
+    const [generatorId, setGeneratorId] = React.useState('');
+    const [
+        popupGeneratorEditRegulatingTerminalFields,
```
AAJELLAL, line 45: popupGeneratorEditRegulatingTerminalFields is too long as name and I don't think we need all this details.
We only have one dialogue so you can use something like this : **openDialog**

src/components/spreadsheet/equipment-table.js
```diff
@@ -29,8 +35,74 @@ export const EquipmentTable = ({
...
+            undefined
+        ).catch((error) => {
+            snackError({
+                messageTxt: error.message,
+                headerId: 'GeneratorModificationError',
+            });
+        });
+    };
+
     const intl = useIntl();
```
AAJELLAL, line 106: Move this to the top.

src/components/spreadsheet/equipment-table.js
```diff
@@ -29,8 +35,74 @@ export const EquipmentTable = ({
...
+        voltageRegulationGenerator
+    ) => {
+        setGeneratorId(clickedCellData.data.id);
+
+        const isDistantRegulation =
+            voltageRegulationGenerator.voltageRegulationType ===
+            REGULATION_TYPES.DISTANT.id;
+
+        modifyGenerator(
+            studyUuid,
```
AAJELLAL, line 61: **modifyGenerator** must absolutely be changed.
To check with the techleads if it should be done now or later?

AAJELLAL (changes requested):

ghelloumi (approved):

src/components/spreadsheet/utils/config-tables.js
```diff
@@ -56,6 +61,35 @@ const propertiesGetter = (params) => {
         return null;
     }
 };
+const handleCellClickEvent = (event) => {
```
jamal-khey, line 64: it better to give another name to this function, because we may register more callback for other cells.

AAJELLAL (approved):

src/components/spreadsheet/equipment-table.js
```diff
@@ -64,14 +131,24 @@ export const EquipmentTable = ({
...
             isEditing: topPinnedData ? true : false,
+            handleCellClick: {
+                //functions for handling cell click
+                openGeneratorDialog: () => {
+                    openGeneratorPopup();
+                },
+            },
         };
     }, [network, topPinnedData]);
+    const cellClickedInitialData = gridContext.dynamicValidation;
```
jamal-khey, line 151: you probablly dont need to create this variable, you can pass gridContext.dynamicValidation, to  the pop-up

AAJELLAL (changes requested):
Here are some bugs found while testing:
            **-**  "" is accepted as name's value. I think empty string shouldn't be allowed as names(to check with the PO).
           **-**  You need to add "Coefficient de participation à la compensation " column as asked in the US( unless I'm mistaken).
              **-**  when I modify all the fields only those which are modified through the popup are taken into account.To check with the PO if it's OK but It's weird as behavior.

PR #1695, opened 2023-11-14
Formatte numeric value with two digits

YenguiSeddik:


src/components/results/loadflow/load-flow-result-utils.ts
```diff
@@ -83,23 +83,19 @@ export const loadFlowCurrentViolationsColumnsDefinition = (
         {
             headerName: intl.formatMessage({ id: 'CurrentViolationLimit' }),
             field: 'limit',
-            valueFormatter: (params: ValueFormatterParams) =>
```
ghelloumi, line 86: I think that `numeric: true` is necessary here, isn't it ?

souissimai (approved):

souissimai:
fix confclicts so that I can approve , code ok and test seems ok

souissimai (approved):

ghelloumi (approved):

PR #1709, opened 2023-11-17
Fix open reac params: switch order.

AAJELLAL:
Following the PO's wish, it is necessary to reverse the order of the title and the info message.


ayolab (approved):

Tristan-WorkGH (commented):
approved

PR #1714, opened 2023-11-20
fix controls not done on batterie spreadsheet

jamal-khey:
allow null spetpoint in groupe modification
add min/max controls in the spreadseet for batteries


etiennehomer (approved):
Code and tests ok

PR #1726, opened 2023-11-23
fix when activePowerOn value not modified in spreadsheet, no logs to show

souissimai:


jamal-khey (approved):
review and test Ok , the logs are not displayed

src/components/spreadsheet/utils/config-tables.js
```diff
@@ -1682,7 +1682,7 @@ export const TABLES_DEFINITIONS = {
                     return {
                         defaultValue:
                             params.data?.activePowerControl
-                                ?.activePowerControlOn | 0,
+                                ?.activePowerControlOn,
```
achour94, line 1685: You should set the value to an empty string instead of undefined to avoid receiving a warning in the console
```suggestion
                                ?.activePowerControlOn || '',
```

achour94 (changes requested):

src/components/spreadsheet/utils/config-tables.js
souissimai: If we include it, the default value won't be displayed when editing if it's set to "false."

achour94 (approved):

achour94 (approved):

jamal-khey (approved):
seems Ok to me

PR #1727, opened 2023-11-23
ignore regulationValue when useless

Mathieu-Deharbe:
The issue is the fact that useless 'regulationValue' are received as NaN from the back and are not handled correctly by the front. 

Now if those values are useless (because of the regulation mode) they will simply be ignored


src/components/dialogs/network-modifications/two-windings-transformer/creation/two-windings-transformer-creation-dialog.js
```diff
@@ -444,9 +444,15 @@ const TwoWindingsTransformerCreationDialog = ({
                         twt?.[PHASE_TAP_CHANGER]
                     ),
                     currentLimiterRegulatingValue:
```
TheMaskedTurtle, line 446: The method getPhaseTapChangerFormData is also called line 266, maybe you should do the change there too ?

src/components/dialogs/network-modifications/two-windings-transformer/creation/two-windings-transformer-creation-dialog.js
```diff
@@ -444,9 +444,15 @@ const TwoWindingsTransformerCreationDialog = ({
...
-                        twt?.[PHASE_TAP_CHANGER]?.regulationValue,
+                        twt?.[PHASE_TAP_CHANGER]?.[REGULATION_MODE] ===
+                        PHASE_REGULATION_MODES.CURRENT_LIMITER.id
+                            ? twt?.[PHASE_TAP_CHANGER]?.regulationValue
+                            : undefined,
```
TheMaskedTurtle, line 450: Maybe null instead of undefined as it is the default values in the method definition

Mathieu-Deharbe: I wondered this too. Sometimes undefined is used (mostly for form default values), sometimes null is (especially for structures default values). I finally picked 'undefined' because that was the case in two-windings-transformer-modification-dialog but personnaly I prefer null.

Maybe a techlead should decide.

src/components/dialogs/network-modifications/two-windings-transformer/creation/two-windings-transformer-creation-dialog.js
Mathieu-Deharbe: Indeed. Even if it is much less likely to happen here, it will be cleaner. Done.

YenguiSeddik (approved):

TheMaskedTurtle (approved):

PR #1738, opened 2023-11-27
Hide CGMES parameter in export case

anistouri:


src/components/utils/constants.ts
```diff
@@ -22,3 +22,4 @@ export enum PREDEFINED_PARAMETERS {
     ICC_MAX_WITH_NOMINAL_VOLTAGE_MAP = 'ICC_MAX_WITH_NOMINAL_VOLTAGE_MAP',
     ICC_MAX_WITH_CEI909 = 'ICC_MAX_WITH_CEI909',
 }
+export const CGMES = 'CGMES';
```
flomillot, line 25: Do we really need to create a const to use it just one time on a temporary code ?

flomillot (approved):

ayolab (approved):

src/components/utils/constants.ts
anistouri: To avoid hardcoding the value in the code, which is a bit ugly

PR #1744, opened 2023-11-28
No longer truncate functional logs

jamal-khey:
remove the limit to 500 items in the logs treeview


Tristan-WorkGH (approved):

jamal-khey:
> maxSubReports is not used anymore and is the only dependency of this callback function. Maybe we should change remove it and clean the code ?

@SlimaneAmar  asked me to keeo the warning log. that's why we still using this value

PR #1746, opened 2023-11-29
Shortcircuit analysis results custom aggrid header + enhancing custom agGrid header solution

ghelloumi:


src/components/results/shortcircuit/shortcircuit-analysis-result-table.tsx
```diff
@@ -84,124 +83,149 @@ const ShortCircuitAnalysisResultTable: FunctionComponent<
...
-                filter:
+                sortProps:
                     analysisType === ShortCircuitAnalysisType.ALL_BUSES
-                        ? 'agTextColumnFilter'
-                        : null,
-                filterParams: textFilterParams,
-            },
-            {
+                        ? sortProps
+                        : undefined,
```
anistouri, line 102: You can create two constants, for example, sortPropsChecked and filterPropsChecked, to avoid having to perform the verification for each column

src/components/results/shortcircuit/shortcircuit-analysis-result-table.tsx
```diff
@@ -341,53 +266,103 @@ const ShortCircuitAnalysisResultTable: FunctionComponent<
...
+                    const fault = faultResult.fault;
+                    const limitViolations = faultResult.limitViolations ?? [];
+                    let firstLimitViolation;
+                    if (limitViolations.length > 0) {
+                        let lv = limitViolations[0];
+                        firstLimitViolation = {
+                            limitType: intl.formatMessage({
+                                id: lv.limitType,
+                            }),
+                            limitName: lv.limitName,
```
anistouri, line 297: Did we need limitName? If so, we need to check with POs if we should add the formatter as in AS

src/hooks/use-aggrid-row-filter.ts
```diff
@@ -0,0 +1,136 @@
...
+
+const changeValueFromArrayWithFieldValue = (
+    filtersArrayToModify: FilterType[],
+    field: string,
+    newData: FilterDataType
+) => {
+    const filterIndex = filtersArrayToModify.findIndex(
+        (f: FilterType) => f.field === field
+    );
+    if (filterIndex === -1) {
```
anistouri, line 55: you can use SORT_WAYS.desc

ghelloumi: it is not related to sort.. It is juste the index of a member that not exist in a array
(if the member with filed === field do not exist in filtersArrayToModify => findIndex will return -1)

anistouri: My bad

src/hooks/use-aggrid-sort.ts
```diff
@@ -7,12 +7,16 @@
 
 import { useCallback, useState } from 'react';
 
-export interface ISortConfig {
+export type SortConfigType = {
     colKey: string;
-    sortWay: number;
-}
+    sortWay: string | number;
```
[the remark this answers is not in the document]
ghelloumi: no. Fixed, thank you

anistouri (approved):

YenguiSeddik (approved):

PR #1752, opened 2023-12-01
Upgrade to react-map-gl 7, using deckgl as a control [DO NOT MERGE]

jonenst:
- with react-map-gl 7, we are forced to choose between
  * Root Deck, Map as child. Deckgl handles all camera and user input like before The major downside is that react-map-gl 7 dropped support for using their controls (NavigationControl, the '+/-/compass' at the top right
  * Root Map, deck as child. React-map-gl handles all camera and user input. some advanced deckgl features are not available, and it changes what we were doing.
- right click on a feature doesn't seem to be exposed in "MapRoot/DeckglChild" mode for now, switch to shiftclick as a stopgap measure
- remove project_size_at_latitude copy, it seems accessible now? Removing it fixes the following error and the arrows are still correctly displayed: luma.gl: GLSL compilation errors in vertex shader arrow-layer-vertex-shader-1 658:  * Just to call it from project_size_all_zoom_levels(). 659:  */ 660: float project_size_at_latitude(float lat) { ^^^ ERROR: 'project_size_at_latitude' : function already has a body deck: initialization of ArrowLayer({id: 'lineLayer-ArrowNominalVoltage400'}): GLSL compilation errors in vertex shader arrow-layer-vertex-shader-1

DO NOT MERGE: need to decide what to do with this


jonenst:
reopened as #1800 because I forgot I had done this one

PR #1753, opened 2023-12-01
maplibre

jonenst:
Allow to use maplibre + cartodb with a user map config parameter. The default is still mapbox


antoinebhs:
We should change this PR description with "using maplibre in addition to mapbox"

src/components/app-wrapper.js
```diff
@@ -106,7 +112,12 @@ let lightTheme = createTheme({
     formFiller: {
         background: '#e6e6e6',
     },
-    mapboxStyle: 'mapbox://styles/mapbox/light-v9',
+    [basemap_style_theme_key(MAP_BASEMAP_MAPBOX)]:
+        'mapbox://styles/mapbox/light-v9',
+    [basemap_style_theme_key(MAP_BASEMAP_CARTO_NOLABEL)]:
+        'https://basemaps.cartocdn.com/gl/positron-nolabels-gl-style/style.json',
```
antoinebhs, line 118: * What are the limitations of this tile provider? I see in this document https://drive.google.com/file/d/1P7bhSE-N9iegI398QYDjKeVhnbS7-Ilk/view that "Customer may use up to one million (1,000,000) tile requests each
calendar month, unless otherwise specified on an Order Form.",  I also see here https://github.com/CartoDB/basemap-styles "Limitations: free usage for up to 75,000 mapviews per month, none-commercial services only. For bigger usage and other cases contact CARTO sales for enterprise service key." 
* Are we sure we can use this in terms of IP property and related stuff? I see in their document that "If Customer is using the Basemap Services on behalf of an organization (a company, governmental agency, non-governmental organization),
Customer agrees to be bound by the Basemap Terms on behalf of itself and such organization, and represents it has the authority to agree to
these Basemap Terms on behalf of such organization. If Customer does not have such authority, Customer may not use the Basemap
Services on behalf of such organization." => Are you entitled to do that?
* Do we have an idea of our current usage? Maybe by getting the statistics of our mapbox token?
We should be self hosted at some point. Tiles for Europe look rather reasonable in terms of data size.

jonenst: this tile provider is meant to be used for none-commercial open source tests by anyone interested in the project

antoinebhs (approved):
Code OK
Tests OK
I think that in the version Carto + label, the names of the regions clutter the map.

thangqp (approved):
Code OK
Test OK

PR #1755, opened 2023-12-01
Initialize geodata from root node rather than from default selected node

klesaulnier:


EstherDarkish (approved):
Test OK Code OK

src/components/network-map-tab.js
```diff
@@ -658,12 +661,14 @@ export const NetworkMapTab = ({
...
+                loadRootNodeGeoData();
+                // set initialized to false to trigger missing geo-data fetching on next render
+                setInitialized(false);
             }
         }
-    }, [studyUuid, loadAllGeoData, loadMissingGeoData, lineFullPath]);
+    }, [studyUuid, loadRootNodeGeoData, loadMissingGeoData, lineFullPath]);
 
-    const loadMapEquipments = useCallback(() => {
+    const loadDefaultNodeMapEquipments = useCallback(() => {
```
etiennehomer, line 671: Should not be rename nope ?
This method loads the equipments of the current node

src/components/network-map-tab.js
```diff
@@ -804,31 +809,50 @@ export const NetworkMapTab = ({
...
         updateMapEquipmentsAndGeoData,
-        loadAllGeoData,
+        loadRootNodeGeoData,
+        isRootNodeDataLoaded,
         isInitialized,
         reloadMapNeeded,
         updatedSubstationsIds,
     ]);
 
+    useEffect(() => {
```
etiennehomer, line 842: Maybe we could avoid this useEffect by chaining loadRootNodeGeoData and loadMissingGeoData ? 
But we need the mapEquipments to be loaded to loadMissingGeoData...

etiennehomer: Ok to keep this hook

etiennehomer (changes requested):
There's a bug : 
in the "first built node", you create a substation, a voltage level and 2 lines attached to this new voltage level.
when you open the study, the substation and the voltage level are well displayed and placed. But the 2 new lines are not displayed. There's no request to the back to get the lines' positions

src/redux/actions.js
```diff
@@ -342,6 +342,15 @@ export function resetMapReloaded() {
     };
 }
 
+export const EQUIPMENTS_INITIALIZED = 'EQUIPMENTS_INITIALIZED';
```
etiennehomer, line 345: You could rename all those variables and functions with the map prefix. 
For example: EQUIPMENTS_INITIALIZED to MAP_EQUIPMENTS_INITIALIZED

src/components/network-map-tab.js
```diff
@@ -804,31 +813,52 @@ export const NetworkMapTab = ({
...
+        // when root node geodata are loaded, we fetch current node missing geo-data
+        // we check if equipments are done initializing because they are checked to fetch accurat missing geo data
+        if (isRootNodeDataLoaded && isEquipmentsInitiliazed && !isInitialized) {
```
etiennehomer, line 851: isInitialized could be renamed to isMissingGeoDataLoaded
and 
isRootNodeDataLoaded to isRootNodeGeoDataLoaded

klesaulnier: I agree for root node name, I will rename isInitialized to isDefaultNodeMissingGeoDataLoaded though, to be more accurate

src/redux/actions.js
klesaulnier: done

src/components/network-map-tab.js
```diff
@@ -118,6 +136,10 @@ export const NetworkMapTab = ({
 
     const reloadMapNeeded = useSelector((state) => state.reloadMap);
 
+    const isMapEquipmentsInitiliazed = useSelector(
```
EstherDarkish, line 139:
```suggestion
    const isMapEquipmentsInitialized = useSelector(
```

src/components/network-map-tab.js
```diff
@@ -80,10 +86,22 @@ export const NetworkMapTab = ({
     const mapEquipments = useSelector((state) => state.mapEquipments);
     const studyUpdatedForce = useSelector((state) => state.studyUpdated);
     const mapDataLoading = useSelector((state) => state.mapDataLoading);
+    const treeModel = useSelector(
+        (state) => state.networkModificationTreeModel
+    );
+
+    const rootNodeId = useMemo(() => {
+        const rootNode = treeModel.treeNodes.find(
```
EstherDarkish, line 94: Potentially, state.networkModificationTreeModel could be null (when we close the study), so technically that would be better.
```suggestion
        const rootNode = treeModel?.treeNodes.find(
```

src/components/network-map-tab.js
```diff
@@ -804,31 +817,60 @@ export const NetworkMapTab = ({
...
+        isRootNodeGeoDataLoaded,
+        isMapEquipmentsInitialized,
         isInitialized,
         reloadMapNeeded,
         updatedSubstationsIds,
     ]);
 
+    useEffect(() => {
+        // when root node geodata are loaded, we fetch current node missing geo-data
+        // we check if equipments are done initializing because they are checked to fetch accurat missing geo data
```
EstherDarkish, line 858:
```suggestion
        // we check if equipments are done initializing because they are checked to fetch accurate missing geo data
```

EstherDarkish (approved):
Test OK Code OK with minor adjustments.

src/components/network-map-tab.js
klesaulnier: both solutions does exist in the project
I'm fixing it the way you ask me to do

src/components/network-map-tab.js
klesaulnier: fixed

PR #1756, opened 2023-12-01
Format new voltage level in the form initialisation

achour94:
Format new voltage level in the form initialisation to avoid the isDirty at the beginning


flomillot (approved):
Code et test OK

PR #1758, opened 2023-12-04
Add button to unbuild node

klesaulnier:


src/components/graph/menus/create-node-menu.js
```diff
@@ -284,11 +290,17 @@ const CreateNodeMenu = ({
                 },
             },
         },
+        UNBUILD_NODE: {
+            onRoot: false,
+            action: () => unbuildNode(),
+            id: 'unbuildNode',
+            disabled: !isNodeUnbuildingOrRemovingAllowed(),
```
Meklo, line 297: Unbuilding the node should also be disabled if the node is already disabled imo

src/components/graph/menus/create-node-menu.js
```diff
@@ -171,7 +177,7 @@ const CreateNodeMenu = ({
         );
     }
 
-    function isNodeRemovingAllowed() {
+    function isNodeUnbuildingOrRemovingAllowed() {
         return !isAnyNodeBuilding && !mapDataLoading;
     }
```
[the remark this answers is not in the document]
klesaulnier: this was nodeBuilding**Allowed**
But I changed it, now there are 2 different methods for building and removing node

src/components/graph/menus/create-node-menu.js
klesaulnier: good point, fixed

Meklo (approved):

PR #1762, opened 2023-12-04
fix threeview take more than 100% height

jamal-khey:
this cas cause an issue when the treeview contains a lot of items, it make navigating difficult

to test this change you need to generate a lot of items in the treeview.
the easiest way to do it , is to run a security analysis


AAJELLAL (commented):
Strange behavior: after some scroll, the scrollbar is no longer visible and yet you can still scroll. To be checked with the PO if it's ok.
Reproducible on firefox and edge.

EstherDarkish (changes requested):
There is a large void below the logs

AAJELLAL (changes requested):
Now the scroll is visible but there is an extra empty space at the bottom of the page.
by scrolling to the end of the page, the extra space is gone and the last element  is not quite visible.

EstherDarkish (commented):
I've got a clue as how to fix this bug : in security-analysis-result-tab.tsx, if we exclude the <ComputationReportViewer> from it's <Box>, then the report is scrollable again (just like in the LOGS tab).

Something with this <Box> messes with the report.

EstherDarkish (approved):
Test OK code OK.
In a future US, we should try to improve the display to not have a void below the table in the LOGS tag.

AAJELLAL (approved):
Test OK and Code OK.

PR #1764, opened 2023-12-05
Fix run-button conditionnals disable

Tristan-WorkGH:
Fix conditionals on index forgotten during #1717
* **use now enum `ComputingType` instead of index**
* remove unusable refs to `ONE_BUS_SHORTCIRCUIT_ANALYSIS` 
* remove unecessary `getText`/`getRunningText`
* factorise functions (start_computation & action_on_runnable)
* remove unused (copy/paste error) prop-types of _run-button-container_


src/components/run-button.js
```diff
@@ -5,104 +5,106 @@
...
+            selectedIndex={activeRunnables.indexOf(selectedRunnable)}
+            onSelectionChange={(index) =>
+                setSelectedRunnable(activeRunnables[index])
             }
-            actionOnRunnable={handleActionOnRunnable}
+            onClick={runnables[selectedRunnable].startComputation}
+            runningStatus={getRunningStatus()}
+            buttonDisabled={disabled || isButtonDisable()}
+            selectionDisabled={disabled}
+            text={runnablesText[selectedRunnable] || ''}
```
Meklo, line 90: The label for the selected runnable isn't showing up, I reckon it is caused by trying to access `runnablesText` with `selectedRunnable` although `runnablesText` is a flat array

AAJELLAL (changes requested):
No label is displayed at the button even when selecting a calculation.

emptyLabel

src/components/run-button-container.js
```diff
@@ -221,289 +192,278 @@ export function RunButtonContainer({ studyUuid, currentNode, disabled }) {
...
-                default:
-                    return;
-            }
-
-            dispatch(setComputingStatus(type, RunningStatus.IDLE));
-            setComputationStopped(true);
+    const startComputationAsync = useCallback(
+        (computingType, fnBefore, fnStart, fnThen, fnCatch, errorHeaderId) => {
+            dispatch(setComputingStatus(computingType, RunningStatus.RUNNING));
+            fnStart.then(fnThen).catch((error) => {
```
Meklo, line 198: Since fnStart is an arrow function, shouldn't it be :
```suggestion
            fnStart().then(fnThen).catch((error) => {
```

src/components/run-button-container.js
```diff
@@ -221,289 +192,278 @@ export function RunButtonContainer({ studyUuid, currentNode, disabled }) {
...
+                    );
+                },
+            },
+            [ComputingType.VOLTAGE_INIT]: {
+                messageId: 'VoltageInit',
+                startComputation() {
+                    startComputationAsync(
+                        ComputingType.VOLTAGE_INIT,
+                        null,
+                        startVoltageInit(studyUuid, currentNode?.id),
```
Meklo, line 389: Should be an arrow function

src/components/run-button-container.js
Tristan-WorkGH: yes, thank

src/components/run-button.js
Tristan-WorkGH: yes that's it

src/components/run-button-container.js
```diff
@@ -221,292 +192,287 @@ export function RunButtonContainer({ studyUuid, currentNode, disabled }) {
...
-                .then(setRanLoadflow(true))
-                .catch((error) => {
-                    dispatch(
-                        setComputingStatus(
-                            ComputingType.LOADFLOW,
-                            RunningStatus.FAILED
-                        )
+    const runnables = useMemo(() => {
+        function actionOnRunnables(type, fnStop) {
+            fnStop.finally(() => {
```
Meklo, line 263:
```suggestion
            fnStop().finally(() => {
```
There's also an error here when I try to interrupt a computation

Tristan-WorkGH: It seem I've a problem...

Meklo (approved):

AAJELLAL (approved):
I can execute and stop voltage profile initialization and the other calculation. Tests OK.

PR #1765, opened 2023-12-05
improve httpd caching configuration (etags, cache-control)

jonenst:
- restore etags on all files except index.html
- add cache-control no-cache to force revalidation on all files with static names (e.g env.json)
- add cache-control "max-age=31536000, immutable" to all files with hash in their names


app-httpd.conf
```diff
@@ -8,9 +8,38 @@
...
+          #   if your editor replaces the file when saving), but this means using Digest doesn't
+          #   solve this issue. This is a known docker limitation: https://github.com/moby/moby/issues/6011
+          FileETag Digest
+        </FilesMatch>
         Options +Includes
+        Header set Cache-Control "no-cache"
+    </Directory>
+
+    <Directory /opt/bitnami/apache/htdocs/gridstudy/static/>
+        Header set Cache-Control "max-age=31536000, immutable"
```
Tristan-WorkGH, line 42: isn't setting a cache of 1 year a little too long? we have etag, so we can set something like 1 week/month instead?

jonenst: Added a comment

```
<Directory /opt/bitnami/apache/htdocs/gridstudy/static/>
# All files in this folder are public and have in their name
# the hash of their content, so cache them very aggressively
# as per recommended best practices
```

Tristan-WorkGH (approved):

PR #1766, opened 2023-12-06
don't hide equipment table headers

AbdelHedhili:


klesaulnier (approved):

AAJELLAL (approved):
Test OK: header visible even if no data to display.

PR #1767, opened 2023-12-06
Fixed display bug on cut/paste node before another one

ne0ds:


EstherDarkish (changes requested):
There's a new bug :
- With a study and (only) node N1
- from N1, create a new node N2 above N1 => Bug, N2 should be between the root node and N1
- delete N2
- from N1, create a new node N3 "in a new branch"
- from N3, cut N3
- from N1, paste N3 above N1 => Same bug, N3 is not displayed between the root node and N1

**EDIT : It is fixed.**

EstherDarkish (approved):
Code OK, Tests OK

klesaulnier (approved):

PR #1775, opened 2023-12-08
update traduction file

ghazwarhili:


Tristan-WorkGH (approved):
Wouldn't it be better to use [with an argument](https://formatjs.io/docs/core-concepts/icu-syntax) for the level?  like `"VoltageLevelIdT3WSide": "Voltage level ID {side}"`

ok otherwise

AAJELLAL (approved):

PR #1779, opened 2023-12-12
Fixes the gap below reports.

EstherDarkish:


jamal-khey (approved):
Test and Code Ok.

dbraquart (changes requested):
code: ok
tests: issue found
- when we open a study and go directly to LOGS tab, the fix does not work (160 gap is visible). When we go to RESULTS tab and go back to LOGS tab, the fix is working.

dbraquart (changes requested):
code : ok?
tests: still one strange issue
- test case: while building a node, clic faster between RESULTS tab and LOGS tab. usually that's enough to cause a crash.
The error is reported in function _areSomeSeveritiesUnchecked_ (because selectedItems is undefined).
Even if I fix this this function (adding a check _selectedItems && ..._) an error is reported elsewhere ...

dbraquart (approved):
code: ok
tests: ok

PR #1780, opened 2023-12-12
Show results even after applying voltage init modifications

FranckLecuyer:


anistouri (approved):

dbraquart (changes requested):
code: ok
tests: issue found
- when we have some voltage-init results, and change the network (ex: open a switch), then the results are cleaned as expected, but the button "VIEW MODIFICATIONS" is still enable, and when we click on it, we have a 404 error snackbar

dbraquart (approved):
code: ok
tests: ok

PR #1787, opened 2023-12-14
WIP Fixes a case where JSON.parse would try to parse undefined.

EstherDarkish:


src/components/graph/menus/modification-list-item.js
```diff
@@ -77,7 +77,9 @@ export const ModificationListItem = ({
         finally, we uses the default value of equipmentId or empty string
     */
     const getComputedLabel = useCallback(() => {
-        const modificationMetadata = JSON.parse(modif.messageValues);
+        const modificationMetadata = modif?.messageValues
+            ? JSON.parse(modif.messageValues)
+            : {};
```
Mathieu-Deharbe, line 82: If {} is affected to modificationMetadata won't you have some similar trouble in the following switch ? (return modificationMetadata.lineToSplitId etc)

EstherDarkish:
This is maybe not the way to solde this issue. Changed the PR to WIP.

SlimaneAmar (changes requested):
Message value is never null !

EstherDarkish:
Not the right way to fix the underlying issue.

PR #1789, opened 2023-12-14
Add battery tabular modifications

FranckLecuyer:


src/components/dialogs/network-modifications/tabular-modification/tabular-modification-form.tsx
```diff
@@ -53,15 +53,19 @@ const TabularModificationForm = () => {
...
-        [intl, watchType]
-    );
+    const explanationComment = useMemo(() => {
+        if (watchType === EQUIPMENT_TYPES.GENERATOR) {
+            return intl.formatMessage({
+                id: 'TabularModificationGeneratorSkeletonComment',
+            });
+        } else if (watchType === EQUIPMENT_TYPES.BATTERY) {
+            return intl.formatMessage({
+                id: 'TabularModificationBatterySkeletonComment',
```
dbraquart, line 63: I think we dont need to manage a comment in the CSV file if there is nothing in the comment (unlike generator case). For instance there is no comment in the voltage-level case.
We can remove it.

dbraquart (changes requested):
code: I think the CSV comment is not required
tests: ok

src/components/dialogs/network-modifications/tabular-modification/tabular-modification-form.tsx
FranckLecuyer: Removed

dbraquart (approved):
code: ok
tests: ok

PR #1800, opened 2023-12-18
Upgrade to react-map-gl 7, using deckgl as a control

jonenst:
- with react-map-gl 7, we are forced to choose between
  * Root Deck, Map as child. Deckgl handles all camera and user input like before The major downside is that react-map-gl 7 dropped support for using their controls (NavigationControl, the '+/-/compass' at the top right
  * Root Map, deck as child. React-map-gl handles all camera and user input. some advanced deckgl features are not available, and it changes what we were doing.
- right click on a feature is not exposed in "MapRoot/DeckglChild" mode for now, implement it manually as per https://github.com/visgl/deck.gl/issues/8326
- remove project_size_at_latitude copy, it seems accessible now? Removing it fixes the following error and the arrows are still correctly displayed: luma.gl: GLSL compilation errors in vertex shader arrow-layer-vertex-shader-1 658:  * Just to call it from project_size_all_zoom_levels(). 659:  */ 660: float project_size_at_latitude(float lat) { ^^^ ERROR: 'project_size_at_latitude' : function already has a body deck: initialization of ArrowLayer({id: 'lineLayer-ArrowNominalVoltage400'}): GLSL compilation errors in vertex shader arrow-layer-vertex-shader-1


EstherDarkish (commented):
Some observations : 
- We cannot resize diagrams (SLD and NAD) anymore
- The sensibility of the camera movement with right click and hold + move the mouse is really high

antoinebhs (commented):
In addition to Charly's comment:
* I notice that when rotating the map and then clicking on the compass, it switched to the default view before, now it only aligns the map with the North direction
* there is a smoothing effect when you drag the map, I don't know if it's expected. (edit: looking at the code, it seems expected)

src/components/diagrams/diagram-common.js
```diff
@@ -117,6 +117,8 @@ export const styles = {
         },
     },
     paperBorders: (theme) => ({
+        position: 'relative',
+        zIndex: 10, // now needs to be at least 2 for the sld to be above our deckgl lines and substations. TODO duplicated on many elements, improve our stacking context handling ?
```
antoinebhs, line 121: at least 10 ?

EstherDarkish (approved):

antoinebhs (approved):
Code OK
Tests OK

PR #1807, opened 2023-12-19
Update row data in spreadsheet after the modification has been applied

ayolab:


Meklo:
It's not really changes requests since I won't be here to finish the review but here's a few points : 
- I guess it's necessary to fetch an equipment data only if the edit failed, but I reckon we currently lack the granularity of information to retrieve this information through notifications. Nevertheless it would be interesting to add a condition to do so if it is possible 
- The retrieval of equipments data is done through the hook `useSpreadsheetEquipments`, would it be possible to add your logic the this hook ?
- Instead of a applying a transaction, if you updated the equipment through `equipments` collection retrieved by `useSpreadsheetEquipments` wouldn't it do the trick ?

klesaulnier (approved):

anistouri (approved):

ayolab:
> It's not really changes requests since I won't be here to finish the review but here's a few points :
> 
> * I guess it's necessary to fetch an equipment data only if the edit failed, but I reckon we currently lack the granularity of information to retrieve this information through notifications. Nevertheless it would be interesting to add a condition to do so if it is possible
> * The retrieval of equipments data is done through the hook `useSpreadsheetEquipments`, would it be possible to add your logic the this hook ?
> * Instead of a applying a transaction, if you updated the equipment through `equipments` collection retrieved by `useSpreadsheetEquipments` wouldn't it do the trick ?

1) as you said we don't have that level of granularity,
2) updating `equipments` will result in a rerender of the whole table

PR #1810, opened 2023-12-20
Add control on sensi compute count

ghazwarhili:


src/components/dialogs/parameters/sensi/sensitivity-analysis-parameters.js
```diff
@@ -544,7 +557,48 @@ export const SensitivityAnalysisParameters = ({
...
+        return (
+            <Box sx={styles.modificationsTitle}>
+                <Box sx={styles.icon}>
+                    <CircularProgress
+                        size={'1em'}
+                        sx={styles.circularProgress}
+                    />
+                </Box>
+                <Typography noWrap>
+                    <FormattedMessage id={'computing in progress'} />
```
AAJELLAL, line 570: You need to move this to translation files instead of hardcoding it.

src/components/dialogs/parameters/sensi/sensitivity-analysis-parameters.js
```diff
@@ -544,7 +557,48 @@ export const SensitivityAnalysisParameters = ({
...
+                </Typography>
+            </Box>
+        );
+    };
+
+    const renderComputingEvent = () => {
+        return (
+            <Alert severity={isMaxReached ? 'error' : 'info'}>
+                {analysisComputeComplexity > 999999 ? (
```
AAJELLAL, line 579: Why 999999? It should be 500 000 No?

src/components/voltage-init-result.js
AAJELLAL, line 1: You need to test voltage init to make sure that there is no regression after those changes.

src/components/dialogs/parameters/sensi/sensitivity-analysis-parameters.js
ghazwarhili: To limit the display of simulated calculations to 6 digits, if the number exceeds 999999
we display the message 'SimulatedCalculationExceedsLimit' with the aim of maintaining the box with a  fixed size."

src/components/dialogs/parameters/sensi/sensitivity-analysis-parameters.js
ghazwarhili: ok

src/components/dialogs/parameters/sensi/sensitivity-analysis-parameters.js
```diff
@@ -279,10 +298,10 @@ export const SensitivityAnalysisParameters = ({
...
                             `${arrayFormName}[${index}].[${COUNT}]`,
-                            value && Math.abs(value)
+                            parseIntData(value, 0)
                         );
                         setAnalysisComputeComplexity(getResultCount());
-                        onFormChanged(false);
+                        setLaunchLoader(false);
                     });
                 })
                 .catch((error) => {
```
anistouri, line 307: If you put it in the then only, you'll have an infinite loading if you receive an error. Either you add it in the catch as well, or you place it after the then and catch

ghazwarhili: done

ghelloumi (approved):

AAJELLAL (approved):

anistouri (approved):

PR #1814, opened 2023-12-21
FIX: computation multiple starts

sBouzols:
disable run button disallowing  computation multi starts. 
Only allow one computation running at a time


AbdelHedhili (approved):

jamal-khey (approved):
Test and code Ok

PR #1819, opened 2023-12-22
map manual refresh, fix overlay zIndex not applied because position is not set

jonenst:
this gives it the same behavior as the initial "map loading" overlay


EstherDarkish (approved):
Code OK Tests OK

antoinebhs (approved):
Code OK 
Test OK

PR #1827, opened 2023-12-22
increase map maxZoom 12->14

jonenst:
The new version of deckgl/react-map-gl fixes the bug where you could zoom in more than the max zoom. So now we realize that users actually need a higher maxZoom.

We had chosen 12 because this is the limit that triggers the bug where the text label of active power on the map moves away from the arrow on the line so this will give this bug a little more visibility but it's not a blocker.


antoinebhs (approved):
Code OK
We'll have to fix this label offset bug when zoom > 12 then... 🎁

EstherDarkish (approved):
Test OK code OK

PR #1831, opened 2023-12-27
Add calls for filtering report severity list

lilian-houdelet:
When we merge this PR we have to invalidate all nodes on dev-opf in order to delete all existing logs. Otherwise, logs wont be correctly displayed (cf invalidate_nodes_builds.py from admin-tools)


src/components/report-viewer/report-viewer.js
```diff
@@ -199,8 +219,12 @@ export default function ReportViewer({
         setSelectedNode(rootId);
         setExpandedNodes([rootId]);
         setLogs(rootReport.current.getAllLogs());
-        setSelectedSeverity(LogReportItem.getDefaultSeverityFilter());
-    }, [jsonReportTree, createReporterItem]);
+        Promise.resolve(getSeverityPromise(rootId)).then((response) => {
```
flomillot, line 222: Why `Promise.resolve `?
Can you catch and log the eventual error ?

src/components/report-viewer/report-viewer.js
```diff
@@ -220,7 +244,12 @@ export default function ReportViewer({
 
     const selectNode = (nodeId) => {
         if (selectedNode !== nodeId) {
-            refreshNode(nodeId, selectedSeverity);
+            Promise.resolve(getSeverityPromise(nodeId)).then((fetchedData) => {
```
flomillot, line 247: Same

src/components/results/common/computation-report-viewer.tsx
```diff
@@ -107,13 +109,32 @@ export const ComputationReportViewer: FunctionComponent<
...
 
+    const subReportSeveritiesPromise = (reportId: string) => {
+        fetchSubReportSeverity(
+            studyUuid.toString(),
+            currentNode.id.toString(),
+            reportId
+        );
+    };
+
+    const reportSeveritiesPromise = (reportId: string) => {
```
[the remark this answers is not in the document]
lilian-houdelet: I used the same parameters than the one that gather the reports

dbraquart (changes requested):
As discussed, severity list should be returned as a taskValue per LogReport (tree node on the left pane). No REST call should be needed.

src/components/report-viewer/log-report.js
```diff
@@ -80,6 +82,14 @@ export default class LogReport {
         return this.logs;
     }
 
+    getSeverityList() {
+        return this.severityList;
+    }
+
+    getAllSeverityList() {
```
flomillot, line 89:
```suggestion
    // Represent all the different severities of this report and it's subreports
    getAllSeverityList() {
```
I would add something like that

src/components/report-viewer/log-report-item.js
```diff
@@ -60,11 +60,19 @@ export default class LogReportItem {
         },
     };
 
-    static getDefaultSeverityFilter() {
+    static getDefaultSeverityFilter(severityList) {
         const filterConfig = {};
-        Object.values(LogReportItem.SEVERITY).forEach((s) => {
-            filterConfig[s.name] = s.displayedByDefault;
-        });
+        if (!severityList) {
```
flomillot, line 65:
```suggestion
        if (!severityList && !severityList.lengh) {
```

src/components/report-viewer/log-report.js
lilian-houdelet: I added it in the constructor and not on the getter

flomillot (approved):
Code OK

dbraquart (changes requested):
code: ok
tests: ok

But we still have an issue: how do we treat the former logs compatibility ? to be discussed with TLs

dbraquart (approved):
code: ok
tests: ok

PR #1834, opened 2023-12-29
Mcs tabular modification

ne0ds:


src/translations/en.json
```diff
@@ -702,6 +702,8 @@
     "TabularModificationError": "Error while creating tabular modification",
     "TabularModificationGeneratorSkeletonComment" : "#,HYDRO | NUCLEAR | WIND | THERMAL | SOLAR | OTHER,,,,,,true | false",
     "TabularModificationLoadSkeletonComment": "#,AUXILIARY | FICTITIOUS,,",
+    "TabularModificationShuntSkeletonComment": "\"# For each shunt compensator it is possible to modify either the maximum reactive power (and the type) or the maximum susceptance. In case of conflicting input, the maximum susceptance will be ignored.\"",
```
ayolab, line 705: By surrounding the `#` with `"`, paparse considers it as part of the cell's text, In order for the comment line to be ignored you should put the `#` in the first column and the rest of the comment in the second one

src/translations/fr.json
```diff
@@ -861,7 +861,7 @@
     "Substation": "Site",
     "BusBarSections": "Sections de jeu de barre",
     "BusBarCount": "Nb de barres",
-    "SectionCount": "Nb de sections",
+    "voltageSectionCount": "Nb de sections",
```
ayolab, line 864: I think that we should find an other name for this key

ayolab (changes requested):

src/translations/fr.json
ne0ds: This is the key used in backend...

ayolab (approved):

YenguiSeddik (approved):

PR #1840, opened 2024-01-05
Redirect to last launched computation result tab

Meklo:


souissimai (approved):

src/components/results/use-results-tab.ts
```diff
@@ -38,11 +38,85 @@ export type ResultTabIndexRedirection =
...
+        case ComputingType.DYNAMIC_SIMULATION:
+        case ComputingType.VOLTAGE_INIT:
+        case ComputingType.ALL_BUSES_SHORTCIRCUIT_ANALYSIS:
+        case ComputingType.ONE_BUS_SHORTCIRCUIT_ANALYSIS:
+        case ComputingType.NON_EVACUATED_ENERGY_ANALYSIS:
+            return rootResultTabIndex;
+        default:
+            return ResultsTabsRootLevel.LOADFLOW;
+    }
+};
```
klesaulnier, line 109: From what I understand, this would work as well and would be easier to understand
I don't understand this switch role

```suggestion
//function to convert computing type to the index corresponding to the resulting computation tab.
//the list of services must also be passed as parameter because its content is dynamic depending on the state of the application (e.g. developer mode disabled)
export const computingTypeToRootTabRedirection = (
    computingType: ComputingType,
    availableServices?: IService[]
): ResultTabIndexRedirection => {
    return availableServices
        ? availableServices
              .map((service) => service.computingType)
              .findIndex((computations) => computations.includes(computingType))
        : ResultsTabsRootLevel.LOADFLOW;
};
```

src/components/results/use-results-tab.ts
```diff
@@ -38,11 +38,85 @@ export type ResultTabIndexRedirection =
...
+
+const useResultsTabRedirectionLock = (): [
+    Boolean,
+    Dispatch<SetStateAction<Boolean>>
+] => {
+    const computationStatus: ComputingStatus = useSelector(
+        (state: ReduxState) => state.computingStatus
+    );
+    const [redirectionLock, setRedirectionLock] = useState<Boolean>(false);
```
klesaulnier, line 63: add some comments around "setRedirectionLock" usages to explain why we should lock/unlock redirections

src/components/result-view-tab.tsx
```diff
@@ -280,20 +307,22 @@ export const ResultViewTab: FunctionComponent<IResultViewTabProps> = ({
         );
     };
 
-    useEffect(() => {
-        if (!enableDeveloperMode) {
-            // a displayed tab may be obsolete when developer mode is disabled, then switch on first one
-            setTabIndex(0);
-        }
-    }, [enableDeveloperMode]);
```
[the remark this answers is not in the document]
Meklo: Indeed, I've added it back again

src/components/results/use-results-tab.ts
Meklo: The switch was originally useful because the specific redirection related to shortcircuit sub computations were also handled here but since I moved them out of it you're right it's now useless

src/components/results/use-results-tab.ts
Meklo: Done

PR #1844, opened 2024-01-05
Fix DS Title in dialog translation warning

sBouzols:
fix(): return only the titleId which is used un ModificationDialog later and translate


thangqp (approved):
Code OK
Test OK

PR #1854, opened 2024-01-10
Fix style in parameters tab

lilian-houdelet:


Tristan-WorkGH (approved):

ghazwarhili (commented):
code => OK
test :
-> the buton validate of sensi tab is not in the right position.
-> the space between the label and the provider component doesn't seem right

souissimai (approved):

PR #1865, opened 2024-01-16
Unify illustrations

souissimai:


Mathieu-Deharbe:
I don't understand the color code. 

Green seems to be what doesn't change. But pink, red and orange ? What do they mean ?

Mathieu-Deharbe:
> > I don't understand the color code.
> > Green seems to be what doesn't change. But pink, red and orange ? What do they mean ?
> > 9DCC9F ==> 81C784 Green
> > A5D6A7 ==> 81C784 Green
> > FFA4C2 Pink
> > 90caf9  ==> 64B5F6 blue
> > FFCC80 ==> FFB74D yellow
> > here the old and the new hex colors

I am taking about what they mean **from the user viewpoint**.
* green => part of the plan that doesn't change
* pink => ??
* blue => ??
* orange => ??

When I made the first illustration pink was the part removed, and blue the part added by the transformation. Now sometimes pink is on the left, sometimes on the right, and in one illustration orange isused. Does it mean something ? or is it random ?

Mathieu-Deharbe (approved):

PR #1869, opened 2024-01-17
Fix cancel property deletion in substation modification

ayolab:


dbraquart (commented):
`// we can mark as deleted only prop having a previous value, not added in current modification`

`if (
  properties[idx][PREVIOUS_VALUE] &&
  properties[idx][ADDED] === false
) {`


why not just making the test on ADDED ? I cant see a case where testing PREVIOUS_VALUE is usefull ??

Meklo:
I reckon the issue now also impacts loads modification form since the ability to add/modify properties was recently added there. It'd may be sound to also implement the fix for the loads form

PR #1875, opened 2024-01-19
Updates the cancel button

EstherDarkish:
Linked pull requests : 
https://github.com/gridsuite/gridexplore-app/pull/330
https://github.com/gridsuite/commons-ui/pull/333


src/components/app-wrapper.js
```diff
@@ -115,8 +114,15 @@ let lightTheme = createTheme({
...
-            main: lightTheme.palette.text.primary,
+        cancelButtonColor: {
+            main: lightTheme.palette.text.secondary,
+        },
+    },
+    components: {
+        CancelButton: {
+            defaultProps: {
+                color: 'cancelButtonColor',
+            },
```
anistouri, line 125: By doing this and not specifying the button color in the props, we will have the same button color for all cancel buttons, whereas in certain cases, the button color should be changed, for example, for the deletion of a study

EstherDarkish: That is true, but this becomes a non-issue because we now want to homogenize every cancel and validate buttons. We do not want different colors anymore (that's the goal of this US).

anistouri (approved):

ghazwarhili (approved):

PR #1876, opened 2024-01-19
Add condition to fix the filter with the wrong unit

lilian-houdelet:
Must be close if https://github.com/gridsuite/gridstudy-app/pull/1891 is accepted and merged


dbraquart (changes requested):
code: just a rename requested
tests: 1 issue to check

When we enter **2.5** as filter value, we ask to the back something different:
filters=[{"column":"ipMin","dataType":"number","type":"lessThanOrEqual","value":"**2000**"}]
pb conversion

src/hooks/use-aggrid-row-filter.ts
```diff
@@ -119,11 +125,17 @@ export const useAggridRowFilter = (
 
             const { value, type, dataType } = selectedValue;
 
+            let backendValue = value;
+
+            if (dataType === 'number' && kiloValues.includes(field)) {
+                backendValue = String(parseInt(value) * 1000);
```
dbraquart, line 131: The issue was not in the roundToPrecision() method.
you cannot use parseInt here : 2.5 => 2 => 2000

dbraquart (changes requested):
code: parseInt issue

dbraquart (approved):
code: ok
tests: ok

src/hooks/use-aggrid-row-filter.ts
```diff
@@ -119,11 +126,17 @@ export const useAggridRowFilter = (
 
             const { value, type, dataType } = selectedValue;
 
+            let backendValue = value;
+
+            if (dataType === 'number' && kiloValues.includes(field)) {
```
dbraquart, line 131: We are using short-circuit specific data to update the value here, in a **common** component: this is not very good.
Could we try to use a callback 'updateValueCallbacl', passed only by short-circuit-analysis-result, to update the value ?

dbraquart (changes requested):
code: maybe a little change to use a callback ?
tests: ok

PR #1877, opened 2024-01-19
Fix first built node not selected at start some times

sBouzols:
We have to manage `checkNetworkExistenceAndRecreateIfNotFound`, `checkStudyIndexation` and `loadTree` orchestration to avoid random bug if status are set in bad order.

Sometimes a study containing a built node just after le root node is mis initialized at start.
A start we want to select the first built node after the root node or the rot node if not exist
To reproduce 100% on the main branch, add this line in `checkStudyIndexation`
```
return fetchStudyIndexationStatus(studyUuid)
        .then(x => new Promise(resolve => setTimeout(() => resolve(x), 1000))) // <- this one
        .then((status) => {
```
This simulate a long query to check indexation status.
Then in loadTree we don't have the `indexationStatus` yet and we select the root node.


EstherDarkish (approved):
Code OK Test OK

src/components/study-container.js
```diff
@@ -530,7 +582,9 @@ export function StudyContainer({ view, onChangeTab }) {
                     if (response.status === HttpStatusCode.OK) {
                         successCallback && successCallback();
                         setIsStudyNetworkFound(true);
-                        loadTree();
+                        checkStudyIndexation().then((status) => {
+                            loadTree(status);
+                        });
```
ne0ds, line 587:
```suggestion
                        checkStudyIndexation().then(loadTree);
```

ne0ds:
Tests OK

ne0ds (approved):

PR #1879, opened 2024-01-22
enhancement for calcul sensi param box

ghazwarhili:


AAJELLAL (changes requested):
There still remains the calculation problem when opening the settings window. If certain lines are disabled then the calculation is not done on the next line.

AAJELLAL (commented):
Test ok. 
One minor code remark.

AAJELLAL (approved):

PR #1888, opened 2024-01-25
fix(properties): handle empty properties when copying

TheMaskedTurtle:
I choose to use an empty list by default for properties and to not send it to the backend when the (filtered) list is empty.


Mathieu-Deharbe (approved):

dbraquart (approved):
code: ok
tests: ok

PR #1897, opened 2024-01-31
Dynamic Simulation - Curve Config - Variable and model translation

thangqp:
Real translation will be done in another PR..


src/components/dialogs/parameters/dynamicsimulation/curve/dialog/model-filter.js
```diff
@@ -91,6 +96,31 @@ const variablesTreeToVariablesArray = (variablesTree, parentId) => {
...
+                id: `${PREFIX_VARIABLE_KEY}/${elem.name}`,
+            });
+        } else {
+            // must be a variable set element
+            return intl.formatMessage({
+                id: `${PREFIX_VARIABLE_SET_KEY}/${elem.name}`,
+            });
+        }
+    }
+};
```
EstherDarkish, line 122: For this type of function, I think it is easier to read if we do not write "else" after a return, like this : 
```suggestion
const makeGetVariableLabel = (intl) => (elem) => {
    if (!elem.parentId) {
        // root element => that is model element in the variable tree
        return intl.formatMessage({ id: `${PREFIX_MODEL_KEY}/${elem.name}` });
    } 
    // either a variable set element or variable element in the variable tree
    if (elem.variableId) {
        // that is a variable element
        return intl.formatMessage({
            id: `${PREFIX_VARIABLE_KEY}/${elem.name}`,
        });
    } 
    // must be a variable set element
    return intl.formatMessage({
        id: `${PREFIX_VARIABLE_SET_KEY}/${elem.name}`,
    });
};
```

src/translations/dynamic-mapping-models-fr.ts
```diff
@@ -0,0 +1,64 @@
+/**
+ * Copyright (c) 2024, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+const dynamic_mapping_models_fr = {
```
EstherDarkish, line 8: French translation missing, this is still in english.

src/translations/dynamic-mapping-models-en.ts
```diff
@@ -0,0 +1,64 @@
+/**
+ * Copyright (c) 2024, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+const dynamic_mapping_models_en = {
+    // --- models --- //
+    'models/GeneratorSynchronousThreeWindings':
```
EstherDarkish, line 10: To be more consistent with the rest of the application, we should use dots instead of slashs for separators

src/translations/dynamic-mapping-models-fr.ts
thangqp: The real translate will be done in another PR later as discussion in the Azure ticket

src/translations/dynamic-mapping-models-en.ts
thangqp: Yes! DONE

src/components/dialogs/parameters/dynamicsimulation/curve/dialog/model-filter.js
thangqp: Yes! DONE

EstherDarkish (approved):

ghazwarhili (approved):

PR #1898, opened 2024-01-31
Handle sharing and saving the security analysis parameters.

AAJELLAL:


src/components/dialogs/parameters/common/parameters-creation-dialog.tsx
```diff
@@ -97,13 +95,13 @@ const CreateParameterDialog: React.FunctionComponent<CreateParameterProps> = ({
     const onSubmit = useCallback(
         (values: FormData) => {
             createParameter(
-                formatNewParams(parameterGetValues()),
+                parameterFormatter(parameterValues()),
```
jamal-khey, line 98: it would be add a .catch to handle errors or put a log in the console

AAJELLAL: I only changed the name to be more generic.

jamal-khey (approved):

jamal-khey (approved):

ayolab (approved):

PR #1902, opened 2024-02-07
add snackbar on maintenance message

AbdelHedhili:


src/components/app.js
```diff
@@ -141,6 +145,12 @@ const App = () => {
 
     const [tabIndex, setTabIndex] = useState(0);
 
+    const [maintenanceSnackBarId, setMaintenanceSnackBarId] = useState(null);
```
jamal-khey, line 148: what happend if we recieve a second message before the first one expire ?

AbdelHedhili: now the previous msg is cancelled when receiving a new one

PR #1903, opened 2024-02-07
Sorting and filtering in shortcircuit results

Mathieu-Deharbe:
Enable sorting and filtering on the 'child' connectableId column of the shortcircuit results

Adds the "secondary" sort concept for this column. Which is a facultative sort applied after the primary sort.


flomillot (commented):
As discussed, maybe try to change the sortProps and get inspiration from the AGGrid API : https://www.ag-grid.com/javascript-data-grid/row-sorting/#sorting-api
Ex : 
```
[
      { colId: 'country', sort: 'asc', type: 'primary' },
      { colId: 'sport', sort: 'asc', type: 'secondary' },
    ]
```

src/components/custom-aggrid/custom-aggrid-header.js
```diff
@@ -247,6 +261,21 @@ const CustomHeaderComponent = ({
...
+                                                <ArrowUpward
+                                                    sx={styles.iconSize}
+                                                />
+                                            ) : (
+                                                <ArrowDownward
+                                                    sx={styles.iconSize}
+                                                />
+                                            )}
+                                        </IconButton>
+                                    </Grid>
```
klesaulnier, line 277: check with designer if it's enough to display "secondary filter"
I would display a "2" next to the children sort arrow

klesaulnier (changes requested):
I did not test short-circuit results yet, but I could see it broke other results pages, multiple arrow are displayed in a weird way

Mathieu-Deharbe:
> As discussed, maybe try to change the sortProps and get inspiration from the AGGrid API : https://www.ag-grid.com/javascript-data-grid/row-sorting/#sorting-api Ex :
> 
> ```
> [
>       { colId: 'country', sort: 'asc', type: 'primary' },
>       { colId: 'sport', sort: 'asc', type: 'secondary' },
>     ]
> ```
> 
> or
> 
> ```
> [
>       { colId: 'country', sort: 'asc', secondary: 'false' },
>       { colId: 'sport', sort: 'asc', secondary: 'true' },
>     ]
> ```
> 
> Then we remove lot of duplicate code

Juste tried this and commited it. I think it is better. Less duplicate and more logic directly into use-aggrid-sort. Tell me if this looks like what you expected.

src/components/custom-aggrid/custom-aggrid-header-utils.ts
```diff
@@ -25,9 +31,10 @@ export const makeAgGridCustomHeaderColumn = ({
     const isSortable = !!sortProps;
     const isFilterable = !!filterProps;
     const isCurrentColumnSorted = sortConfig?.colKey === field;
+    const isColumnSecondarySorted = sortConfig?.secColKey === field;
```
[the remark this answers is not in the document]
Mathieu-Deharbe: Done.

src/components/custom-aggrid/custom-aggrid-header.js
Mathieu-Deharbe: I did last week. He didn't like the "2" idea, and made a proposition which looks better but won't be done for now. I precised it in the main US.

src/hooks/use-aggrid-sort.ts
```diff
@@ -7,13 +7,49 @@
...
+export function getSecondarySort(
+    multipleSort: SortConfigType
+): ColumnSortConfig | undefined {
+    for (const sort of multipleSort) {
+        if (sort.secondary) {
+            return sort;
+        }
+    }
+    return undefined;
+}
```
flomillot, line 45: No need these methods

src/hooks/use-aggrid-sort.ts
```diff
@@ -32,22 +68,61 @@ export const getSortValue = (sortWay: number) => {
...
+                        {
+                            colKey: colKey,
+                            sortWay: sortWay,
+                            secondary: false,
+                        },
+                    ]);
+                }
+            }
+        },
+        [sortConfig]
```
flomillot, line 114: It's not really what I was thinking about.. what about that instead ? I am right ?
```suggestion
        (colKey: string, sortWay: number, secondary: boolean) => {
            setSortConfig(
                sortConfig
                    .filter((sort) => sort.secondary !== secondary)
                    .concat({ colKey, sortWay, secondary })
            );
        },
        [sortConfig]
```

flomillot (changes requested):
You don't need to use primarySort or secondarySort principle inside the column because you already know it thanks to the column definition. Here it's too similar to what we had before...

PR #1908, opened 2024-02-08
Add generators tabular creations

FranckLecuyer:


EstherDarkish (commented):
If we do not select a type but click on "generate skeleton" anyway, the button is gray but works, and it downloads a "null_skeleton.csv" file that is not correctly built.

FranckLecuyer:
> If we do not select a type but click on "generate skeleton" anyway, the button is gray but works, and it downloads a "null_skeleton.csv" file that is not correctly built.

Fixed (also in tabular modifications)

EstherDarkish (approved):
Test OK Code OK

PR #1909, opened 2024-02-08
Use customHeader in spreedsheet and implement sorting and filtering

achour94:


src/components/spreadsheet/utils/config-tables.js
klesaulnier, line 1: create objects like this to improve maintenance :
const integerFilterConf = {
    filterDataType: FILTER_DATA_TYPES.NUMBER,
    filterComparators: [
        FILTER_NUMBER_COMPARATORS.GREATER_THAN_OR_EQUAL,
        FILTER_NUMBER_COMPARATORS.LESS_THAN_OR_EQUAL,
        FILTER_NUMBER_COMPARATORS.NOT_EQUAL,
    ],
}

and use it everywhere it's needed
It could even include "filter: 'agNumberColumnFilter'" and "numeric: true"

same for FILTER_DATA_TYPES.TEXT

src/components/spreadsheet/utils/config-tables.js
```diff
@@ -241,13 +254,29 @@ export const TABLES_DEFINITIONS = {
...
                 field: 'id',
-                sort: DEFAULT_SORT_ORDER,
+                customFilterParams: {
+                    filterDataType: FILTER_DATA_TYPES.TEXT,
+                    filterComparators: [
+                        FILTER_TEXT_COMPARATORS.STARTS_WITH,
+                        FILTER_TEXT_COMPARATORS.CONTAINS,
+                    ],
+                },
+                filter: 'agTextColumnFilter',
```
klesaulnier, line 264: do we still need "filter: 'agTextColumnFilter'" ?

src/hooks/use-aggrid-sort-impl.ts
klesaulnier, line 1: I think this hook name is a bit confusing
Change it to something like "useAggridLocalSort" or add a comment to explain its usage

src/hooks/use-aggrid-row-filter-impl.ts
klesaulnier, line 1: I think this hook name is a bit confusing
Change it to something like "useAggridLocalRowFilter" or add a comment to explain its usage

src/hooks/use-aggrid-sort-impl.ts
```diff
@@ -0,0 +1,44 @@
...
+                state: [
+                    {
+                        colId: sortConfig.colKey,
+                        sort: getSortValue(sortConfig.sortWay),
+                    },
+                ],
+                defaultState: { sort: null },
+            });
+        },
+        [gridRef]
```
klesaulnier, line 36: it's a bit weird to pass a ref as useCallback dependency but I did not find any way to do without it

src/hooks/use-aggrid-row-filter-impl.ts
```diff
@@ -0,0 +1,62 @@
...
+    updateFilterCallback = () => {}
+): UseAggridRowFilterOutputType => {
+    const { updateFilter, filterSelector, initFilters } = useAggridRowFilter(
+        filterSelectorKeys,
+        updateFilterCallback
+    );
+
+    const formatCustomFiltersForAgGrid = useCallback(
+        (filters: FilterSelectorType[]) => {
+            const agGrifFilterModel: Record<string, any> = {};
```
klesaulnier, line 29:
```suggestion
            const agGridFilterModel: Record<string, any> = {};
```

klesaulnier: I think we should change "Record<string, any>" to either : 
- "any" -> it's the type expected "setFilterModel"
- "Record<string, FilterType>" with FilterType a new TypeScript interface with "filterType", "type" and "filter" properties

src/components/spreadsheet/table-wrapper.js
```diff
@@ -202,6 +206,36 @@ const TableWrapper = (props) => {
...
+
+    // This is a dummy callback to avoid the warning of missing dependencies and infinite loop
+    const updateFilterCallback = useCallback(() => {}, []);
+
+    const { updateFilter, filterSelector, initFilters } =
+        useAggridRowFilterImpl(
+            gridRef,
+            equipementFiltersSelectorKeys,
+            updateFilterCallback
+        );
```
klesaulnier, line 237: I think it would be cleaner to change "useAggridRowFilter" definition and make "updateFilterCallback" not mandatory

you could then remove "updateFilterCallback" from "useAggridRowFilterImpl" props

src/components/spreadsheet/table-wrapper.js
```diff
@@ -258,12 +316,18 @@ const TableWrapper = (props) => {
...
+    }, [tabIndex, initFilters, defaultSortColKey, initSort]);
+
     const getRows = useCallback(() => {
         if (props.disabled || !equipments) {
             return [];
         }
+        let rowData = [...equipments];
 
-        return equipments;
+        return rowData;
```
klesaulnier, line 330: do we need this change ?
this could trigger unnecessary rerenders : "[...equipments]"

klesaulnier (changes requested):

EstherDarkish (commented):
Got some warning about missing translations in the console :
`Error: [@formatjs/intl Error MISSING_TRANSLATION] Missing message: "grid.ariaFilterValue" for locale "fr", using default message (Filter Value) as fallback.`

(maybe this one is not the only one)

src/hooks/use-aggrid-row-filter-impl.ts
achour94: `setFilterModel = (
    model: FilterModel | null
) => void;

interface FilterModel {
  [colId: string]: any;
}`
this is from the documentation of AgGrid https://www.ag-grid.com/react-data-grid/filter-api/#get--set-all-filter-models

src/components/spreadsheet/table-wrapper.js
achour94: It's already not mandatory, i deleted the dummy function, we don't need it anymore

src/components/spreadsheet/table-wrapper.js
achour94: My bad, it's a change from the previous code when i did the filtering manually

src/hooks/use-aggrid-sort-impl.ts
achour94: Yes, just to avoid eslint error

src/hooks/use-aggrid-row-filter-impl.ts
klesaulnier: I see, when you check "gridApi.d.ts", it's not the case 
`setFilterModel(model: any): void;`

If it's documented like this, let's keep it this way

klesaulnier (approved):
Code: OK

Meklo (changes requested):
I noticed there's an infinite rerender loop by adding a debug message in TableWrapper's body

src/components/spreadsheet/utils/config-tables.js
```diff
@@ -91,6 +96,29 @@ export const DEFAULT_SORT_ORDER = 'asc';
```
Meklo, line 96: You can remove DEFAULT_SORT_ORDER (which is located just above this line but I can't comment it) now that it is no longer in use

Meklo (approved):

PR #1917, opened 2024-02-12
Dynamically created translation keys are now in separate files.

EstherDarkish:


src/translations/dynamic/csv-locale-en.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * Copyright (c) 2024, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+const csv_locale_en = {
+    // Used in the tabular creation and modification
+    'TabularCreationSkeletonComment/GENERATOR':
```
FranckLecuyer, line 10: '.' instead of '/' as separator in all keyx

FranckLecuyer (changes requested):

src/translations/dynamic/csv-locale-en.js
EstherDarkish: Done

FranckLecuyer (approved):

src/translations/dynamic/network-modifications-locale-en.js
```diff
@@ -0,0 +1,92 @@
...
+    'network_modifications.GENERATOR_SCALING':
+        'Generator scaling {computedLabel}',
+    'network_modifications.VSC_CREATION': 'Creating vsc {computedLabel}',
+    'network_modifications.GROOVY_SCRIPT': 'Modification by script',
+    'network_modifications.EQUIPMENT_ATTRIBUTE_MODIFICATION':
+        '{equipmentAttributeName, select, open {{equipmentAttributeValue, select, true {Open {computedLabel}} other {Close {computedLabel}}}} other {Equipment modification {computedLabel}}}',
+    'network_modifications.creatingModification': 'Creating modification ...',
+    'network_modifications.deletingModification': 'Deleting modification ...',
+    'network_modifications.updatingModification': 'Updating modification ...',
+    'network_modifications.modifications': 'Updating modification list ...',
```
dbraquart, line 64: the 4 above are not "dynamic"
(we can leave them, they have the same prefix)

dbraquart (approved):
code: ok
tests: ok

src/translations/dynamic/network-modifications-locale-en.js
EstherDarkish: Yes, separating them is confusing. Maybe if we reorganize the way we name our keys, it would make sense and we could move them.

PR #1920, opened 2024-02-14
Add modal and methods to save and load parameters

lilian-houdelet:


AAJELLAL (changes requested):
Initial values are always sent regardless of what is entered by the user.

AAJELLAL (approved):
Tests and code are ok.

jamal-khey (approved):
Code and tests Ok

PR #1927, opened 2024-02-19
Homogenize batteries and generators attributes

souissimai:


src/components/dialogs/network-modifications/battery/creation/battery-creation-dialog.js
```diff
@@ -129,9 +128,9 @@ const BatteryCreationDialog = ({
                     ? 'MINMAX'
                     : 'CURVE',
                 minimumReactivePower:
-                    battery?.minMaxReactiveLimits?.minimumReactivePower ?? null,
```
Mathieu-Deharbe, line 132: Why don’t you change the form fields like `minimumReactivePower` just above or maximumReactivePower below while you’re at it? I think it would make the code clearer.

src/components/utils/field-constants.js
```diff
@@ -73,8 +77,10 @@ export const REACTIVE_CAPABILITY_CURVE_CHOICE = 'reactiveCapabilityCurveChoice';
 export const REACTIVE_CAPABILITY_CURVE_TABLE = 'reactiveCapabilityCurveTable';
 export const MINIMUM_REACTIVE_POWER = 'minimumReactivePower';
 export const MAXIMUM_REACTIVE_POWER = 'maximumReactivePower';
-export const Q_MAX_P = 'qmaxP';
-export const Q_MIN_P = 'qminP';
+export const MIN_Q = 'minQ';
+export const MAX_Q = 'maxQ';
+export const Q_MAX_P = 'maxQ';
+export const Q_MIN_P = 'minQ';
```
Mathieu-Deharbe, line 83: Is it really necessary to keep Q_MAX_P and Q_MIN_P ?

Mathieu-Deharbe:
I think there is a problem in the modifying generator interface. If I set a value to "rated nominal power" : 

entrer 10

Then validate and reopen : 

reouvre

The value disappeared.

There seem to have the same problem with QminPmin, QmaxPmin, QminPmax and QmaxPmax. It looks like those values are never loaded. And this in true in generator modification and generator creation.

This a new bug (not this problem on OPF).

Mathieu-Deharbe (approved):

SlimaneAmar (approved):

PR #1937, opened 2024-02-21
Fix map VLs filter on VL update

SlimaneAmar:


src/components/network/map-equipments.js
```diff
@@ -198,7 +198,7 @@ export default class MapEquipments {
     }
 
     completeSubstationsInfos(equipementsToIndex) {
-        const nominalVoltagesSet = new Set(this.nominalVoltages);
```
Meklo, line 201: It was specifically added in this PR #1470 to fix a bug apparently, why the removal ?

SlimaneAmar:
No longer relevant

PR #1940, opened 2024-02-22
Add properties to all equipments

AAJELLAL:


TheMaskedTurtle (changes requested):
It does not work for voltage levels and there is no key duplication check on shunt compensators, but I don't know why ^^

AAJELLAL:
> It does not work for voltage levels and there is no key duplication check on shunt compensators, but I don't know why ^^

Voltage level fixed. The check of key duplication is OK for shunt compensators(creation and modification).

shunt_creation
shunt_modification

jamal-khey (approved):

PR #1949, opened 2024-02-26
Add country fields to all spreadsheet equipments

Meklo:


FranckLecuyer (approved):
Just one remark : 
I don't know if it has really a sense to show country1, country2, country3 for 2WT and 3WT, as usually these equipments are in one country only ...
see with the PO

PR #1952, opened 2024-02-27
homogenize danglingline attributes

souissimai:


src/components/spreadsheet/utils/config-tables.js
```diff
@@ -3265,14 +3265,14 @@ export const TABLES_DEFINITIONS = {
             },
             {
                 id: 'NominalV',
-                field: 'nominalVoltage',
+                field: 'nominalV',
```
Mathieu-Deharbe, line 3268: In my PR I decided to use the NOMINAL_V exported const. It doesn't make much difference but I think using a constant is better.

Mathieu-Deharbe (approved):

AAJELLAL (approved):

PR #1962, opened 2024-02-28
Homogenize static var compensator attributes to iidm format

Mathieu-Deharbe:
I only renamed nominalVoltage to nominalV.

I think all the others members are fine or don't exist in powsybl.


thangqp (approved):
Code OK
Test OK

PR #1964, opened 2024-02-28
Dynamic Simulation - result tab Status shown by aggrid

thangqp:
- This PR replaces the use of VirtualizedTable by Aggrid for tab Status
- Use React.memo() HOC for all sub-tabs under the result tab to avoid unnecessary re-rendering when switching between tabs

**Demo**

aggrid_status


jamal-khey (approved):
Code and Test Ok

Mathieu-Deharbe:
The header displaying only "status" doesn't look very useful. We are already in the header tab. Will it be enriched soon and therefore useful?

We already discussed it but I don't remember your answer : all the other computation result occupy the whole width of the csreen but dynamic simulation doesn't. Why not ? : 

statut

thangqp:
> The header displaying only "status" doesn't look very useful. We are already in the header tab. Will it be enriched soon and therefore useful?
> 
> We already discussed it but I don't remember your answer : all the other computation result occupy the whole width of the csreen but dynamic simulation doesn't. Why not ? :
> 
> statut
> 
> Extra small remark : the green round is not aligned with the text. I know it was there before but I had to say it.

1. Not only status, in the near future at least this information in result will be shown in the table, see https://github.com/powsybl/powsybl-core/blob/main/dynamic-simulation/dynamic-simulation-api/src/main/java/com/powsybl/dynamicsimulation/DynamicSimulationResult.java#L30
2. In a precedent PR, the cells do not cover the whole larger of table because we dont want the icon of filter is at the most right of the table. It not practice for users.
3.  Thanks for the remark, I will see if possible correct it quickly...

PR #1965, opened 2024-02-29
Fix code after security-analysis-server homogenization

klesaulnier:


achour94 (approved):

src/components/results/securityanalysis/security-analysis-result-tab.tsx
```diff
@@ -102,8 +102,25 @@ export const SecurityAnalysisResultTab: FunctionComponent<
         sortWay: SORT_WAYS.asc,
     });
 
+    const resultType = useMemo(() => {
+        if (tabIndex === N_RESULTS_TAB_INDEX) {
+            return RESULT_TYPE.N;
+        }
+
+        if (nmkType === NMK_TYPE.CONSTRAINTS_FROM_CONTINGENCIES) {
```
flomillot, line 110:
```suggestion
        else if (nmkType === NMK_TYPE.CONSTRAINTS_FROM_CONTINGENCIES) {
```
To be coherent with the else above

src/components/results/securityanalysis/security-analysis-result-tab.tsx
```diff
@@ -102,8 +102,25 @@ export const SecurityAnalysisResultTab: FunctionComponent<
         sortWay: SORT_WAYS.asc,
     });
 
+    const resultType = useMemo(() => {
```
flomillot, line 105: Why create useMemo for this ?

flomillot: Ah ok it already exist. I think it's counterproductive and useless. Or I missed something ?

src/components/results/securityanalysis/security-analysis-result-tab.tsx
```diff
@@ -102,8 +102,25 @@ export const SecurityAnalysisResultTab: FunctionComponent<
...
+        }
+
+        if (nmkType === NMK_TYPE.CONSTRAINTS_FROM_CONTINGENCIES) {
+            return RESULT_TYPE.NMK_CONTINGENCIES;
+        } else {
+            return RESULT_TYPE.NMK_LIMIT_VIOLATIONS;
+        }
+    }, [tabIndex, nmkType]);
+
+    const columnToFieldMapping = useMemo(
```
flomillot, line 117: Same, IMO you should directly call getColumnToFieldMapping, no need memo for this

src/components/results/securityanalysis/security-analysis-result-utils.ts
```diff
@@ -594,10 +594,26 @@ export const SECURITY_ANALYSIS_RESULT_INVALIDATIONS = [
     'securityAnalysisResult',
 ];
 
-export const FROM_COLUMN_TO_FIELD: Record<string, string> = {
-    subjectId: 'subjectId',
-    contingencyId: 'contingencyId',
-    status: 'status',
+export enum RESULT_TYPE {
```
flomillot, line 597: You didn't change it, put it back where it was ?

klesaulnier: well, since it's javascript, enum declarations need to be done before enum uses
it's either this, or I need to move "getColumnToFieldMapping" below
It doesn't matter to me though

src/components/results/securityanalysis/security-analysis-result-tab.tsx
klesaulnier: fixed

src/components/results/securityanalysis/security-analysis-result-tab.tsx
klesaulnier: where does it exist ? I did not see it
I agree we don't have to use "useMemo" since it's only a boolean calculation though, if that was your point

src/components/results/securityanalysis/security-analysis-result-tab.tsx
klesaulnier: fixed

flomillot: No, I mean you should directly call getColumnToFieldMapping(resultType), otherwise with this fetchSecurityAnalysisResultWithQueryParams will be triggered at every rerender

src/components/results/securityanalysis/security-analysis-result-utils.ts
flomillot: By setting getColumnToFieldMapping below the previous RESULT_TYPE, we can keep the same position and the original git history

src/components/results/securityanalysis/security-analysis-result-tab.tsx
flomillot: You moved the code from line 237, no ?
Yeah nevermind there is useMemo and useCallback everywhere in that file anyway

src/components/results/securityanalysis/security-analysis-result-tab.tsx
klesaulnier: it's true, I fixed it too fast
I was memoizing it because it was used twice in the file, it was to avoid calling the method several times with the same exact params

But alright I'm changing it

src/components/results/securityanalysis/security-analysis-result-utils.ts
klesaulnier: fixed

flomillot (approved):

PR #1966, opened 2024-02-29
Update to commons-ui v0.55.0 and add possibility to replace chip by clicking on them in autocomplete

YenguiSeddik:
It seems good for me, but when selecting multiple filters where 2 of them are already selected, an error message is show "This element is already used" but maybe we should add the name of filters and adapt the message (check with PO)


souissimai:
Click on a filter chip.
    Reopen the filter selection window.
    Navigate to the folder containing the filter which is selected.
    Select one more filter and keep the selected filter on the list.
    Click on the "Replace" button.

And the observed result is that the second filter gets added, but the previously selected one is not present on the list of chips, and an error message stating "element already used" appears.

src/components/utils/rhf-inputs/directory-items-input.tsx
```diff
@@ -58,6 +60,10 @@ const DirectoryItemsInput: FunctionComponent<DirectoryItemsInputProps> = ({
     const types = useMemo(() => [elementType], [elementType]);
     const [directoryItemSelectorOpen, setDirectoryItemSelectorOpen] =
         useState(false);
+
+    const [selectedChip, setSelectedChip] = useState<any[] | undefined>([]);
+    const [expandedNodes, setExpandedNodes] = useState<any[] | undefined>([]);
```
anistouri, line 65: I don't think they could be undefined
Replace 'any' with a specific type if possible

src/components/utils/rhf-inputs/directory-items-input.tsx
```diff
@@ -76,12 +82,23 @@ const DirectoryItemsInput: FunctionComponent<DirectoryItemsInputProps> = ({
...
+                });
+            }
+
             values.forEach((value) => {
                 const { icon, children, ...otherElementAttributes } = value;
 
                 // Check if the element is already present
                 if (
-                    getValues(name).find(
+                    allChips.find(
```
anistouri, line 101: If we select several elements while keeping the element we clicked on, we will get the error message 'This element is already used'. Is this intended?

souissimai (approved):

anistouri (approved):

PR #1967, opened 2024-02-29
Show the calculated voltage profile in the voltage init results

FranckLecuyer:


Meklo (approved):

src/components/voltage-init-result.jsx
```diff
@@ -375,7 +420,13 @@ const VoltageInitResult = ({ result, status, tabIndex, setTabIndex }) => {
                     {result &&
                         tabIndex === 1 &&
                         renderReactiveSlacksTable(result.reactiveSlacks)}
-                    {tabIndex === 2 && renderReportViewer()}
+                    {result &&
+                        tabIndex === 2 &&
+                        enableDeveloperMode &&
+                        renderBusVoltagesTable(result.busVoltages)}
+                    {((tabIndex === 3 && enableDeveloperMode) ||
+                        (tabIndex === 2 && !enableDeveloperMode)) &&
```
Meklo, line 428: Not in the scope of this feature but I think we ought to refactor sub tabs management in order to handle them dynamically as it is done for the main result tabs

PR #1977, opened 2024-03-06
Refactor CheckboxList component

sBouzols:
Make it a complete controlled component and remove complex useEffects which were there to synchronize the internal `checked` state and an external one updated with the `onChecked` callback.
Use a Set to avoid duplicates at insert, not inside this component.


src/components/dialogs/contingency-list-selector.jsx
```diff
@@ -146,30 +141,25 @@ const ContingencyListSelector = (props) => {
     };
 
     const removeFromFavorite = (toRemove) => {
-        const toDelete = new Set(toRemove);
```
achour94, line 149: The Set performs existence checks more efficiently than Array.includes, i think this is the purpose of doing this

src/components/dialogs/contingency-list-selector.jsx
```diff
@@ -146,30 +141,25 @@ const ContingencyListSelector = (props) => {
...
         saveFavorite(
             contingencyList
                 .map((e) => e.id)
-                .filter((item) => !toDelete.has(item))
+                .filter((item) => !toRemove.includes(item))
```
achour94, line 147: why you dont filter directly like this contingencyList.filter((item) => !toRemoveSet.includes(item.id))

src/components/dialogs/contingency-list-selector.jsx
```diff
@@ -146,30 +141,25 @@ const ContingencyListSelector = (props) => {
...
         );
-        const newChecked = checkedContingencyListUuids.filter(
-            (item) => !toDelete.has(item)
+
+        setCheckedContingencyList((oldCheked) =>
+            oldCheked.filter((item) => !toRemove.includes(item))
```
achour94, line 151: you are checking if an array of ids includes an object (item)

src/components/dialogs/contingency-list-selector.jsx
```diff
@@ -146,30 +141,25 @@ const ContingencyListSelector = (props) => {
     };
 
     const removeFromFavorite = (toRemove) => {
```
achour94, line 143: this function receive sometimes an array of ids, and another time an array of objects, for concistency with addFavorites use the same pattern

src/components/dialogs/contingency-list-selector.jsx
sBouzols: Those are really litlle arrays, I wanted to keep it simple but ths is right. will revert

src/components/utils/checkbox-list.jsx
```diff
@@ -5,74 +5,41 @@
...
     values,
     onChecked,
-    initialSelection,
-    itemComparator = (a, b) => a.id === b.id,
+    checkedValues,
+    itemComparator = defaultComparator,
     ...props
 }) => {
-    const [checked, setChecked] = useState(new Set(initialSelection));
+    const valuesRef = useRef();
```
achour94, line 24: i don't see where this variable are used !?

src/components/utils/checkbox-list.jsx
```diff
@@ -5,74 +5,41 @@
...
-        [checked, onChecked, itemComparator]
-    );
-
-    useEffect(() => onChecked && onChecked(checked), [checked, onChecked]);
-
-    const isCheckboxInCheckedSet = (checkedSet, checkBoxToCheck) => {
-        return Array.from(checkedSet).some((element) =>
-            itemComparator(element, checkBoxToCheck)
-        );
+        onChecked([...newCheckedValues]);
```
achour94, line 42: newCheckedValues is already a new array , no need to spread it again

src/components/utils/checkbox-list.jsx
sBouzols: arf forget to clean this

src/components/dialogs/contingency-list-selector.jsx
sBouzols: fixed

src/components/dialogs/contingency-list-selector.jsx
sBouzols: Should be Array of Objects as arguments for `addFavorites` and `removeFromFavorites` now

src/components/dialogs/contingency-list-selector.jsx
sBouzols: Should be Array of Objects as arguments for addFavorites and removeFromFavorites now

src/components/dialogs/contingency-list-selector.jsx
sBouzols: saveFavorites expect only some ids, then we need to filter and map.

achour94 (approved):

thangqp (approved):

PR #1983, opened 2024-03-07
Refactor table custom header's sort mechanism to add children sorting

flomillot:
Based on AGGrid's sorting API, which give better compatibility and flexibility for future new features.
Add children sorting for SCA and SA.


Mathieu-Deharbe (approved):

src/components/results/sensitivity-analysis/paged-sensitivity-analysis-result.jsx
```diff
@@ -128,16 +128,18 @@ const PagedSensitivityAnalysisResult = ({
...
-
-        const sortSelector =
-            colKey && sortWay
-                ? {
-                      sortKeysWithWeightAndDirection: {
-                          [DATA_KEY_TO_SORT_KEY[colKey]]: sortWay,
-                      },
-                  }
-                : {};
+        const sortSelector = sortConfig?.length
```
klesaulnier, line 131: if sortConfig is an array, I don't think we need to check length, reduce will return default value if it's empty anyway
maybe it was a protection since it's not TS yet ?

src/hooks/use-aggrid-sort.ts
```diff
@@ -8,47 +8,50 @@
...
-
-export const getSortValue = (sortWay: number) => {
-    if (sortWay > 0) {
-        return 'asc';
-    } else {
-        return 'desc';
-    }
+    asc: 'asc' as const,
+    desc: 'desc' as const,
 };
```
klesaulnier, line 26: make an enum instead to make it usable in interfaces definition

src/services/study/short-circuit-analysis.js
```diff
@@ -121,8 +120,9 @@ export function fetchShortCircuitAnalysisPagedResults({
     urlSearchParams.append('page', page);
 
     if (sort) {
-        const { colKey, sortWay } = sort;
-        urlSearchParams.append('sort', `${colKey},${getSortValue(sortWay)}`);
+        sort.map((value) =>
+            urlSearchParams.append('sort', `${value.colId},${value.sort}`)
+        );
     }
```
klesaulnier, line 126: use "sort?." like other files instead of using "if" condition

klesaulnier (commented):
Test OK except for the loadflow result bug we've discussed 
- First accept to "Voltage Violations" tab is not working

src/components/results/sensitivity-analysis/paged-sensitivity-analysis-result.jsx
```diff
@@ -128,16 +128,18 @@ const PagedSensitivityAnalysisResult = ({
...
+            ? {
+                  sortKeysWithWeightAndDirection: sortConfig.reduce(
+                      (acc, value) => ({
+                          ...acc,
+                          [DATA_KEY_TO_SORT_KEY[value.colId]]:
+                              value.sort === 'desc' ? -1 : 1,
```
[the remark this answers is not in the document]
flomillot: Oh yeah I forgot to change but actually I can use SORT_WAYS now.
I will just change the custom grid sort part btw.

src/components/results/sensitivity-analysis/paged-sensitivity-analysis-result.jsx
flomillot: It's just to avoid adding `sortSelector` with empty `sortKeysWithWeightAndDirection` if the sort array is empty

src/hooks/use-aggrid-sort.ts
flomillot: But we can't use enum in JS code right ?

src/services/study/short-circuit-analysis.js
flomillot: Yeah I just wanted to change the minimum code possible

src/components/results/sensitivity-analysis/paged-sensitivity-analysis-result.jsx
klesaulnier: You're right, I did not see the difference at first

src/hooks/use-aggrid-sort.ts
klesaulnier: I think you can, ComputingType enum is used in "reducer.js"

flomillot: Ah OK great, my bad

klesaulnier (approved):

PR #1986, opened 2024-03-08
Fixed text overlay in settings panel

ne0ds:


EstherDarkish (commented):
There is still an issue with the bottom row, where the buttons are.

EstherDarkish:
Fixed it myself by encapsulating Flex parts. Can someone else review this ?

src/components/dialogs/parameters/load-flow-parameters.jsx
```diff
@@ -582,108 +589,132 @@ export const LoadFlowParameters = ({ parametersBackend }) => {
     // we must keep the line of the simulator selection visible during scrolling
     // only specifics parameters are dependents of simulator type
     return (
-        <>
-            <Grid sx={{ height: '100%' }} xl={6}>
-                <Grid
-                    container
-                    spacing={1}
+        <Grid item sx={{ height: '100%' }} xl={8} lg={10} md={12}>
```
thangqp, line 592: A grid item should be wrapped by a grid container, however at the place which uses this component, the grid is an item too, see https://github.com/gridsuite/gridstudy-app/blob/main/src/components/parameters-tabs.tsx#L493

In my opinion, this should be a container gird

EstherDarkish: This is already in a container, see parameters-tabs.tsx line 376

src/components/dialogs/parameters/load-flow-parameters.jsx
```diff
@@ -34,6 +34,7 @@ import { DirectoryItemSelector } from '@gridsuite/commons-ui';
 import { fetchLoadFlowParameters } from '../../../services/loadflow';
 import { fetchDirectoryContent, fetchRootFolders } from 'services/directory';
 import { fetchElementsMetadata } from 'services/explore';
+import { Box } from '@mui/system';
```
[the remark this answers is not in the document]
EstherDarkish: Thanks, fixed.

src/components/dialogs/parameters/load-flow-parameters.jsx
EstherDarkish: Note : we cannot put a container here, it breaks everything.
These whole pages have to be refactored someday because they are all over the place, but for now, this works.

src/components/dialogs/parameters/load-flow-parameters.jsx
```diff
@@ -588,108 +589,132 @@ export const LoadFlowParameters = ({ parametersBackend }) => {
...
+                            values={LoadFlowProviders}
+                            callback={updateLfProviderCallback}
                         />
+                    </Grid>
+                    <Grid
+                        container
+                        sx={mergeSx(styles.scrollableGrid, {
+                            maxHeight: '100%',
+                        })}
+                        key="lfParameters"
```
ne0ds, line 638: Key is useless

ne0ds:
image

There is a double scroll bar now => NOK

EstherDarkish (approved):

thangqp (approved):

PR #1987, opened 2024-03-11
Fix error handling and add reset function

jamal-khey:
This pull request fixes the error handling in the VscModificationDialog component and adds a reset function. The error handling was previously displaying an error message, but now it calls the reset function to reset the form data. This behavior match other modification


src/components/dialogs/network-modifications/vsc/modification/vsc-modification-dialog.tsx
```diff
@@ -228,10 +228,7 @@ const VscModificationDialog: React.FC<any> = ({
                     .catch((error) => {
                         setVcsToModify(null);
                         setDataFetchStatus(FetchStatus.FAILED);
-                        snackError({
-                            messageTxt: error.message,
-                            headerId: 'VscModificationError',
-                        });
+                        reset(emptyFormData);
```
anistouri, line 231: There's no need to reset the form every time we enter the catch block, since no data has been received. The form should only be reset if it already contains data. We can follow the approach used in other modification forms, where resetting is contingent on the presence of pre-existing data

anistouri, line 231:
```
setDataFetchStatus(FetchStatus.FAILED);
                        if (editData?.equipmentId !== equipmentId) {
                            setVcsToModify(null);
                            reset(emptyFormData);
                        }
```

anistouri (approved):

src/components/dialogs/network-modifications/vsc/modification/vsc-modification-dialog.tsx
jamal-khey: i think the reset is done on the others modification, for example `generator-modification-dialog.js:334` or l`oad-modification-dialog.js:159 `

PR #1988, opened 2024-03-11
Refactor ListItem usage replacing button API by ListItemButton component

sBouzols:
button and other `ButtonBase` props are deprecated over `ListItem` component. 
Use `ListItemButton` wrapper component instead


EstherDarkish (approved):
Code OK tests OK

PR #1990, opened 2024-03-11
Add filters on some columns of SA

flomillot:
Update filter data type to NUMBER for Limit column and add filters for CalculatedValue, Loading, Overload and LimitSide columns.


src/components/results/securityanalysis/security-analysis-result-utils.ts
```diff
@@ -369,10 +379,21 @@ export const securityAnalysisTableNmKContingenciesColumnsDefinition = (
             field: 'acceptableDuration',
             valueFormatter: (value: ValueFormatterParams) =>
                 convertDuration(value.data.acceptableDuration),
+            filterProps,
+            filterParams: {
+                filterDataType: FILTER_DATA_TYPES.NUMBER,
+                filterComparators: Object.values(FILTER_NUMBER_COMPARATORS),
+                parser: parseDuration,
```
achour94, line 386: This parser function returns always '0' , so the filter is not working on this column

src/components/results/securityanalysis/security-analysis-result-utils.ts
```diff
@@ -481,10 +512,21 @@ export const securityAnalysisTableNmKConstraintsColumnsDefinition = (
...
```
achour94, line 519: Same problem

achour94 (changes requested):

src/components/results/securityanalysis/security-analysis-result-utils.ts
flomillot: Thanks, bug fixed in #2003

achour94 (approved):

PR #1992, opened 2024-03-12
Change environnement variable

lilian-houdelet:


src/services/study/sensitivity-analysis.js
```diff
@@ -108,7 +108,9 @@ export function getSensitivityAnalysisParameters(studyUuid) {
 
 export function fetchSensitivityAnalysisParameters(parameterUuid) {
     console.info('get sensitivity analysis parameters');
-    const url = `${process.env.REACT_APP_API_GATEWAY}/sensitivity-analysis/v1/parameters/${parameterUuid}`;
+    const url = `${
+        import.meta.env.VITE_API_GATEWAY
```
AAJELLAL, line 112: You need to define this at the top of the file and use it here.
This way it can be reusable.
Search for "**import. meta. env. VITE_API_GATEWAY**" on other files as an  example

AAJELLAL (changes requested):

AAJELLAL (approved):

PR #1995, opened 2024-03-12
ESlint not block Vite build when in development mode

Tristan-WorkGH:
Update to:
* Set ESlint to not fail if warnings when in [dev mode](https://vitejs.dev/guide/env-and-mode#modes) (`npm start` serve)
* Set Vite to run ESlint directly after starting and not waiting to have the page asked to compile & check


.eslintrc.yml
```diff
@@ -0,0 +1,13 @@
+root: true
+
+ignorePatterns:
+    - "**/node_modules/**"
```
flomillot, line 4: Why ? node_modules already ignored

.eslintrc.yml
flomillot, line 1: Can't we use .js or .json instead ?

Tristan-WorkGH: yes, but found YAML more readable than JSON

vite.config.ts
```diff
@@ -31,23 +30,27 @@ const serverSettings = {
...
 };
 
-export default defineConfig({
+export default defineConfig((config) => ({
     plugins: [
         react(),
         eslint({
-            failOnWarning: true,
+            failOnWarning: config.mode !== 'development',
+            lintOnStart: config.command === 'serve',
```
flomillot, line 38: It's false by default no ?

vite.config.ts
```diff
@@ -31,23 +30,27 @@ const serverSettings = {
...
         }),
         svgr(), // works on every import with the pattern "**/*.svg?react"
         reactVirtualized(),
         tsconfigPaths(), // to resolve absolute path via tsconfig cf https://stackoverflow.com/a/68250175/5092999
     ],
+    resolve: {
+        preserveSymlinks: true, //else eslint don't detect "npm link" packages
```
flomillot, line 45: Why would we want that ? Which case was a problem ?

vite.config.ts
Tristan-WorkGH: yep

vite.config.ts
Tristan-WorkGH: when using `npm link` (to test commons-ui for example), we Vite call eslint with symlinks resolved and analyse the node_modules of the linked dependencies because it is outside of the node_modules of the project

.eslintrc.yml
flomillot: IMO we should stick to what we use for other config files

vite.config.ts
flomillot: Ok I didn't know what it was doing until I read the description of the US.
Should we just set it to `true`, why keeping it `false` when building ?

vite.config.ts
flomillot: Ok got it thanks

vite.config.ts
Tristan-WorkGH: I did that to not modify the behavior with other commands. But I haven't problems with that.

.eslintrc.yml
Tristan-WorkGH: I can, but we lost comments for example...

flomillot: Not in JS

Tristan-WorkGH: okay

vite.config.ts
Tristan-WorkGH: > Check all matching files on project startup, too slow, turn on discreetly.

maybe yes

.eslintrc.yml
Tristan-WorkGH: oh ! [comments is supported](https://eslint.org/docs/latest/use/configure/configuration-files#comments-in-configuration-files) (it's jsonc format)

flomillot (approved):

sBouzols (approved):

PR #2003, opened 2024-03-14
Fix duration filters of columns in custom headers

flomillot:


achour94 (changes requested):
`const handleFilterComparatorChange = (event) => {
        const newType = event.target.value;
        setSelectedFilterComparator(newType);
        debouncedUpdateFilter(field, {
            value: parser ? parser(selectedFilterData) : selectedFilterData,
            type: newType,
            dataType: filterDataType,
        });
    };`
we have to parse the value if we change the comparator too else we have an error

anistouri (approved):

PR #2018, opened 2024-03-20
add and handle tieLine in geo data

ghazwarhili:


SlimaneAmar (approved):

src/components/network/gs-map-equipments.js
```diff
@@ -58,7 +71,35 @@ export default class GSMapEquipments extends MapEquipments {
...
+                );
+            })
+            .catch((error) => {
+                console.error(error.message);
+                if (this.errHandler) {
+                    this.errHandler({
+                        messageTxt: error.message,
+                        headerId: 'MapEquipmentsLoadError',
+                    });
+                }
```
Mathieu-Deharbe, line 90: That part is copy-pasted 8 times. It could be a function.

src/redux/reducer.js
```diff
@@ -1057,6 +1061,8 @@ function getEquipmentTypeFromUpdateType(updateType) {
     switch (updateType) {
         case 'lines':
             return EQUIPMENT_TYPES.LINE;
+        case 'tieLines':
```
Mathieu-Deharbe, line 1064: What does this update type String refer to? To some data update type somewhere in the back? Shouldn’t it be referenced/commented somewhere?

Mathieu-Deharbe (approved):

PR #2021, opened 2024-03-21
Send actual data to shortcircuit limits values getters

Meklo:


TheMaskedTurtle (approved):
params.data?. maybe ? Might be safer

PR #2027, opened 2024-03-25
Add filter to boolean column in spreadsheet

YenguiSeddik:


src/components/custom-aggrid/custom-aggrid-header-utils.ts
```diff
@@ -20,7 +20,11 @@ export const makeAgGridCustomHeaderColumn = ({
     const { filterDataType, filterEnums = {} } = filterParams || {};
 
     const customFilterOptions =
-        filterDataType === FILTER_DATA_TYPES.TEXT ? filterEnums[field] : [];
+        filterDataType === FILTER_DATA_TYPES.TEXT
+            ? filterEnums[field]
+            : filterDataType === FILTER_DATA_TYPES.BOOLEAN
+            ? ['true', 'false']
```
TheMaskedTurtle, line 26: Then we don't list the actual values in the data, is it ok from a functional point of view ?

src/components/custom-aggrid/custom-aggrid-header.jsx
```diff
@@ -322,7 +326,7 @@ const CustomHeaderComponent = ({
                 >
                     {isAutoCompleteFilter ? (
                         <Autocomplete
```
achour94, line 328: Using autocomplete for a boolean filter is not suitable. It would be better to create a new component, such as a Select dropdown, to handle the boolean filter type.

src/components/spreadsheet/utils/config-tables.js
```diff
@@ -126,6 +126,35 @@ const defaultEnumFilterConfig = {
...
+                    // if cellValue is undefined or null, we return true if the filter selected is false
+                    return filterValue === 'false';
+                },
+            },
+        ],
+    },
+    customFilterParams: {
+        filterDataType: FILTER_DATA_TYPES.BOOLEAN,
+    },
+    isBoolean: true,
```
TheMaskedTurtle, line 155: I think this is not used anymore

src/components/spreadsheet/utils/config-tables.js
```diff
@@ -126,6 +126,35 @@ const defaultEnumFilterConfig = {
...
 
+/**
+ * Default configuration for a boolean filter
+ */
+const defaultBooleanFilterConfig = {
+    filter: 'agTextColumnFilter',
+    agGridFilterParams: {
+        filterOptions: [
+            {
+                displayKey: 'customInRange',
```
achour94, line 137: Choose a better descriptive displayKey and displayKey like 'booleanMatches' for example

src/components/spreadsheet/utils/config-tables.js
achour94, line 155: This parameter is not used anywhere

src/components/custom-aggrid/custom-aggrid-header-utils.ts
achour94: i think we should not handle a boolean filter as an enum, for a boolean column its always yes or no

src/components/spreadsheet/utils/config-tables.js
```diff
@@ -126,6 +126,35 @@ const defaultEnumFilterConfig = {
...
+                displayName: 'customInRange',
+                predicate: ([filterValue], cellValue) => {
+                    // We receive here the filter boolean values as a string (filterValue)
+                    // we check if the cellValue is not null neither undefined
+                    if (cellValue) {
+                        return filterValue === cellValue.toString();
+                    }
+
+                    // if cellValue is undefined or null, we return true if the filter selected is false
+                    return filterValue === 'false';
```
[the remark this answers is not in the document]
achour94: In certain instances, the value of a column may be null or undefined. Consequently, these rows are only displayed when 'false' is selected in the filter.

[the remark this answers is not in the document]
TheMaskedTurtle: Yes but still, this is quite an unexpected behavior. Why would we display null values when we select false ? Why not when we select true, both or none ? Please confirm that behavior with a PO.

src/components/custom-aggrid/custom-aggrid-header.jsx
```diff
@@ -95,14 +97,15 @@ const CustomHeaderComponent = ({
...
     const intl = useIntl();
 
     const [filterAnchorElement, setFilterAnchorElement] = useState(null);
     const [isHoveringColumnHeader, setIsHoveringColumnHeader] = useState(false);
     const [selectedFilterComparator, setSelectedFilterComparator] =
         useState('');
+    const [selectFilterBoolean, setSelectFilterBoolean] = useState('');
     const [selectedFilterData, setSelectedFilterData] = useState(undefined);
 
     const shouldDisplayFilterIcon =
```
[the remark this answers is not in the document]
YenguiSeddik: selectedFilterData is initialized with undefined which will cause an error with select component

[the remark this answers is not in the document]
achour94: you can do this value={selectedFilterData || ''}

achour94 (approved):

src/components/custom-aggrid/custom-aggrid-header-utils.ts
```diff
@@ -85,3 +85,9 @@ export const mapFieldsToColumnsFilter = (
         column: columnToFieldMapping[filter.column],
     }));
 };
+
+export const BOOLEAN_FILTER_VALUES = {
+    TRUE: 'true',
+    FALSE: 'false',
+    UNDEFINED: 'undefinedValue',
+}
```
TheMaskedTurtle: With Typescript you can define an `enum`

src/components/custom-aggrid/custom-aggrid-header.jsx
```diff
@@ -28,6 +27,7 @@ import {
 } from './custom-aggrid-header.type';
 import { mergeSx } from '../utils/functions';
 import { useLocalizedCountries } from 'components/utils/localized-countries-hook';
+import CustomAggridBooleanFilter from './custom-aggrid-filters/custom-aggrid-boolean-filter.tsx';
```
TheMaskedTurtle, line 30: You can remove .tsx

src/components/spreadsheet/utils/config-tables.js
```diff
@@ -46,6 +46,7 @@ import {
 import { NOMINAL_V } from '../../utils/field-constants';
 import CountryCellRenderer from '../country-cell-render';
 import EnumCellRenderer from '../enum-cell-renderer';
+import { BOOLEAN_FILTER_VALUES } from 'components/custom-aggrid/custom-aggrid-header-utils.ts';
```
TheMaskedTurtle, line 49: You can remove .ts

src/components/custom-aggrid/custom-aggrid-header-utils.ts
```diff
@@ -86,8 +86,8 @@ export const mapFieldsToColumnsFilter = (
     }));
 };
 
-export const BOOLEAN_FILTER_VALUES = {
-    TRUE: 'true',
-    FALSE: 'false',
-    UNDEFINED: 'undefinedValue',
-};
+export enum BOOLEAN_FILTER_VALUES {
```
TheMaskedTurtle, line 89: Can you  rename BooleanFilterValue, case would be more appropriate ?

TheMaskedTurtle (approved):

PR #2028, opened 2024-03-25
feat(voltage-init-parameters): automatically apply modifications according to parameter

TheMaskedTurtle:
Add a new Tab for general parameters in voltage init parameters


src/components/dialogs/parameters/voltageinit/use-get-voltage-init-parameters.tsx
```diff
@@ -0,0 +1,45 @@
+import { VoltageInitParam } from './voltage-init-utils';
```
FranckLecuyer, line 1: Missing license header

src/components/dialogs/parameters/voltageinit/voltage-init-parameters-form.tsx
```diff
@@ -0,0 +1,114 @@
+import yup from '../../../utils/yup-config';
```
FranckLecuyer, line 1: Missing license header

FranckLecuyer (changes requested):

FranckLecuyer (changes requested):
Tests: OK

src/components/dialogs/parameters/voltageinit/use-get-voltage-init-parameters.tsx
TheMaskedTurtle: Fixed

FranckLecuyer (approved):

PR #2036, opened 2024-03-28
Announcements

flomillot:
Add the possibility to see announcements from admins.
Use the hook created in commons-ui https://github.com/gridsuite/commons-ui/pull/405
From https://github.com/gridsuite/gridstudy-app/pull/1902

### DIFF
--- a/src/services/config-notification.js
+++ b/src/services/config-notification.js
@@ -29,3 +29,8 @@ export function connectNotificationsWsUpdateConfig() {
     };
     return reconnectingWebSocket;
 }
+
+export function getAnnouncementsWsUrl() {
+    const webSocketBaseUrl = getWsBase();
+    return webSocketBaseUrl + PREFIX_CONFIG_NOTIFICATION_WS + '/global';
+}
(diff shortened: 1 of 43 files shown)


src/services/config-notification.js
```diff
@@ -29,3 +29,8 @@ export function connectNotificationsWsUpdateConfig() {
     };
     return reconnectingWebSocket;
 }
+
+export function getAnnouncementsWsUrl() {
+    const webSocketBaseUrl = getWsBase();
+    return webSocketBaseUrl + PREFIX_CONFIG_NOTIFICATION_WS + '/global';
```
TheMaskedTurtle, line 35: Could be inlined

TheMaskedTurtle (approved):

src/services/config-notification.js
flomillot: I know I just kept the same style as above
https://github.com/gridsuite/gridstudy-app/blob/7112ee75c9faae09e23d3194be899615ef10145c/src/services/config-notification.js#L15-L20

PR #2045, opened 2024-04-03
Resolve legacy peers

Tristan-WorkGH:
Resolve the legacy-peer-deps problem which is caused currently by @gridsuite/commons-ui < 0.53.0

Keep versions of eslint and eslint-config-react-app aligned to gridsuite/commons-ui#350.

Also remove react-json-view which is not used and depends on archive "flux" library, which depend to React 17.


Tristan-WorkGH:
Is techleads ok that ZLib is compatible with MLP2 ?  
<https://choosealicense.com/licenses/zlib/>

SlimaneAmar (approved):

klesaulnier (approved):

PR #2048, opened 2024-04-04
Call a new endpoint to get the list of nominalVoltages used in a netw…

EstherDarkish:
…ork instead of using the mapEquipments cache.


thangqp (approved):
Code OK
Test OK

PR #2049, opened 2024-04-05
Add spreadsheet for tie lines.

FranckLecuyer:


AAJELLAL (changes requested):
The filter is not working for any of the columns and I get a console error each time I try to use the filter. 
The rest is OK.

FranckLecuyer:
> The filter is not working for any of the columns and I get a console error each time I try to use the filter. The rest is OK.

Fixed

AAJELLAL (approved):
Code and tests are OK

PR #2057, opened 2024-04-11
Fix limit type in short circuit one bus result

ghazwarhili:


src/components/results/shortcircuit/shortcircuit-analysis-one-bus-result.tsx
```diff
@@ -54,7 +54,7 @@ export const ShortCircuitAnalysisOneBusResult: FunctionComponent<
             studyUuid,
             currentNodeUuid: currentNode?.id,
             type: ShortCircuitAnalysisType.ONE_BUS,
-            mode: 'BASIC',
+            mode: 'FULL',
```
etiennehomer, line 57: You can remove the mode. Default mode is FULL in study-server

etiennehomer (changes requested):

etiennehomer (approved):

PR #2059, opened 2024-04-12
fix: check if user header is present in notif before filtering

AbdelHedhili:


src/components/study-container.jsx
```diff
@@ -255,7 +255,7 @@ export function StudyContainer({ view, onChangeTab }) {
             const updateTypeHeader = eventData.headers[UPDATE_TYPE_HEADER];
             const errorMessage = eventData.headers[ERROR_HEADER];
             const userId = eventData.headers[USER_HEADER];
-            if (userId !== userName) {
+            if (userId != null && userId !== userName) {
```
dbraquart, line 258: With your new condition, every user will have a message if the user is not set (like a build failure). I think this is not something we always want.

dbraquart (approved):
code: ok
tests: ok (this fixes the error test case)
remark: the errorMessage is not limited to the user that has ran the built

src/components/study-container.jsx
AbdelHedhili: At least for now, it is something we always want (always being this one specific case because no other exists yet) or am I missing something ?

PR #2062, opened 2024-04-16
Remove useless file, move html tag

flomillot:


Tristan-WorkGH (approved):
Okay, but ask a TL if it's okay to completely remove `src/setupTests.js`.

PR #2066, opened 2024-04-18
Refactor elements endpoint

jamal-khey:
This pull request refactors the elements endpoint to include additional parameters in the URL search params. Previously, the parameters that are 'specific' to an equipement type are mixed with generic params for all equipements


Mathieu-Deharbe (approved):

src/services/study/network.js
```diff
@@ -192,13 +192,13 @@ export function fetchNetworkElementInfos(
     }
     urlSearchParams.append('elementType', elementType);
     urlSearchParams.append('infoType', infoType);
-    const additionalParams = new Map();
+    const optionalParams = new Map();
```
Mathieu-Deharbe: Personnaly I prefered "additionalParams". 'inUpstreamBuiltParentNode' by example is an optional param but is not inside `optionalParameters'. It is a contradiction.

But it is not a big deal.

jamal-khey: name choosen by TL :)

PR #2072, opened 2024-04-25
Hiding of omnibus section in VL form

flomillot:
Refactor coupling omnibus form to display only when bus bar or section > 1


thangqp (approved):
Code OK, Test OK

Mathieu-Deharbe (approved):
I know this is not the subject of this US but when the user adds a coupler and then decides to add an other section count, the previously entered coupler is removed and has to be reentered.

I don't think this is very user friendly. Maybe it should be discussed with the PO.

ayolab (approved):

flomillot:
> I know this is not the subject of this US but when the user adds a coupler and then decides to add an other section count, the previously entered coupler is removed and has to be reentered.
> 
> I don't think this is very user friendly. Maybe it should be discussed with the PO.

Problem already present, I copy the comment in the ticket

PR #2073, opened 2024-04-25
Add free properties in HVDC forms.

flomillot:
Also extract anf factorize getConcatenatedProperties.

### DIFF
--- a/src/components/dialogs/network-modifications/common/properties/property-utils.ts
+++ b/src/components/dialogs/network-modifications/common/properties/property-utils.ts
@@ -13,8 +13,8 @@ import {
     PREVIOUS_VALUE,
     VALUE,
 } from 'components/utils/field-constants';
-import { fetchAppsAndUrls } from '../../../../../services/utils';
 import { isBlankOrEmpty } from 'components/utils/validation-functions';
+import { fetchStudyMetadata } from 'services/metadata';
 
 export type Property = {
     [NAME]: string;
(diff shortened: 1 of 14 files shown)


Mathieu-Deharbe (changes requested):
When an HVDC is loaded in the HVDC modification interface, the contained properties are not loaded (or, more likely, loaded but not displayed).

When you modify a battery which has some properties, they are loaded and displayed in the additional property part : 

modif batterie

This is not the case for this HVDV (that I created in the previous node, giving it a property that can be observed in the spreadsheet) : 

modif HVDC

---------------------------
Second strange problem but probably not because of your dev : if I create a HVDC in a node, I can load it in the HVDC modification interface. But I can't find and load it in the HVDC creation interface. It is simply not in the list.

flomillot:
> When an HVDC is loaded in the HVDC modification interface, the contained properties are not loaded (or, more likely, loaded but not displayed).
> 
> When you modify a battery which has some properties, they are loaded and displayed in the additional property part :
> 
> modif batterie
> 
> This is not the case for this HVDC (that I created in the previous node, giving it a property that can be observed in the spreadsheet) :
> 
> modif HVDC
> 
> Second strange problem but probably not because of your dev : if I create a HVDC in a node, I can load it in the HVDC **modification** interface. But I can't find and load it in the HVDC **creation** interface. It is simply not in the list.

Actually it was not loaded because I had to change network-map-server also : https://github.com/gridsuite/network-map-server/pull/187

src/components/dialogs/network-modifications/common/properties/property-utils.ts
```diff
@@ -120,6 +120,18 @@ export const copyEquipmentPropertiesForCreation = (
     };
 };
 
+export function getConcatenatedProperties(
+    equipment: any,
```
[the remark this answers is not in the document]
flomillot: Yes you're right, thanks

Mathieu-Deharbe (approved):
OK for me.

I am going to report the small loading bug on slack.

src/components/dialogs/network-modifications/vsc/hvdc-line-pane/vsc-hvdc-line-pane.tsx
```diff
@@ -220,6 +221,8 @@ const VscHvdcLinePane: FunctionComponent<VscHvdcLinePaneProps> = ({
                 {gridItem(droopField, 6)}
                 {gridItem(p0Field, 6)}
             </Grid>
+
+            <PropertiesForm isModification={isEquipementModification} />
```
ghazwarhili, line 225: networkElementType ?

flomillot: Not for HVDC, that's why I changed the code of PropertiesForm

ghazwarhili (approved):

PR #2076, opened 2024-04-30
clean enum ComputingType

dbraquart:
- remove one of the 3 "SHORTCIRCUIT" computations
- use same enum as ComputationType from study-server (just for clarity)


src/components/results/use-results-tab.ts
anistouri, line 1: You can alse change line 17 and 23 so that it's consistent everywhere

dbraquart: yes, done

anistouri (approved):

PR #2077, opened 2024-05-02
Add  history to equipment search

klesaulnier:


src/redux/local-storage/search-equipment-history.tsx
```diff
@@ -0,0 +1,95 @@
...
+    }
+    try {
+        const currentHistory = JSON.parse(currentHistoryJson);
+        const validatedCurrentHistory: EquipmentInfos[] =
+            equipmentListSchema.validateSync(currentHistory);
+
+        return validatedCurrentHistory;
+    } catch (e) {
+        // if an error is caught when validating or parsing JSON, emptying local storage to prevent blocking bug
+        saveLocalStorageSearchEquipmentHistory(studyUuid, []);
```
sBouzols, line 82: That's a bit brutal to lose all the history here no ?

src/redux/local-storage/search-equipment-history.tsx
```diff
@@ -0,0 +1,95 @@
...
+        return [];
+    }
+};
+
+export const saveLocalStorageSearchEquipmentHistory = (
```
sBouzols, line 87: don't need to export this function.
Actually with all those actions : `*LocalStorageSearchEquipmentHistory` I'm wondering if we could implement something like namespace in javascript with some public or private acces to some function
to do something like 
```
LocalStorageSearchEquipmentHistory.add(studyUuid, equipmentToAdd); // OK
LocalStorageSearchEquipmentHistory.get(studyUuid); // OK
LocalStorageSearchEquipmentHistory.save(studyUuid, equipmentToAdd); // KO private
```

src/redux/local-storage/search-equipment-history.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+const LOCAL_STORAGE_SEARCH_EQUIPMENT_HISTORY_KEY = (
+    APP_NAME + '_SEARCH_EQUIPMENT_HISTORY_'
+).toUpperCase();
+
+const getLocalStorageSearchEquipmentHistoryKey = (studyUuid: UUID) => {
+    return LOCAL_STORAGE_SEARCH_EQUIPMENT_HISTORY_KEY + studyUuid;
+};
+
+const equipmentSchema = yup.object().shape({
```
sBouzols, line 23: Nice for type checking at running !
Maybe we should share with the team about this technique ?

src/redux/reducer.type.ts
```diff
@@ -15,6 +15,19 @@ export enum UpdateTypes {
     STUDY = 'study',
 }
 
+export enum StudyIndexationStatus {
```
sBouzols, line 18: I'm Ok, moving those enum here is better, but is it related to the scope of this PR ? Maybe we should have change this in another PR

src/components/top-bar-equipment-seach-dialog/use-search-matching-equipments.tsx
```diff
@@ -0,0 +1,125 @@
...
+import { searchEquipmentsInfos } from '../../services/study';
+import { UUID } from 'crypto';
+
+interface UseSearchMatchingEquipmentsProps {
+    studyUuid: UUID;
+    nodeUuid: UUID;
+    inUpstreamBuiltParentNode?: boolean;
+    equipmentType?: EquipmentType;
+    makeItems?: (
+        equipmentsInfos: Equipment[],
```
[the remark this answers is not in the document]
sBouzols: I see this is a misnaming in commons-ui as well

src/components/top-bar-equipment-seach-dialog/use-search-matching-equipments.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+export const useSearchMatchingEquipments = (
+    props: UseSearchMatchingEquipmentsProps
+) => {
+    const {
+        studyUuid,
+        nodeUuid,
+        inUpstreamBuiltParentNode,
+        equipmentType,
+        makeItems = getEquipmentsInfosForSearchBar,
```
sBouzols, line 42: well it's never defined to another value than this default `getEquipmentsInfosForSearchBar`, maybe we could simplify removing this prop and calling `getEquipmentsInfosForSearchBar` directly ?

src/components/top-bar-equipment-seach-dialog/use-search-matching-equipments.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+    const updateSearchTerm = useCallback(
+        (newSearchTerm: string) => {
+            setSearchTerm(newSearchTerm);
+            // if user input is empty, return empty array and set isLoading to false without debouncing
+            if (newSearchTerm.length === 0) {
+                setEquipmentsFound([]);
+                setIsLoading(false);
+                return;
+            }
```
sBouzols, line 112: Ok I understand now !

src/redux/local-storage/search-equipment-history.tsx
klesaulnier: I don't think there is any better way, if you can't parse the JSON or if it fails to validate, it means the history is corrupted

src/redux/local-storage/search-equipment-history.tsx
klesaulnier: I don't think there is a way to make private method, but I agree for the "export" removal here

src/redux/local-storage/search-equipment-history.tsx
klesaulnier: I agree, even if this is something we won't often do

src/redux/reducer.type.ts
klesaulnier: I moved it here because it was previously not an enum, it was a javascript Object
I needed it as an enum to use it as a typescript type, and you can't define enums in JS/JSX files

src/components/top-bar-equipment-seach-dialog/use-search-matching-equipments.tsx
klesaulnier: I agree, fixed

src/components/top-bar-equipment-seach-dialog/use-search-matching-equipments.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+            setIsLoading(true);
+            debouncedSearchMatchingEquipments(newSearchTerm);
+        },
+        [debouncedSearchMatchingEquipments]
+    );
+
+    useEffect(() => {
+        setEquipmentsFound([]);
+    }, [equipmentType]);
```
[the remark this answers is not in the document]
klesaulnier: this was the previous behaviour, I'll make some tests with "searchMatchingEquipments"

src/components/top-bar-equipment-seach-dialog/use-search-matching-equipments.tsx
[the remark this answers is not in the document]
klesaulnier: removed now we don't need makeItems method anymore

src/components/top-bar-equipment-seach-dialog/use-search-matching-equipments.tsx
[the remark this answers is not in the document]
klesaulnier: Something changes slightly : 
with 2 tabs open, make a research from the 1st tab
on the 2nd tab, change from "Id" to "Name" for equipment labels
-> on the 1st tab, you will see that the searched elements have been emptied

I think we should keep the behaviour as it was before, what do you think ?

souissimai (approved):

src/redux/reducer.type.ts
sBouzols: damned OK

src/redux/local-storage/search-equipment-history.tsx
sBouzols: Ok then be sure to log an error in the console in this case ?

src/components/app-top-bar.jsx
```diff
@@ -217,16 +94,12 @@ const AppTopBar = ({ user, tabIndex, onChangeTab, userManager }) => {
...
-
     const [themeLocal, handleChangeTheme] = useParameterState(PARAM_THEME);
 
-    const [searchMatchingEquipments, equipmentsFound] =
-        useSearchMatchingEquipments(studyUuid, currentNode?.id);
-
     const showVoltageLevelDiagram = useCallback(
         // TODO code factorization for displaying a VL via a hook
         (optionInfos) => {
+            addToLocalStorageSearchEquipmentHistory(studyUuid, optionInfos);
```
sBouzols, line 102: move this call into TopBarEquipmentSearchDialog::onSelectionChange I think

src/components/top-bar-equipment-seach-dialog/use-search-matching-equipments.tsx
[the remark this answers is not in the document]
sBouzols: Ok keep it like before and we will discuss the expected behavior with PO

sBouzols (approved):
Code review OK
Tests OK
Console warning check OK

PR #2078, opened 2024-05-03
Add short circuit parameter

souissimai:


AAJELLAL (changes requested):
The **validate** button is not disabled after updating the params.

souissimai:
> The **validate** button is not disabled after updating the params.

it was the actual behavior, but fixed it

AAJELLAL (approved):

PR #2079, opened 2024-05-03
Use the localized country hook from commons-ui

flomillot:
Also remove localized-countries and notistack deps


TheMaskedTurtle (approved):

flomillot:
Need new version of commons-ui

PR #2081, opened 2024-05-03
remove operation parameter

anistouri:


dbraquart (approved):
code: ok
tests: ok (I still have a snackbar with error msg in case of non-linear mcs update)

PR #2089, opened 2024-05-15
Do not block the user while deleting stashed modifications

flomillot:
Refactor network modification node editor to handle stashing and restoring modifications in progress, update translations for stashing and restoring modifications.


etiennehomer (commented):
Some postgres deadlocks can be raised now that the user can add a modification while modifications are being deleted. 
We need to investigate while we have the deadlocks and if we can keep this behavior and not having the deadlocks

flomillot:
> Some postgres deadlocks can be raised now that the user can add a modification while modifications are being deleted. We need to investigate while we have the deadlocks and if we can keep this behavior and not having the deadlocks

Will be fixed in next PR

PR #2091, opened 2024-05-15
Powsybl 2024.1.0 - Adaptations - CLA model translation

thangqp:
It is not blocking, we can merge after migration powsybl

See the change in powsybl-dynawo: https://github.com/powsybl/powsybl-dynawo/pull/334/files#diff-a2960bf3c88efc15ceff4ce5a990820d15853a8bc25ca19cae997e14436fb161R383


thangqp:
Expired

thangqp:
Expired

PR #2095, opened 2024-05-21
remove some equipements types from the filters created using the map

Mathieu-Deharbe:
The filter creation on the map now uses directly the list of filterable equipment from commons-ui.


src/components/network/filter-creation-panel.tsx
```diff
@@ -28,11 +29,8 @@ import {
 import { fetchElementsMetadata } from 'services/explore';
 import { UniqueNameInput } from 'components/dialogs/commons/unique-name-input';
 import { useSelector } from 'react-redux';
-import {
-    equipementTypeToLabel,
-    EQUIPMENT_TYPES,
-} from '../utils/equipment-types';
 import { UUID } from 'crypto';
+import { FormEquipment } from '@gridsuite/commons-ui/dist/components/filter/utils/filter-form-utils';
```
jamal-khey, line 33: i think it not normal to include components from @gridsuite/commons-ui/**dist/**

jamal-khey (commented):
imports to be changed

src/components/network/filter-creation-panel.tsx
Mathieu-Deharbe: Indeed. I simply switched it to `from '@gridsuite/commons-ui';`. Webstorm complains but it works fine.

I guess I miss some kind of config for the auto import.

jamal-khey (commented):
Code and test Ok , but you have a error in the console:

```
@formatjs/intl Error MISSING_TRANSLATION] Missing message: "Hvdc" for locale "fr", using id as fallback.

    at new MissingTranslationError2
```

jamal-khey (approved):
Test and code OK
Console logs Ok

jamal-khey (approved):

PR #2096, opened 2024-05-22
Add voltage init alert threshold on shunt compensator activation

FranckLecuyer:


src/components/dialogs/parameters/widget/parameter-float.tsx
```diff
@@ -14,24 +14,35 @@ export interface ParameterFloatProps {
...
 }: ParameterFloatProps) => {
     return (
         <>
-            <Grid item container direction={'row'} spacing={1} paddingTop={3}>
-                <Grid item xs={10} sx={style}>
-                    <FormattedMessage id={label} />
-                </Grid>
-                <Grid item xs={2}>
-                    <FloatInput name={name} adornment={adornment} />
+            <Tooltip title={tooltip ? <FormattedMessage id={tooltip} /> : ''}>
```
jonenst, line 29: don't put Tooltip component if tooltip is not defined ?

src/components/dialogs/parameters/widget/parameter-float.tsx
```diff
@@ -14,24 +14,35 @@ export interface ParameterFloatProps {
...
 export const ParameterFloat = ({
     name,
     label,
     style,
     adornment,
+    tooltip,
 }: ParameterFloatProps) => {
     return (
         <>
-            <Grid item container direction={'row'} spacing={1} paddingTop={3}>
```
jonenst, line 27: previous grid was an item, it means the parent (parent of this component) was a grid container. It should not be ok to replace it with the Tooltip component?

src/components/dialogs/parameters/widget/parameter-float.tsx
FranckLecuyer: Done

src/components/dialogs/parameters/widget/parameter-float.tsx
FranckLecuyer: Done

ayolab (approved):

PR #2098, opened 2024-05-23
add composite modification component

ghazwarhili:


etiennehomer (changes requested):

etiennehomer (changes requested):
https://github.com/gridsuite/gridstudy-app/blob/23a91083007c78fc4e8154c17cf35062f1563c9a/src/translations/en.json#L1190

This translation should be changed to something like : 
2 network modifications successfully composed in one composite modification in /Filtre_expert

etiennehomer (approved):

PR #2099, opened 2024-05-23
Mutualize fetching functions

Meklo:
Need #2187

Depends on gridsuite/commons-ui#455


src/services/utils.js
```diff
@@ -5,139 +5,13 @@
...
-                    text
-            );
-            error.status = response.status;
-        }
-        throw error;
-    });
-};
+import { fetchAppsAndUrls, fetchEnv } from '@gridsuite/commons-ui';
 
 export const getToken = () => {
```
flomillot, line 10: It's still needed ?

flomillot (commented):
Need to update commons-ui or it won't works no ?

src/services/utils.js
flomillot: Well it's the point of the ticket, mutualize what we can, it won't be done after if we don't do it now, i'm pretty sure, so well...
You can check with a TL but for me, it need to be done here alos.

Tristan-WorkGH (changes requested):
Found that
* `src/services/explore.ts`
  - `createFilter()` not removed
  - `elementExists()` not removed
* `commons-ui` version not updated in `package.json`

src/services/utils.js
Meklo: I've just talked about it with @TheMaskedTurtle, we ought to prioritize fixing commons-ui to release gridstudy deadlock.
I can create a following ticket to clean up  the rest of utils functions getVersion, auth function, etc...

Meklo: This said I can clean up `getToken` since it's already in commons ui, didn't noticed earlier

Meklo: Yes I figured this too later, fine by me I'll do a second PR later then

dbraquart (commented):
code: ok
tests: to be done after merge wConclicts

PR #2101, opened 2024-05-28
Fix empty couplingDevices when VL creation by copy

flomillot:
Fixed a bug in voltage-level-creation-dialog.jsx where COUPLING_OMNIBUS was not being properly handled, now defaults to an empty array.

probably introduced by #2072


thangqp (approved):
Code OK, Test OK

PR #2105, opened 2024-05-30
Fix select provider in paramaters

achour94:
To fix the autocomplete render issues in the Edge browser, we have replaced it with a simple MUI Select component provided by commons-ui.


ayolab (changes requested):
The problem is still present on "Generation kind" fields in "non-evacuated energy" parameters, "Sensitivity type" and "Distribution type" in "sensitivity analysis" parameters.
And by the way there is "DropDown" component that is used in "Dynamic simulation" parameters for the same purpose, I think that we should either reuse it or replace it with the MuiSelectInput

achour94:
> The problem is still present on "Generation kind" fields in "non-evacuated energy" parameters, "Sensitivity type" and "Distribution type" in "sensitivity analysis" parameters. And by the way there is "DropDown" component that is used in "Dynamic simulation" parameters for the same purpose, I think that we should either reuse it or replace it with the MuiSelectInput

Thank you for noticing, i resolve the mentionned issues and another one related to 'Solver type' in 'Dynamic simulation'.
Regarding 'DropDown', its a component not connected to react hook form.

ayolab (approved):

anistouri (approved):

PR #2108, opened 2024-05-31
Fix map reloading

sBouzols:
Disable condition to avoid map reloading if there is no impact on map elements at modification creation but this condition breaks update for other uses cases (node change, node build, etc...)


Tristan-WorkGH (approved):
okay as it's not impacting functional, but need to fix it in another ticket for listing all cases & conditions.

achour94 (approved):

PR #2110, opened 2024-05-31
"Fix bug causing tooltip to be disabled"

jamal-khey:
This pull request fixes a bug that was causing the tooltip to be disabled in the MapViewer component. The bug was caused by the incorrect handling of the drawing mode state. This PR removes the unnecessary state and ensures that the tooltip is enabled correctly.


src/redux/actions.js
```diff
@@ -509,6 +509,12 @@ export function setModificationsInProgress(isModificationsInProgress) {
     };
 }
 
+export const STUDY_DISPLAY_MODE = {
```
jonenst, line 512: is this needed ?

jamal-khey: has been included by mistake during the merge with main

jonenst (approved):

sBouzols (approved):
Test OK
Code Review OK
Console Warning check OK

PR #2111, opened 2024-06-03
Add filter name validation

jamal-khey:
This pull request adds validation for the filter name input field. Previously, the validation was not triggered when the name was unchanged. With this change, the validation will be triggered even if the name is unchanged. This ensures that the input field is properly validated.

### DIFF
--- a/src/components/dialogs/commons/unique-name-input.tsx
+++ b/src/components/dialogs/commons/unique-name-input.tsx
@@ -28,6 +28,7 @@ interface UniqueNameInputProps {
     autoFocus?: boolean;
     onManualChangeCallback?: () => void;
     activeDirectory: UUID | null;
+    triggerValidation?: boolean;
     formProps?: Omit<
         TextFieldProps,
         | 'value'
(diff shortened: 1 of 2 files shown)


src/components/map-viewer.jsx
```diff
@@ -202,8 +202,11 @@ const MapViewer = ({
                     }),
                     headerId: 'FilterCreationError',
                 });
+                setIsLoading(false);
+                return false;
             }
             setIsLoading(false);
+            return true; // success
```
sBouzols, line 209:
```suggestion
                return false;
            } finally {
               setIsLoading(false);
            }
            return true; // success
```

jamal-khey: done

sBouzols (commented):
Please Check console warning about translation error
Change behaviour to show name input normally at filter save and not in red as error which could be confusing for the user (does my saving worked ?)
I created a Refactor PR #2112 , we could discuss about it later.

src/components/network/selection-creation-panel.tsx
```diff
@@ -274,21 +281,13 @@ const SelectionCreationPanel: React.FC<SelectionCreationPanelProps> = ({
...
-                                if (isValid && defaultFolder) {
-                                    onSaveSelection(
-                                        formMethods.getValues() as ISelectionCreation,
-                                        defaultFolder,
-                                        setSavingState
-                                    );
-                                }
-                            });
-                        }}
+                        disabled={
```
souissimai, line 284: we do not use isValid, it's isDirty

jamal-khey: i have some strange behavior with isDirty.
i suggest to keep it as it is , because the behavior of this button was like that, i just moved the code

souissimai (approved):
test ok

sBouzols (approved):
Code review OK
Tests OK
console warning check OK

PR #2112, opened 2024-06-04
Refactor saveMapFilter

sBouzols:
Add custom hook to manage saveMapFilter and its pending state.
Add a prop callback to th FilterCreationPanel to get equipments from its parent MapViewer which hold the networkMapref at submit 
some renamings


sBouzols:
done by #2111

PR #2114, opened 2024-06-05
[Dynamic Simulation] Can not trip the line from DSL

thangqp:
Due to the modif : https://github.com/gridsuite/commons-ui/pull/432/files#diff-b891b5128b3aa4efdf36efac0cbadfadfa2cd2a755849485ebe3959a49914965L97
=> can not trip the line from DSL

Solution: set default value as null which is an accepted value for select input


etiennehomer (approved):
Test ok

PR #2121, opened 2024-06-13
updates sensi local parameters

Mathieu-Deharbe:


souissimai (approved):
test ok but maybe need to refactor in other Ts

Mathieu-Deharbe:
Agreed. I did a small correction but that part of the code needs a big refacto.

PR #2135, opened 2024-06-24
For sort/filter/shading pb for normalized columns using flux convention settings

FranckLecuyer:


dbraquart (approved):
code: ok
tests: ok

PR #2136, opened 2024-06-25
Add short-circuit parameters load and save

dbraquart:


FranckLecuyer (changes requested):
Code review : OK
Tests: 
When loading a settings element which have 'Predefined parameters' set on 'ISC max with standard IEC 909', the 'Init voltage profile' radio button remains unselected (it should be set on 'IEC 909')

dbraquart:
> Tests:
> When loading a settings element which have 'Predefined parameters' set on 'ISC max with standard IEC 909', the 'Init voltage profile' radio button remains unselected (it should be set on 'IEC 909')

Yes it is a special case: CONFIGURED is part of a 3-state enum, but displayed with a 2-state radio button in the form.

FranckLecuyer (approved):
Tests: OK

PR #2141, opened 2024-06-26
Fix null param in getQueryParamsList

jamal-khey:
This pull request fixes a null parameter issue in the `getQueryParamsList` function. The issue was causing errors when the function was called with a null parameter. This PR ensures that the function handles null parameters correctly, preventing any potential errors.

### DIFF
--- a/src/services/study/network-map.js
+++ b/src/services/study/network-map.js
@@ -86,13 +86,16 @@ export function fetchVoltageLevelEquipments(
 export function fetchEquipmentsIds(
     studyUuid,
     currentNodeUuid,
-    substationsIds,
+    substationsIds = undefined,
     equipmentType,
     inUpstreamBuiltParentNode,
     nominalVoltages = undefined
 ) {
+    const substationsCount = substationsIds ? substationsIds.length : 0;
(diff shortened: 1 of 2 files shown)


src/services/study/network-map.js
```diff
@@ -92,7 +92,7 @@ export function fetchEquipmentsIds(
     nominalVoltages = null
```
sBouzols, line 92:
```suggestion
    nominalVoltages = undefined
```
Do not modify `getQueryParamsList` which is used in a lot of places

src/services/utils.js
```diff
@@ -196,7 +196,7 @@ export const fetchDefaultParametersValues = () => {
     });
 };
 export const getQueryParamsList = (params, paramName) => {
-    if (params !== undefined && params.length > 0) {
+    if (params != null && Array.isArray(params) && params.length > 0) {
```
souissimai, line 199: (params != null && Array.isArray(params)) || params !== undefined

jamal-khey: `params !== null ` is equivalent to `params != null  && params != undefined`

sBouzols (approved):

PR #2142, opened 2024-06-27
Fix display of alert on reactive slacks threshold field in voltage init parameters.

FranckLecuyer:


dbraquart (approved):
code: ok
tests: ok

PR #2147, opened 2024-07-01
add status FAILED for shortcircuit computation

EtienneLt:
Create a new status for failures of shortcircuit computation


antoinebhs (approved):
Code OK

PR #2149, opened 2024-07-03
fix serachbar filter style

YenguiSeddik:


src/translations/fr.json
```diff
@@ -1372,6 +1372,6 @@
     "startStateEstimationError": "Une erreur est survenue lors du lancement de l'estimation d'état",
     "stateEstimationError": "Une erreur est survenue lors de l'exécution de l'estimation d'état",
     "NoFilter": "Aucun filtre",
-    "EquipmentSearchPlaceholder": "Rechercher un poste, une ligne, un ouvrage...",
+    "EquipmentSearchPlaceholder": "Recherche un poste, une ligne, un ouvrage...",
```
EstherDarkish, line 1375: Should be 
`"EquipmentSearchPlaceholder": "Recherche",`
instead.
And the en.json file should also be fixed.

EstherDarkish (changes requested):

src/translations/fr.json
YenguiSeddik: done

EstherDarkish (approved):

PR #2150, opened 2024-07-03
fix hvdc meaning, to line hvdc with the conversion stations

EtienneLt:
change hvdc meaning, it means the line with the conversion stations


ghazwarhili (changes requested):
revert modifications in package-lock.json file

src/components/dialogs/network-modifications/equipment-deletion/equipment-deletion-form.jsx
```diff
@@ -53,7 +53,11 @@ const DeleteEquipmentForm = ({ studyUuid, currentNode, editData }) => {
     const { setValue } = useFormContext();
 
     const richTypeLabel = (rt) => {
-        return intl.formatMessage({ id: rt });
+        if (rt === EQUIPMENT_TYPES.HVDC_LINE) {
+            return intl.formatMessage({ id: 'Hvdc' });
+        } else {
+            return intl.formatMessage({ id: rt });
+        }
```
ghazwarhili, line 60: update the variable name 'rt' to make it more descriptive
for example : 

```const richTypeLabel = (richType) => {
    if (richType === EQUIPMENT_TYPES.HVDC_LINE) {
        return intl.formatMessage({ id: 'Hvdc' });
    }
    return intl.formatMessage({ id: richType });
};
```

ghazwarhili (approved):

PR #2152, opened 2024-07-04
hide tab N if not developper mode for AS results

souissimai:


AAJELLAL (changes requested):
The results are poorly displayed: the content of tab N is displayed in the n-K tab and the content of N-K in the logs tab.

AAJELLAL (approved):
Code and tests are OK.

**I noticed a strange behaviour(this behavior is reproducible on devopf too) but not linked to this development:**

- open the application in a first window then go to the security analysis results tab.
- open  the application in a second window and go to the settings.
- deactivate develop mode using the second window.
- realize that the result tab (first window) is automatically placed on the loadflow part.

jamal-khey (changes requested):
found a bug related to `enableDeveloperMode ? LOGS_TAB_INDEX : LOGS_TAB_INDEX - 1` , when logs are selected and we disable developper mode, the tabs contents are not displayed any more

src/components/results/securityanalysis/security-analysis-result-tab.tsx
```diff
@@ -252,6 +281,10 @@ export const SecurityAnalysisResultTab: FunctionComponent<
     const { loading: filterEnumsLoading, result: filterEnums } =
         useFetchFiltersEnums();
 
+    useEffect(() => {
+        console.log('$$$$$$$$$$$ ', tabIndex);
+    }, [tabIndex]);
+
```
jamal-khey, line 287: to delete

jamal-khey (commented):
i am not a fan of using useEffect, but i don't have a better solution.
one console.log to remove

AAJELLAL (commented):
Seen together: the change between tabs when you activate developer mode is done strangely.

jamal-khey (approved):
Test OK -> some strange behavior with tabs , but not related to this PR
console log check Ok

PR #2157, opened 2024-07-08
Enable cgmes export

etiennehomer:
Since https://github.com/powsybl/powsybl-network-store-server/commit/257f14a07f02c59aabfc2323a7a7495a2594b223 was merged, we don't have performance issues anymore and we can enable CGMES export.


antoinebhs (approved):
Code OK

PR #2160, opened 2024-07-09
fix reset for modification dialog

EtienneLt:
fix a bug on modification dialog, 
when modifying an element on an not build node, if coming back to the editor the value disappeared.


sBouzols (changes requested):
missing changes on src/components/dialogs/network-modifications/vsc/modification/vsc-modification-dialog.tsx I think
will test

PR #2167, opened 2024-07-12
Draft:[POC websocket] setting usage example of websocket on gridstudy

capyq:


src/components/app.jsx
```diff
@@ -541,6 +543,7 @@ const App = () => {
         setTabIndex(newTabIndex);
     }, []);
 
+    const urlMapper = useWebsocketUrlGenerator();
```
jonenst, line 546: I think we should keep the different sockets separated for now. Any code can call addListener/removeListener and refcount to open/close the websocket accordingly

src/components/study-container.jsx
```diff
@@ -438,49 +438,28 @@ export function StudyContainer({ view, onChangeTab }) {
...
-                        studyParentDirectoriesUuidsRef.current.includes(
-                            eventData.headers['directoryUuid']
-                        )
-                    ) {
-                        fetchStudyPath();
-                    }
-                }
+    useListener(WS_URL_KEYS.DIRECTORIES, (event) => {
+        const eventData = JSON.parse(event.data);
+        dispatch(studyUpdated(eventData));
```
jonenst, line 443: I think the objective was to get rid of studyUpdateForce in the reducer ? maybe we got lost in our deliberations, should we refocus on this ?

capyq: The main goal of this draft PR is to discuss the usage of the websocket component.

The rework from the current usage to this should be done in another PR after merging.

src/components/study-container.jsx
```diff
@@ -438,49 +438,28 @@ export function StudyContainer({ view, onChangeTab }) {
...
+        if (
+            eventData.headers &&
+            eventData.headers['notificationType'] ===
+                directoriesNotificationType.UPDATE_DIRECTORY
+        ) {
```
sBouzols, line 448: GridStudy and GridExplore often filter notifications like that on a header value.
Maybe we should add this ability directly when registering a listener ?
something like
```
useListener(WS_URL_KEYS.DIRECTORIES, eventComparator, (event) => {
```
with eventComparator : (event) => boolean ?

src/services/utils.js
```diff
@@ -14,7 +14,11 @@ export const FetchStatus = {
     RUNNING: 'RUNNING',
 };
 
-export const getWsBase = () => document.baseURI.replace(/^http:\/\//, 'ws://').replace(/^https:\/\//, 'wss://');
+export const getWsBase = () =>
+    document.baseURI
+        .replace(/^http:\/\//, 'ws://')
+        .replace(/^https:\/\//, 'wss://')
+        .replace(/:3004/, ':84');
```
sBouzols, line 21: why this ?

PR #2174, opened 2024-07-17
Keep sort when switching between tabs

YenguiSeddik:


souissimai:
tested, it seems good for me

souissimai (approved):
tested, it seems good for me

src/hooks/use-aggrid-sort.ts
```diff
@@ -26,29 +28,33 @@ export enum SortWay {
...
                     (sort) =>
                         (sort.children ?? false) !==
                         (newSortConfig.children ?? false)
                 )
-                .concat(newSortConfig)
-        );
-    }, []);
+                .concat(newSortConfig);
+
+            setSortConfig(updatedSortConfig);
```
klesaulnier, line 49: the previous way to update state was correct, that's how we should implement it when the state depends on the previous state value

plus, I'm a bit confused, why do we need "sortConfig" useState while everything is now stored in redux

src/redux/actions.js
```diff
@@ -688,6 +688,16 @@ export function setLoadflowResultFilter(filterTab, loadflowResultFilter) {
...
+export const LOADFLOW_RESULT_SORT = 'LOADFLOW_RESULT_SORT';
+
+export function setLoadflowResultSort(tab, loadflowResultSort) {
+    return {
+        type: LOADFLOW_RESULT_SORT,
+        tab: tab,
+        loadflowResultSort: loadflowResultSort,
+    };
+}
+
```
klesaulnier, line 700: I know it's done the same with filters, but it would be a bit better to have only one action instead of an action for each result type

src/components/spreadsheet/table-wrapper.jsx
```diff
@@ -388,6 +385,7 @@ const TableWrapper = (props) => {
                 };
             }
 
+            console.log('test useCallback sortConfig', sortConfig);
```
klesaulnier, line 388: to remove

klesaulnier (changes requested):
I think there is a bug with initial sort
On first loading of each aggrid table depending on backend, this sort value is sent to backend : "sort=subjectId,**undefined**"
it breaks the API call

For everything else, it seems to work (except for sensitivity analysis, as stated in another comment)

src/hooks/use-aggrid-sort.ts
YenguiSeddik: i removed the useState

src/hooks/use-aggrid-sort.ts
```diff
@@ -26,29 +28,33 @@ export enum SortWay {
 }
 
 export const useAgGridSort = (
-    initSortConfig: SortConfigType
+    initSortConfig: SortConfigType[],
+    sortAction?: (tab: string, sortConfig: SortConfigType[]) => AnyAction,
+    tab?: string
```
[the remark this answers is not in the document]
YenguiSeddik: i changed the solution after using one action to set the sort for all the tables

src/components/spreadsheet/table-wrapper.jsx
```diff
@@ -430,8 +431,12 @@ const TableWrapper = (props) => {
     }, [errorMessage, snackError]);
 
     useEffect(() => {
-        initSort(defaultSortColKey);
-    }, [tabIndex, defaultSortColKey, initSort]);
+        /*gridRef.current?.columnApi?.applyColumnState({
```
souissimai, line 434: maybe to remove comment?

src/redux/reducer.type.ts
```diff
@@ -29,6 +38,31 @@ export enum StudyDisplayMode {
     DRAW = 'Draw',
 }
 
+interface AnalysisResultSort {
+    [key: string]: SortConfigType[];
+}
+
+interface SpreadsheetSort {
+    [key: string]: SortConfigType[];
+}
```
klesaulnier, line 47: do we need to separate them ?

klesaulnier (approved):
Test: OK
Small type changes

souissimai (approved):

PR #2178, opened 2024-07-18
rework useComputingStatus hooks

capyq:
remove duplication and add generic mapper. reduce complexity of this hook


Tristan-WorkGH (approved):

src/components/computing-status/use-computing-status.ts
```diff
@@ -76,43 +64,88 @@ function isWorthUpdate(
...
-    invalidations,
-    completions,
     resultConversion,
     computingType,
     optionalServiceAvailabilityStatus = OptionalServicesStatus.Up
 ) => {
+    const nodeUuid = useSelector(
+        (state: ReduxState) => state.currentTreeNode?.id
+    );
+    const studyUuid = decodeURIComponent(useParams().studyUuid ?? '') as UUID;
```
klesaulnier, line 128: makes me wonder why we are storing study uuid in redux store

src/components/computing-status/use-computing-status.ts
```diff
@@ -76,43 +64,88 @@ function isWorthUpdate(
...
+        (runningStatus: RunningStatus) =>
+            [RunningStatus.FAILED, RunningStatus.SUCCEED].includes(
+                runningStatus
+            ) &&
+            status.completions.includes(
                 lastUpdateRef.current?.studyUpdatedForce.eventData?.headers
                     ?.updateType ?? ''
             ),
-        [completions]
+        [status.completions]
```
klesaulnier, line 148: since status isn't memoized, I think this callback will be recreated every time anyway
memoize status or remove this useCallback ?

capyq: status is a constant because it is defined outside the function and thus becomes a pointer to the fixed memory object

src/components/computing-status/use-computing-status.ts
sBouzols: I would prefer using the one from the store instead of the one from the URL.
It's a matter of responsibility for me, why a custom hook for computation status should be able to parse something in the URL... ? seems strange
Otherwise it's not strange if this code access to global states no ?

capyq: @jonenst what is your point of view on this ?

jonenst: useParams is obviously less flexible than taking the value from something else from the react context (e.g. a store, or our own contextprovider) because it means you force the code that can be deeply nested to know that there is a <Route match=..> high above and that the url has some kind of known structure.

I don't think it matters much in this case though. Maybe juste add a wrapper useStudyUuid (implemented with useParams) so we can track more easily which component access this value, and change it if we ever need to display this components with a different study than the one in the url or in a context without a router

capyq:
@klesaulnier, rebase done

klesaulnier (approved):
Test: OK
Code: OK except for use-study-uuid hook that needs to be validated

src/components/computing-status/use-computing-status.ts
```diff
@@ -71,38 +57,63 @@ function isWorthUpdate(
...
     optionalServiceAvailabilityStatus = OptionalServicesStatus.Up
 ) => {
+    const nodeUuid = useSelector((state: AppState) => state.currentTreeNode?.id);
+    const studyUuid = useSelector((state: AppState) => state.studyUuid);
     const nodeUuidRef = useRef<UUID | null>(null);
     const studyUpdatedForce = useSelector((state: AppState) => state.studyUpdated);
     const lastUpdateRef = useRef<LastUpdateProps | null>(null);
-    const dispatch = useDispatch<AppDispatch>();
-
+    const dispatch = useDispatch();
```
Tristan-WorkGH, line 108: rollback
```suggestion
    const dispatch = useDispatch<AppDispatch>();
```

src/components/computing-status/use-computing-status.ts
```diff
@@ -113,23 +124,24 @@ export const useComputingStatus: UseComputingStatusProps = (
...
+                })
+                .catch(() => {
+                    if (!canceledRequest) {
+                        dispatch(setComputingStatus(computingType, RunningStatus.FAILED));
+                    }
+                });
+        }
         return () => {
             canceledRequest = true;
         };
```
Tristan-WorkGH, line 147: isn't the request and cancel must be in same condition here?

Tristan-WorkGH (changes requested):

src/components/computing-status/use-computing-status.ts
Tristan-WorkGH: ping @TheMaskedTurtle

PR #2180, opened 2024-07-19
update printWidth to 120

capyq:
set print width to 120 instead of default 80.

then running npx eslint --fix  . to auto fix all issue

Following gridsuite/commons-ui#533


jonenst (approved):

Tristan-WorkGH (approved):
:tada:

PR #2196, opened 2024-07-30
Study Export with choice of format: editing the file name

Mathieu-Deharbe:
allows the user to pick a filename when exporting a node from a study.

prefills the text field with [studyName]_[nodeName]


src/components/dialogs/export-dialog.jsx
```diff
@@ -79,6 +82,23 @@ const ExportDialog = ({
             });
     }, [studyUuid, snackError]);
 
+    const fetchNodeName = useCallback(
```
ghazwarhili, line 85: You don't need to do a fetch for node name,  it is stored in redux "CurrentTreeNodeData"
you can get it from the redux like CurrentTreeNode.data.label

ghazwarhili (changes requested):
1) click on Export case button 
2) select name and format then click on Export button
3) click on results tab ( or another tab)
Capture d’écran du 2024-08-02 15-02-50

src/components/dialogs/export-dialog.jsx
thangqp: nodeUuid is passed by contextual menu, not the current node selected, so we can not use currentTreeNode in redux
In contrast, we can use another item in redux, i.e.   

```js
    const treeModel = useSelector(
        (state) => state.networkModificationTreeModel
    );
    const	nodeName = treeModel.find((node) => node.id === nodeUuid))?.data.label
```

src/components/dialogs/export-dialog.jsx
```diff
@@ -177,9 +200,20 @@ const ExportDialog = ({
...
-                <div style={{ marginTop: '0.8em' }} />
+            <DialogTitle>{title}</DialogTitle>
+            <DialogContent>
+                <TextField
+                    key="fileName"
+                    margin="dense"
+                    label={<FormattedMessage id="download.fileName" />}
+                    id="fileName"
+                    value={fileName}
+                    style={{ width: '100%' }}
```
thangqp, line 211: use sx instead of style, add a margin bottom 8px to insert a space to the below component

Mathieu-Deharbe: OK

src/components/dialogs/export-dialog.jsx
Mathieu-Deharbe: Thanks.

Mathieu-Deharbe:
> 1. click on Export case button
> 2. select name and format then click on Export button
> 3. click on results tab ( or another tab)
>    Capture d’écran du 2024-08-02 15-02-50

I can't find any any export case button. Is it in gridexplore ?

Also I can’t enlarge your image. I can't see what is written.

thangqp (approved):
Code OK, test OK

ghazwarhili (approved):

PR #2203, opened 2024-08-06
Open nad from map selection

klesaulnier:
When selecting multiple VLs from the map and attempting to open NAD, an error occurs stating that the diagram cannot be loaded. We should consider consulting with the PO to determine whether we should limit the number of VLs a user can select. Alternatively, we could display a clearer error message to the user.


src/components/diagrams/diagram-header.jsx
```diff
@@ -22,6 +22,10 @@ const BLINK_LENGTH_MS = 1800;
 
 const styles = {
     header: (theme) => ({
+        // prevent header from making the window wider, prevent bugs when displaying a lot of different voltage levels
+        position: 'absolute',
+        width: '100%',
+        ////
```
souissimai, line 28: to remove comment

src/components/diagrams/diagram-pane.jsx
```diff
@@ -234,7 +234,7 @@ const useDisplayView = (studyUuid, currentNode) => {
             } else if (diagramState.svgType === DiagramType.SUBSTATION) {
                 return createSubstationDiagramView(diagramState.id, diagramState.state);
             } else if (diagramState.svgType === DiagramType.NETWORK_AREA_DIAGRAM) {
-                return createNetworkAreaDiagramView(diagramState.ids, diagramState.state, diagramState.depth);
+                return createNetworkAreaDiagramView(diagramState.ids.sort(), diagramState.state, diagramState.depth);
```
souissimai, line 237: If we want to open NAD for multiple VLs using a selection map, sorting the IDs can be beneficial. However, in cases where NAD is opened one by one, it may be preferable to maintain the order in which the user requested it.

souissimai:
test ok, code ok with minor remarks

souissimai (approved):

EstherDarkish (approved):
Code OK Tests OK

PR #2206, opened 2024-08-07
Add parameters to initialize Nad with Geo Data

souissimai:


src/components/dialogs/parameters/network-area-diagram-parameters.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export const NetworkAreaDiagramParameters = () => {
+    return (
+        <>
+            <Grid
+                xl={6}
+                container
+                spacing={1}
+                sx={styles.scrollableGrid}
+                key={'networkAreaDiagramParameters'}
```
klesaulnier, line 20: no need to set a key since it's not within an list

klesaulnier (approved):
Test: OK
Code: OK

Rename "initNadGeoData" to "withGeoData" / "withCoordinates"

PR #2208, opened 2024-08-09
Tests/haromize tests

capyq:
Implement same test librairy than common-ui


capyq:
@jonenst, rebase done.

src/components/computing-status/use-computing-status.ts
```diff
@@ -71,38 +58,64 @@ function isWorthUpdate(
     return false;
 }
 
+const genetateStatusFromComputingType = (base: string) => ({
```
jonenst, line 61: Typo generate

jonenst (approved):

PR #2212, opened 2024-08-16
Add missing code for debugging Redux from DevTools

Tristan-WorkGH:


src/redux/store.ts
```diff
@@ -5,12 +5,52 @@
...
+function sanitizeUser(data: User | null) {
+    return (
+        data &&
+        ({
+            ...data,
+            id_token: data?.id_token && '*****',
+            access_token: data?.access_token && '*****',
+            refresh_token: data?.refresh_token && '*****',
+            //state: data.state && '<?>',
+        } as User | null)
```
capyq, line 25: RMQ => do not cast as null.

src/redux/store.ts
```diff
@@ -5,12 +5,52 @@
...
+    );
+}
+function stateSanitizer<S>(state: S, index: number): S {
+    // @ts-expect-error haven't found why generic can't be set to `AppState`...
+    return { ...state, user: sanitizeUser(state.user) };
+}
```
capyq, line 31: Check with this
```suggestion
function stateSanitizer<TState extends { user: User | null }>(state: TState, index: number): TState {
    return { ...state, user: sanitizeUser(state.user) };
}
```

capyq (changes requested):

PR #2213, opened 2024-08-19
Add icons to the guidance popup text.

AAJELLAL:


src/components/map-viewer.jsx
```diff
@@ -5,7 +5,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import { useState, useCallback, useRef, useEffect } from 'react';
+import React, { useState, useCallback, useRef, useEffect } from 'react';
```
jamal-khey, line 8: you don't need to include React

jamal-khey (approved):

PR #2217, opened 2024-08-22
implement RHF in security analysis parameters form

jamal-khey:
replace manual validation by yup schema
use submit button to submit all the params at once


src/components/dialogs/parameters/common/limitreductions/columns-definitions.ts
```diff
@@ -84,6 +91,23 @@ export const getLimitReductionsFormSchema = (nbTemporaryLimits: number) => {
...
+                .min(0, 'NormalizedPercentage')
+                .max(100, 'NormalizedPercentage')
+                .required(),
+            [PARAM_SA_LOW_VOLTAGE_ABSOLUTE_THRESHOLD]: yup.number().required(),
+            [PARAM_SA_HIGH_VOLTAGE_PROPORTIONAL_THRESHOLD]: yup
+                .number()
+                .min(0, 'NormalizedPercentage')
+                .max(100, 'NormalizedPercentage')
+                .required(),
+            [PARAM_SA_HIGH_VOLTAGE_ABSOLUTE_THRESHOLD]: yup.number().required(),
```
antoinebhs, line 110: This is shared between LF parameters and SA parameters, I don't think we want checks specific to SA here.

src/components/dialogs/parameters/parameters.jsx
```diff
@@ -130,7 +130,7 @@ export const styles = {
     singleItem: (theme) => ({
         display: 'flex',
         flex: 'auto',
-        alignItems: 'center',
+        alignItems: 'flex-start',
```
antoinebhs, line 133: why ?

src/components/dialogs/parameters/security-analysis/security-analysis-parameters-selector.tsx
```diff
@@ -54,14 +54,21 @@ const SecurityAnalysisParametersSelector: FunctionComponent<{
...
+                            sx={{
+                                width: '85%',
+                                display:
+                                    tabValue === TAB_VALUES.LimitReductions && params.limitReductions
+                                        ? 'block'
+                                        : 'none',
+                            }}
+                        >
+                            <LimitReductionsTableForm limits={params.limitReductions} />
+                        </Box>
```
antoinebhs, line 71: why ? what's the link with RHF ?

antoinebhs (changes requested):
LF limit reductions does not display with your changes

src/components/dialogs/parameters/common/limitreductions/limit-reductions-table-form.tsx
```diff
@@ -54,30 +48,6 @@ const LimitReductionsTableForm: FunctionComponent<{
...
-        limits.forEach((limit, index) => {
-            formValues[LIMIT_DURATION_FORM + index] = limit.reduction;
-        });
-        return formValues;
-    };
-
-    useEffect(() => {
-        reset(toFormValues());
-    }, [reset, toFormValues]);
-
```
antoinebhs, line 80: added back only in SA ?

src/components/dialogs/parameters/security-analysis/security-analysis-parameters.tsx
```diff
@@ -123,12 +139,19 @@ export const SecurityAnalysisParameters: FunctionComponent<{
...
         (formLimits: Record<string, any>) => {
             updateParameters({
                 ...params,
+                [PARAM_SA_FLOW_PROPORTIONAL_THRESHOLD]: formLimits[PARAM_SA_FLOW_PROPORTIONAL_THRESHOLD] / 100,
+                [PARAM_SA_LOW_VOLTAGE_PROPORTIONAL_THRESHOLD]:
+                    formLimits[PARAM_SA_LOW_VOLTAGE_PROPORTIONAL_THRESHOLD] / 100,
+                [PARAM_SA_LOW_VOLTAGE_ABSOLUTE_THRESHOLD]: formLimits[PARAM_SA_LOW_VOLTAGE_ABSOLUTE_THRESHOLD],
+                [PARAM_SA_HIGH_VOLTAGE_PROPORTIONAL_THRESHOLD]:
+                    formLimits[PARAM_SA_HIGH_VOLTAGE_PROPORTIONAL_THRESHOLD] / 100,
+                [PARAM_SA_HIGH_VOLTAGE_ABSOLUTE_THRESHOLD]: formLimits[PARAM_SA_HIGH_VOLTAGE_ABSOLUTE_THRESHOLD],
```
antoinebhs, line 154: strange to update general parameters in `updateLimitReductions` function ?

src/components/dialogs/parameters/parameters.jsx
jamal-khey: center the elements of the container may cause form validation error messages to appear misaligned.
the form is looking better with  'flex-start':

image

image

src/components/dialogs/parameters/security-analysis/security-analysis-parameters-selector.tsx
jamal-khey: we need to keep the inputs of the form always instantiated 
- state can be lost if we mount / unmount the component ( previously state was managed with a 'useState'
- initialisation of the form crashed the app if i remember correctly.

src/components/dialogs/parameters/security-analysis/security-analysis-parameters.tsx
jamal-khey: i renamed the function to 'updateSaParametters'

src/components/dialogs/parameters/common/limitreductions/limit-reductions-table-form.tsx
jamal-khey: you are right.
i fixed it by adding a new custuom hook to reuse this logic . useLimitReductionsForm.ts

src/components/dialogs/parameters/common/limitreductions/columns-definitions.ts
jamal-khey: fixedi removed this change ad added specific schema rules to AS

jamal-khey (commented):
fix review comments

antoinebhs (changes requested):
* When the provider is not OLF, the application crashes with `limits is null`.
* Can we improve the fact that when we switch to the security-analysis tab in the parameters we see the value change from "0" to the actual DB value?

src/components/dialogs/parameters/common/limitreductions/limit-reductions-table-form.tsx
antoinebhs: If it's to use it in LF and SA, I'm not sure why we need to move it and not keep it here?

src/components/dialogs/parameters/security-analysis/security-analysis-parameters.tsx
```diff
@@ -123,12 +164,19 @@ export const SecurityAnalysisParameters: FunctionComponent<{
         [params]
     );
 
-    const { handleSubmit, formState } = formMethods;
+    const { handleSubmit, formState, reset } = formMethods;
 
-    const updateLimitReductions = useCallback(
+    const updateSaParametters = useCallback(
```
antoinebhs, line 169:
```suggestion
    const updateParameters = useCallback(
```
could be fine

src/components/dialogs/parameters/security-analysis/security-analysis-parameters.tsx
```diff
@@ -139,6 +187,53 @@ export const SecurityAnalysisParameters: FunctionComponent<{
         setHaveDirtyFields(!!Object.keys(formState.dirtyFields).length);
     }, [formState, setHaveDirtyFields]);
 
+    const { toFormValuesLimitReductions } = useLimitReductionsForm(params?.limitReductions);
+
+    useEffect(() => {
+        const limitReductionFormValues = toFormValuesLimitReductions();
+        const asSpecificFormValues = {
```
antoinebhs, line 194:
```suggestion
        const SASpecificFormValues = {
```
as is french?

src/components/dialogs/parameters/common/limitreductions/useLimitReductionsForm.ts
```diff
@@ -0,0 +1,40 @@
...
+            [LIMIT_REDUCTIONS_FORM]: limits.map((vlLimits) => ({
+                [VOLTAGE_LEVELS_FORM]: vlLimits.voltageLevel.nominalV + ' (kV)',
+                [IST_FORM]: vlLimits.permanentLimitReduction,
+                ...toFormValuesFromTemporaryLimits(vlLimits.temporaryLimitReductions),
+            })),
+        };
+    }, [limits]);
+
+    return { toFormValuesLimitReductions };
+};
```
klesaulnier, line 40: since you don't use anything from a state, from the store, I don't think we need a hook here

you could create "toFormValuesLimitReductions" as a utility method instead, and create a useMemo in components that would call this method and store its result depending on limits

src/components/dialogs/parameters/security-analysis/security-analysis-parameters-selector.tsx
klesaulnier: I think I prefer the old solution as well, everything should be stored in RHF here, so it should not cause any problem

src/components/dialogs/parameters/common/limitreductions/columns-definitions.ts
```diff
@@ -4,6 +4,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
+
```
klesaulnier, line 7: remove change to prevent unnecessary file change

src/components/dialogs/parameters/security-analysis/security-analysis-parameters.tsx
```diff
@@ -92,15 +102,46 @@ export const SecurityAnalysisParameters: FunctionComponent<{
...
+                .min(0, 'NormalizedPercentage')
+                .max(100, 'NormalizedPercentage')
+                .required(),
+            [PARAM_SA_HIGH_VOLTAGE_ABSOLUTE_THRESHOLD]: yup.number().required(),
+        });
+
+        return yup.object().shape({
+            ...limitReductionsSchema.fields,
+            ...thresholdsSchema.fields,
+        });
```
klesaulnier, line 133: move all of this to utils file

src/components/dialogs/parameters/security-analysis/security-analysis-parameters.tsx
```diff
@@ -92,15 +102,46 @@ export const SecurityAnalysisParameters: FunctionComponent<{
...
 
     const formMethods = useForm({
-        defaultValues: { [LIMIT_REDUCTIONS_FORM]: [] },
+        defaultValues: {
+            [LIMIT_REDUCTIONS_FORM]: [],
+            [PARAM_SA_FLOW_PROPORTIONAL_THRESHOLD]: '',
+            [PARAM_SA_LOW_VOLTAGE_PROPORTIONAL_THRESHOLD]: '',
+            [PARAM_SA_LOW_VOLTAGE_ABSOLUTE_THRESHOLD]: '',
+            [PARAM_SA_HIGH_VOLTAGE_PROPORTIONAL_THRESHOLD]: '',
+            [PARAM_SA_HIGH_VOLTAGE_ABSOLUTE_THRESHOLD]: '',
```
klesaulnier, line 143:
```suggestion
            [PARAM_SA_FLOW_PROPORTIONAL_THRESHOLD]: 0,
            [PARAM_SA_LOW_VOLTAGE_PROPORTIONAL_THRESHOLD]: 0,
            [PARAM_SA_LOW_VOLTAGE_ABSOLUTE_THRESHOLD]: 0,
            [PARAM_SA_HIGH_VOLTAGE_PROPORTIONAL_THRESHOLD]: 0,
            [PARAM_SA_HIGH_VOLTAGE_ABSOLUTE_THRESHOLD]: 0,
```

all of these data are numbers, not string
or set them to "null" if you really want to make them empty by default

src/components/dialogs/parameters/security-analysis/security-analysis-parameters.tsx
```diff
@@ -139,6 +187,53 @@ export const SecurityAnalysisParameters: FunctionComponent<{
...
+
+    useEffect(() => {
+        const limitReductionFormValues = toFormValuesLimitReductions();
+        const SASpecificFormValues = {
+            [PARAM_SA_FLOW_PROPORTIONAL_THRESHOLD]: params[PARAM_SA_FLOW_PROPORTIONAL_THRESHOLD] * 100,
+            [PARAM_SA_LOW_VOLTAGE_PROPORTIONAL_THRESHOLD]: params[PARAM_SA_LOW_VOLTAGE_PROPORTIONAL_THRESHOLD] * 100,
+            [PARAM_SA_LOW_VOLTAGE_ABSOLUTE_THRESHOLD]: params[PARAM_SA_LOW_VOLTAGE_ABSOLUTE_THRESHOLD],
+            [PARAM_SA_HIGH_VOLTAGE_PROPORTIONAL_THRESHOLD]: params[PARAM_SA_HIGH_VOLTAGE_PROPORTIONAL_THRESHOLD] * 100,
+            [PARAM_SA_HIGH_VOLTAGE_ABSOLUTE_THRESHOLD]: params[PARAM_SA_HIGH_VOLTAGE_ABSOLUTE_THRESHOLD],
+        };
```
klesaulnier, line 200: move it to utils file just like toFormValuesLimitReductions()

src/components/dialogs/parameters/security-analysis/security-analysis-parameters.tsx
```diff
@@ -139,6 +187,53 @@ export const SecurityAnalysisParameters: FunctionComponent<{
...
+                    ...toFormValuesFromTemporaryLimits(vlLimits.temporaryLimitReductions),
+                };
+            }),
+            [PARAM_SA_FLOW_PROPORTIONAL_THRESHOLD]: params[PARAM_SA_FLOW_PROPORTIONAL_THRESHOLD] * 100,
+            [PARAM_SA_LOW_VOLTAGE_PROPORTIONAL_THRESHOLD]: params[PARAM_SA_LOW_VOLTAGE_PROPORTIONAL_THRESHOLD] * 100,
+            [PARAM_SA_LOW_VOLTAGE_ABSOLUTE_THRESHOLD]: params[PARAM_SA_LOW_VOLTAGE_ABSOLUTE_THRESHOLD],
+            [PARAM_SA_HIGH_VOLTAGE_PROPORTIONAL_THRESHOLD]: params[PARAM_SA_HIGH_VOLTAGE_PROPORTIONAL_THRESHOLD] * 100,
+            [PARAM_SA_HIGH_VOLTAGE_ABSOLUTE_THRESHOLD]: params[PARAM_SA_HIGH_VOLTAGE_ABSOLUTE_THRESHOLD],
+        };
+    }, [params]);
```
klesaulnier, line 223: same here if possible

src/components/dialogs/parameters/security-analysis/security-analysis-parameters.tsx
```diff
@@ -123,12 +164,19 @@ export const SecurityAnalysisParameters: FunctionComponent<{
         [params]
     );
 
-    const { handleSubmit, formState } = formMethods;
+    const { handleSubmit, formState, reset } = formMethods;
 
-    const updateLimitReductions = useCallback(
+    const updateSAParameters = useCallback(
         (formLimits: Record<string, any>) => {
```
klesaulnier, line 170: TODO maybe later : type this formLimits object since we now know what to expect according to yup schema

klesaulnier (commented):
small visual difference : 
in first tab, text in input is aligned to the left, it's aligned to the right in main branch

antoinebhs (changes requested):
application still crashes for different provider in the SA:
https://github.com/gridsuite/gridstudy-app/pull/2217/files#diff-a2606fc7df595e7f7429fce15248ca1f76b4ace6d7955f6bd3fbcf5da5e2f7e6R41
```
CardErrorBoundary caught:  TypeError: limits is null
    columnsDefinition limit-reductions-table-form.tsx:41
    React 3
    LimitReductionsTableForm limit-reductions-table-form.tsx:35
    React 9
    workLoop scheduler.development.js:266
    flushWork scheduler.development.js:239
    performWorkUntilDeadline scheduler.development.js:533
 ```

src/components/dialogs/parameters/security-analysis/security-analysis-parameters-selector.tsx
jamal-khey: i did a test, you are right, it is working with the old sollution.
i fixed it

klesaulnier (approved):

PR #2218, opened 2024-08-22
invalidate sa when parameters changes

jamal-khey:
this PR depend on [study server](https://github.com/gridsuite/study-server/pull/604)


jamal-khey:
fix to be done in the backend

PR #2223, opened 2024-08-26
AS parametter :  buttom bar backGround transparent on small screens

jamal-khey:
this PR fix visual anomaly on the buttom buttons.

before: 
image

after:
image


SlimaneAmar (approved):

souissimai (changes requested):
Adjustment needed for the size of the limit reduction table for SA: Not OK
==> make the width 100% for SA SecurityAnalysisParametersSelector

   <Grid sx={{ width: '100%' }}>
                                <Limi

souissimai (approved):

PR #2228, opened 2024-08-28
Fix reactive capability check when equipment not found

basseche:


Mathieu-Deharbe (approved):
Your correction revealed an other bug. I approved anyway because it is unlikely that you caused it but you should show this to a PO : 

Peek 29-08-2024 14-12

As you can what is supposed to be "validable" (0 in PMin) is not anymore after a (failed) modification of QminPmin and QmaxPmin.

basseche:
@Mathieu-Deharbe 
i've resolved the problem yo mentionned in the lase commit.

src/components/dialogs/reactive-limits/reactive-capability-curve/reactive-capability-curve-table.jsx
```diff
@@ -27,6 +28,17 @@ export const ReactiveCapabilityCurveTable = ({
...
+    // Respect minimum size (2 rows)
+    if (rows.length < MIN_LENGTH) {
+        for (let i = 0; i < MIN_LENGTH - rows.length; i++) {
+            rows.push({
+                [MAX_Q]: null,
+                [MIN_Q]: null,
+                [P]: null,
+            });
+        }
+    }
```
Mathieu-Deharbe, line 40:
```suggestion
   const handleInsertRow = useCallback(() => {
        if (previousValues && updatePreviousReactiveCapabilityCurveTable) {
            updatePreviousReactiveCapabilityCurveTable(INSERT, rows.length - 1);
        }
        insert(rows.length - 1, {
            [P]: null,
            [MIN_Q]: null,
            [MAX_Q]: null,
        });
    }, [updatePreviousReactiveCapabilityCurveTable, previousValues, rows.length, insert]);

    // Respect minimum size (2 rows)
    useEffect(() => {
        if (rows.length < MIN_LENGTH) {
            for (let i = 0; i < MIN_LENGTH - rows.length; i++) {
                handleInsertRow();
            }
        }
    }, [rows, handleInsertRow]);
```

Mathieu-Deharbe (changes requested):

src/components/dialogs/reactive-limits/reactive-capability-curve/reactive-capability-curve-table.jsx
basseche: Thanks i've added the useEffect only because i prefer push rather than insert

Mathieu-Deharbe (approved):

src/components/dialogs/reactive-limits/reactive-capability-curve/reactive-capability-utils.js
```diff
@@ -156,11 +156,11 @@ export const calculateCurvePointsToStore = (reactiveCapabilityCurve, equipmentTo
             if (point) {
                 let pointToStore = {
                     p: point?.[P],
-                    oldP: equipmentToModify.reactiveCapabilityCurveTable?.[index]?.p ?? null,
+                    oldP: equipmentToModify?.reactiveCapabilityCurveTable?.[index]?.p ?? null,
```
ghazwarhili, line 159: This version is not enough to safely handle equipmentToModify being null or undefined

I suggest to enhance this check by:

`const reactiveCapabilityCurveTable = equipmentToModify?.reactiveCapabilityCurveTable || [];` 
=> This avoids any issues with accessing properties on null or undefined

`const previousPoint = reactiveCapabilityCurveTable[index] || {};`
=> avoid redundancy by checking for equipmentToModify?.reactiveCapabilityCurveTable?.[index] once

(I have another remarqs about this function (you are not required to take these remarks into account) 
like using access to p, minQ, and maxQ directly, reducing redundancy by
`if (reactiveCapabilityCurve.every(({ p, minQ, maxQ }) => p == null && minQ == null && maxQ == null)) {
        return null;
    }`
also removing unnecessary else clause, replace foreach by map to iterate ...
it could be more efficient)

src/components/dialogs/reactive-limits/reactive-capability-curve/reactive-capability-curve-table.jsx
```diff
@@ -27,6 +28,19 @@ export const ReactiveCapabilityCurveTable = ({
 }) => {
     const { fields: rows, insert, remove } = useFieldArray({ name: `${id}` });
 
+    // Respect minimum size (2 rows)
+    useEffect(() => {
+        if (rows.length < MIN_LENGTH) {
+            for (let i = 0; i < MIN_LENGTH - rows.length; i++) {
+                rows.push({
```
ghazwarhili, line 35: Argument type {p: null, minQ: null, maxQ: null} is not assignable to parameter type FieldArrayWithId<FieldValues, string, "id">

useFieldArray expects the id field and the objects that you are trying to add do not include this id

To fix this error, i suggest to use the insert function like 

`insert(rows.length, {
                [MAX_Q]: null,
                [MIN_Q]: null,
                [P]: null,
            });`

ghazwarhili (changes requested):

src/components/dialogs/reactive-limits/reactive-capability-curve/reactive-capability-curve-table.jsx
basseche: modified

src/components/dialogs/reactive-limits/reactive-capability-curve/reactive-capability-curve-table.jsx
```diff
@@ -45,6 +46,22 @@ export const ReactiveCapabilityCurveTable = ({
...
+                    updatePreviousReactiveCapabilityCurveTable(INSERT, rows.length);
+                }
+
+                insert(rows.length, {
+                    [P]: null,
+                    [MIN_Q]: null,
+                    [MAX_Q]: null,
+                });
+            }
+        }
```
Mathieu-Deharbe, line 62: This is now a copy paste of `handleInsertRow`. Surprisingly sonar doesn't complain, but I think this should be refactored into a function.

Mathieu-Deharbe (approved):

ghazwarhili (approved):
Code review => OK
Test => OK

PR #2234, opened 2024-09-02
Display report containers as logs in the right panel

dbraquart:


src/utils/report-log.mapper.ts
```diff
@@ -8,27 +8,30 @@
...
     const highestSeverity = mapSeverity(report.severities ?? [REPORT_SEVERITY.UNKNOWN.name]);
-    // For now, we want to display only reports that are not containers in the "log" view
+    // We display a report line in the "log" view for both
+    // - a leaf (no sub-report)
+    // - and a container (have sub-reports), if its highest severity belongs to the severity filter (or unk)
     if (
-        report.subReports.length === 0 &&
-        highestSeverity.name !== REPORT_SEVERITY.UNKNOWN.name &&
-        report.parentId !== null
+        report.parentId &&
```
TheMaskedTurtle, line 23: I prefer to keep explicit check on null, in JS empty string is falsy for example so `if (report.parentId)` could mean a lot of different stuff

TheMaskedTurtle (approved):

src/utils/report-log.mapper.ts
dbraquart: I removed '!== null" on purpose to solve an issue ...

When we retrieve a report[] from the back, we build a meta-report called "Logs" having an undefined parent_id. Then we display a log "Logs" on the right side.

Now I set a null parent in the "Logs" report.

TheMaskedTurtle: oh ok, so maybe let it undefined and check report.parentId !== undefined ?

dbraquart: yes I was hesitating between the 2 options. Let's use undefined.

dbraquart: or "!= null" which works for both null and undefined

PR #2244, opened 2024-09-05
[SLD] highlighting overload in SLD

jamal-khey:
this PR adds an animation on the line in overload is the SLD

depend on [powsybl-diagram](https://github.com/powsybl/powsybl-diagram/pull/643)


src/index.css
```diff
@@ -493,6 +554,7 @@ code {
     --sld-vl-color: #abae28;
 }
 
+
```
AAJELLAL, line 557: unnecessary change.

src/index.css
```diff
@@ -240,6 +240,67 @@ code {
     dominant-baseline: middle;
 }
 
+@keyframes sld-overload-blink-animation {
```
AAJELLAL, line 243: Duplication to be removed.

AAJELLAL (commented):
One duplication to be removed.

AAJELLAL (approved):
Tests are ok. Test was done with VL and 2WT overload and overvoltage.

src/index.css
jamal-khey: done

PR #2254, opened 2024-09-11
Update contingency list name after renaming

souissimai:


jamal-khey (approved):
Looks good to me.
Code and test Ok.
console warning checked

PR #2256, opened 2024-09-12
add filtering on report msg

AbdelHedhili:


src/components/report-viewer/report-viewer.jsx
```diff
@@ -56,59 +62,57 @@ export default function ReportViewer({ report, reportType }) {
...
+    useEffect(() => {
+        refreshLogsOnSelectedReport(selectedReportId, severityFilter, messageFilter);
+    }, [messageFilter, severityFilter, selectedReportId, refreshLogsOnSelectedReport]);
+
     useEffect(() => {
         const reportTree = mapReportsTree(report);
         treeView.current = initializeTreeDataAndComponent(reportTree);
-        setSelectedReportId(report.id);
         setExpandedTreeReports([report.id]);
         setLogs(mapReportLog(report, reportTree.severities));
```
TheMaskedTurtle, line 100: You can remove that then the method mapReportLog will be unused so you can delete it

src/hooks/use-report-fetcher.tsx
```diff
@@ -104,28 +114,22 @@ export const useReportFetcher = (
...
-                        severityList,
-                        computingAndNetworkModificationType
-                    );
+                fetchPromise = (severityList: string[], reportId: string) =>
+                    fetchNodeReportLogs(studyUuid, currentNode!.id, reportId, severityList, messageFilter, false);
             }
-            return fetch(() => fetchPromise(severityList, reportId)).then((r) =>
-                r === undefined ? undefined : mapReportLog(r, severityList)
-            );
+            return fetchPromise(severityList, reportId).then((r) => {
```
TheMaskedTurtle, line 127: Here you could also call mapReportLogs and return Log[] directly

src/components/report-viewer/text-filter-button.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * FilterButton wraps a MultiSelectList with a button which has a visual indication to indicate when the user alters the default state of the list
+ * @param {String} filterText - Value of the filter
+ * @param {Function} setFilterText - Setter needed to update the filter
+ */
+
+interface TextFilterButtonProps {
+    filterText: string;
+    setFilterText: Dispatch<SetStateAction<string>>;
```
TheMaskedTurtle, line 40:
```suggestion
    setFilterText: (text: string) => void;
```
This is enough from TextFilterButton perspective

src/components/report-viewer/text-filter-button.tsx
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export const TextFilterButton: FunctionComponent<TextFilterButtonProps> = ({ filterText, setFilterText }) => {
+    const intl = useIntl();
+
+    const [anchorEl, setAnchorEl] = useState<HTMLButtonElement | null>();
```
TheMaskedTurtle, line 46: You can initialize with null, so we don't have to manage both null and undefined

TheMaskedTurtle (approved):

PR #2269, opened 2024-09-23
Fix the color of disconnected lines (on both edges) in NAD

EstherDarkish:


antoinebhs (approved):
Code OK
Test OK, need the line to be disconnected on both sides

PR #2282, opened 2024-09-30
add side for one bus shortCircuit result

souissimai:


src/components/results/shortcircuit/shortcircuit-analysis-result-table.tsx
```diff
@@ -281,6 +290,13 @@ const ShortCircuitAnalysisResultTable: FunctionComponent<ShortCircuitAnalysisRes
 
                 const current = getCurrent(faultResult);
 
+                const convertSide = (side: string | undefined) => {
+                    return side === BranchSide.ONE
+                        ? intl.formatMessage({ id: 'Side1' })
+                        : side === BranchSide.TWO
+                        ? intl.formatMessage({ id: 'Side2' })
+                        : undefined;
+                };
```
YenguiSeddik, line 299: I think this function can be outside the component:
function convertSide(side?: string) {
    switch (side) {
        case BranchSide.ONE:
            return'Side1';
        case BranchSide.TWO:
            return 'Side2'
        default:
            return undefined;
    }
}; or as const const sideMap = {
    [BranchSide.ONE]: 'Side1',
    [BranchSide.TWO]:'Side2',
}; andthen
 side: intl.formatMessage({ id: convertSide(side) })

souissimai: No need to duplicate it I noticed that it's used by LF and SA  also I just imported it

capyq (approved):

YenguiSeddik (approved):

PR #2291, opened 2024-10-04
Fixes the position-diagram header

EstherDarkish:


jamal-khey (approved):
code OK
tested : SLD via Map, SDL via search bar, SLD via modifications forms

PR #2301, opened 2024-10-08
fix equipment modification in spreadsheet

anistouri:
Bug fix related to the modification of certain equipment values in the spreadsheet. The error originated from the backend, caused by changes made to the parameters of certain functions, without these changes being properly reflected in all their calls. A slight refactoring was also carried out, which now allows the parameters to be called in any order


dbraquart (approved):
code: ok
tests: ok all eqpt validated on both spreadsheet and modifications

rq: 1 new issue is reported by Sonar, but I dont see why

PR #2305, opened 2024-10-09
Fix: warnings caused by Divider used as Tabs child

sBouzols:
A good solution is to use a disabled Tab component with a Divider icon en some sx

fix some warnings like : 
Warning: Received `false` for a non-boolean attribute `indicator`.
Warning: React does not recognize the `selectionFollowsFocus` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `selectionfollowsfocus` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
 Warning: React does not recognize the `fullWidth` prop on a DOM element.
Warning: React does not recognize the `textColor` prop on a DOM element.
...


src/components/parameters-tabs.tsx
```diff
@@ -377,7 +381,7 @@ const ParametersTabs: FunctionComponent<OwnProps> = (props) => {
                                 label={<FormattedMessage id="VoltageInit" />}
                                 value={TAB_VALUES.voltageInitParamsTabValue}
                             />
-                            <Divider />
+                            <Tab sx={styles.dividerTab} label="" icon={<Divider sx={{ flexGrow: 1 }} />} disabled />
```
Meklo, line 384: Maybe add a comment to explain the syntax so we don't reproduce the issue elsewhere ? To make obvious that instantiating a component that's not a Tab directly under a Tabs will induce console warnings

Meklo (approved):

PR #2309, opened 2024-10-09
Fix warning provider init in Select component for Sensi

sBouzols:
warning: MUI: You have provided an out-of-range value `null` for the select component.
Consider providing a value that matches one of the available options or ''. The available values are `OpenLoadFlow`.


Meklo (approved):
Also fixes "Warning: `value` prop on `input` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.
input"

PR #2310, opened 2024-10-09
Fix(TableRowComponent): warning unique key

sBouzols:
Warning: Each child in a list should have a unique "key" prop.


Meklo (approved):

Meklo (changes requested):
You should also remove the prop `row` from TableRowComponent since it has no longer any usage

Meklo (approved):

PR #2330, opened 2024-10-17
 load geo data when initialized

souissimai:


jamal-khey (approved):
Looks good to me. tests Ok

PR #2338, opened 2024-10-21
Fix invalid hook call while using flux convention

dbraquart:


FranckLecuyer (changes requested):
Tests: the filter 'not equal' does not apply correctly : it is equivalent to 'greater than or equal to' ...

FranckLecuyer (changes requested):
Tests: NOK

ayolab (approved):
Tests : OK

FranckLecuyer (approved):

PR #2340, opened 2024-10-22
Feat : create new spreadsheet

ayolab:


src/components/spreadsheet/add-new-spreadsheet/add-new-spreadsheet-dialog.tsx
```diff
@@ -0,0 +1,204 @@
...
+    const { handleSubmit, reset } = formMethods;
+
+    const onSubmit = useCallback(
+        (newParams: any) => {
+            if (selectedOption?.id === NEW_SPREADSHEET_CREATION_OPTIONS.EMPTY.id) {
+                const equipmentType = newParams.equipmentType as keyof typeof TABLES_DEFINITIONS;
+                const newTableDefinition = {
+                    ...TABLES_DEFINITIONS[equipmentType],
+                    name: newParams[SPREADSHEET_NAME],
+                    index: tablesDefinitionIndexes.size,
```
anistouri, line 90: The created spreadsheet should be added at the beginning of the list of tabs

src/components/spreadsheet/add-new-spreadsheet/add-new-spreadsheet-dialog.tsx
```diff
@@ -0,0 +1,204 @@
...
+                        []
+                    )
+                );
+            } else {
+                getSpreadsheetModel(newParams[SPREADSHEET_MODEL][0].id).then(
+                    (selectedModel: { customColumns: any; sheetType: keyof typeof TABLES_DEFINITIONS }) => {
+                        const newTableDefinition = {
+                            ...TABLES_DEFINITIONS[selectedModel.sheetType],
+                            name: newParams[SPREADSHEET_NAME],
+                            index: tablesDefinitionIndexes.size,
```
anistouri, line 105: Same remark above

src/components/spreadsheet/add-new-spreadsheet/add-new-spreadsheet-dialog.tsx
```diff
@@ -0,0 +1,204 @@
...
+        </Grid>
+    );
+
+    const addSpreadsheetFromModelContent = (
+        <Grid container spacing={2} direction="column">
+            <Grid item xs>
+                <TextInput
+                    name={SPREADSHEET_NAME}
+                    label={'spreadsheet/create_new_spreadsheet/spreadsheet_name'}
+                    formProps={{ autoFocus: true }}
```
anistouri, line 156: If the template is selected, the Name field does not take the name of the template

anistouri (changes requested):

src/redux/actions.ts
```diff
@@ -1191,3 +1192,16 @@ export function setCustomColumDefinitions(
         filter: filter,
     };
 }
+
+export const SET_TABLE_DEFINITIONS = 'SET_TABLE_DEFINITIONS';
+export const UPDATE_TABLE_DEFINITION = 'UPDATE_TABLE_DEFINITION';
+
+export type UpdateTableDefinitionAction = {
+    type: typeof UPDATE_TABLE_DEFINITION;
+    payload: { key: string; value: TablesDefinitionsType; customColumns: any };
```
anistouri, line 1201:
```suggestion
    payload: { key: string; value: TablesDefinitionsType; customColumns: ColumnWithFormula[] };
```

src/redux/actions.ts
```diff
@@ -1191,3 +1192,16 @@ export function setCustomColumDefinitions(
...
+};
+
+export const updateTableDefinition = (key: string, value: any, customColumns: any): UpdateTableDefinitionAction => ({
```
anistouri, line 1204:
```suggestion
export const updateTableDefinition = (key: string, value: any, customColumns: ColumnWithFormula[]): UpdateTableDefinitionAction => ({
```

src/components/spreadsheet/add-new-spreadsheet/add-new-spreadsheet-dialog.tsx
```diff
@@ -0,0 +1,204 @@
...
+                dispatch(
+                    updateTableDefinition(
+                        'new' + tablesDefinitionIndexes.size + newParams.equipmentType,
+                        newTableDefinition,
+                        []
+                    )
+                );
+            } else {
+                getSpreadsheetModel(newParams[SPREADSHEET_MODEL][0].id).then(
+                    (selectedModel: { customColumns: any; sheetType: keyof typeof TABLES_DEFINITIONS }) => {
```
anistouri, line 101:
```suggestion
                    (selectedModel: { customColumns: ColumnWithFormula[]; sheetType: keyof typeof TABLES_DEFINITIONS }) => {
```

src/components/spreadsheet/add-new-spreadsheet/add-new-spreadsheet-dialog.tsx
ayolab: seen with the PO, we can keep it like this for now

src/components/spreadsheet/add-new-spreadsheet/add-new-spreadsheet-dialog.tsx
ayolab: same as above

FranckLecuyer (changes requested):
Tests: 
After adding a new empty spreadsheet, nothing is displayed when we select this new spreadsheet (only 'Loading...' is shown) : we have to select first another existing spreadsheet and then to select this newly created spreadsheet in order to see the data (for the predefined fixed columns selected by default).

No error is shown when adding a new spreadsheet from a template seen in the directory tree, but that no longer exists ...

After adding a new spreadsheet named 'titi', error messages are shown in the console : 
Error: [@formatjs/intl Error MISSING_TRANSLATION] Missing message: "titi" for locale "en", using id as fallback.

After adding a new spreadsheet from template, all the columns are custom columns (ok), but the 'Show / id columns' button displays an empty column list ...maybe this button should be disabled ...

FranckLecuyer (approved):

anistouri (approved):

PR #2350, opened 2024-10-24
display only xiidm format when developer mode is disabled

ghazwarhili:


src/components/dialogs/export-dialog.jsx
```diff
@@ -75,21 +77,24 @@ const ExportDialog = ({ open, onClose, onClick, studyUuid, nodeUuid, title }) =>
     useEffect(() => {
         if (open) {
             getAvailableExportFormats().then((formats) => {
+                let availableFormats = enableDeveloperMode
```
basseche, line 80: you don't change this variable. Why don't you use const  instead of let ?

src/components/dialogs/export-dialog.jsx
```diff
@@ -75,21 +77,24 @@ const ExportDialog = ({ open, onClose, onClick, studyUuid, nodeUuid, title }) =>
...
+                    ? formats
+                    : Object.fromEntries(Object.entries(formats).filter(([key]) => key === 'XIIDM'));
```
basseche, line 82: use constante for 'XIIDM' Don't hardcode

basseche (approved):

PR #2354, opened 2024-10-25
fix negative number for section in voltage<Level reztion form

ghazwarhili:


etiennehomer (changes requested):
Row 97 of this file, change the condition : 
if (isNaN(watchSectionCount) || watchSectionCount <= 1) {

ghazwarhili:
Okay, I can do it, but it doesn’t relate to the bug related t this PR. Otherwise, could you tell me the bug so I can adjust your proposal?

etiennehomer:
> Okay, I can do it, but the changes request doesn’t related to this PR. Otherwise, could you tell me the bug so I can adjust your proposal?

It's kinda related to it. The "Switches between sections" field is visible when you type '-' in "Sections count". It should be hidden.

ghazwarhili:
isNaN(watchSectionCount) : Argument type DeepPartialSkipArrayKey<FieldValues> is not assignable to parameter type number 

when I tried to set "-" in sectionsCount field 
Capture d’écran du 2024-10-30 16-10-20

what is the bug that you found to be fixed by this check isNaN(watchSectionCount)  ?

etiennehomer (approved):

PR #2365, opened 2024-10-29
Enable creation of a spreadsheet on busbar sections.

FranckLecuyer:


src/components/spreadsheet/equipment-tabs.tsx
```diff
@@ -41,6 +42,12 @@ export const EquipmentTabs: FunctionComponent<EquipmentTabsProps> = ({ tabIndex,
                 >
                     {tablesNames.map((table) => (
                         <Tab
+                            sx={{
```
achour94, line 45: Instead of handling this with css, you can just filter the tables names ?

`
  const visibleTables = tablesNames.filter(
    (table) => table !== TABLES_DEFINITIONS.BUSBAR_SECTION.name
  );
`

achour94: With this logic, we can't create an empty spreadsheet named 'BusBarSections.' We will receive the error message 'Spreadsheet name already exists,' even though it's not visible. (To discuss with the Product Owner)

FranckLecuyer: Busbar sections table is now always visible

PR #2366, opened 2024-10-29
Fix EllipsisCellRenderer : take into consideration zoom and window resize

ayolab:


src/components/spreadsheet/utils/cell-renderers.tsx
```diff
@@ -163,13 +163,33 @@ export const DefaultCellRenderer = (props: any) => {
...
+        };
+
+        window.addEventListener('resize', handleResize);
+        window.addEventListener('zoom', handleResize);
+
+        return () => {
+            window.removeEventListener('resize', handleResize);
+            window.removeEventListener('zoom', handleResize);
+        };
+    }, []);
```
anistouri, line 191: It is better to use ResizeObserver because it specifically observes changes in the size of cells, without relying on global window events

`const checkEllipsis = () => {
        if (textRef.current) {
            const zoomLevel = window.devicePixelRatio;
            const adjustedScrollWidth = textRef.current.scrollWidth / zoomLevel;
            const adjustedClientWidth = textRef.current.clientWidth / zoomLevel;
            setIsEllipsisActive(adjustedScrollWidth > adjustedClientWidth);
        }
    };

    useEffect(() => {
        checkEllipsis();
        const resizeObserver = new ResizeObserver(() => checkEllipsis());
        if (textRef.current) {
            resizeObserver.observe(textRef.current);
        }

        return () => {
            resizeObserver.disconnect();
        };
    }, [value]);`

anistouri (approved):

PR #2384, opened 2024-11-12
Remove unrecognized props from MUI Dialog

Meklo:
Scenario to replicate: Open a modification creation dialog for a Voltage Level most of those dialogs are impacted 

`onValidated` and `onValidationError` are already handled in `ModificationDialog` so there's no need to pass them down to `ModificationDialogContent` thus fixing the following warnings when they get to MUI `Dialog` : 

- Unknown event handler property `onValidationError`. It will be ignored.
- Unknown event handler property `onValidated`. It will be ignored.


dbraquart (approved):
code: ok
tests: ok, warning is gone

PR #2386, opened 2024-11-12
Fixes sensitivity-analysis-tabs warnings

Meklo:
Fixes the following warnings when opening sensitivity analysis results tab :
    Warning: Failed prop type: Invalid prop `rowsPerPageOptions[3]` of type `object` supplied to `CustomTablePagination`, expected `number`.
    Warning: Each child in a list should have a unique "key" prop.

Enriches CustomTablePagination's rowsPerPage prop proptype by accounting for { label:string, value: number } instead of exclusively numbers 
Adds a key to the results tabs


dbraquart (approved):
code: ok
tests: warnings are gone

PR #2393, opened 2024-11-14
Fix sensi analysis parameters warning

Meklo:
Prevents the following warning :

    Warning: validateDOMNesting(...): <th> cannot appear as a child of <thead>.

By following `TableHead` -> `<thead>` of `SensitivityTable` with a `TableRow` -> `<tr>` before inserting a `TableCell` -> `<th>` as it is advised here https://developer.mozilla.org/en-US/docs/Web/HTML/Element/thead


dbraquart (approved):
ok warning removed

PR #2397, opened 2024-11-15
Fix sensitivity analysis parameters warning

Meklo:
Fixes the following warning when opening the sensitivity analysis parameters panel :

    Warning: React does not recognize the `selectionFollowsFocus` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute,
    Warning: Received `false` for a non-boolean attribute `indicator`.
    Warning: React does not recognize the `fullWidth` prop on a DOM element.

By removing non `Tab` elements under `Tabs` component


dbraquart (approved):
ok warnings removed

PR #2398, opened 2024-11-15
Fix csv export filename not translatable warning

Meklo:
Fixes this type of warnings when exporting a CSV dataset whose name is not translatable :

    Tab Current Violations from Loadflow results :
        Error: [@formatjs/intl Error MISSING_TRANSLATION] Missing message: "Current violations" for locale "en", using id as fallback.
    Tab Voltage violations from Loadflow results :
        Error: [@formatjs/intl Error MISSING_TRANSLATION] Missing message: "Voltage violations" for locale "en", using id as fallback.
    Tab Status from Loadflow results:
        Error: [@formatjs/intl Error MISSING_TRANSLATION] Missing message: "Status" for locale "en", using id as fallback.

By checking if the provided filename is in intl messages, if it's not, use it raw. If it's undefined a default filename is used instead

### DIFF
--- a/src/components/spreadsheet/export-csv.tsx
+++ b/src/components/spreadsheet/export-csv.tsx
@@ -22,6 +22,8 @@ interface CsvExportProps {
     skipColumnHeaders?: boolean;
 }
 
+const DEFAULT_FILENAME = 'csv-export';
+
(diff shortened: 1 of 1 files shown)


Meklo:
Fixed by https://github.com/gridsuite/gridstudy-app/pull/2399

PR #2408, opened 2024-11-20
View state estimation balance results (quality, status, quality criterion, quality per region)

FranckLecuyer:


src/redux/actions.ts
```diff
@@ -1251,3 +1253,19 @@ export const addSortForNewSpreadsheet = (
...
+export function setStateEstimationResultFilter(
+    filterTab: keyof AppState[typeof STATEESTIMATION_RESULT_STORE_FIELD],
+    stateEstimationResultFilter: MutableUnknownArray
+): StateEstimationResultFilterAction {
+    return {
+        type: STATEESTIMATION_RESULT_FILTER,
+        filterTab: filterTab,
+        [STATEESTIMATION_RESULT_STORE_FIELD]: stateEstimationResultFilter,
+    };
+}
```
dbraquart, line 1271: if it does not work and is not requested by PO, maybe we should rollback everything dealing with sort/filter in redux ?

src/components/results/stateestimation/state-estimation-status-result.tsx
```diff
@@ -0,0 +1,50 @@
...
+                            <FormattedMessage id="StateEstimationQuality" />
+                        </Typography>
+                        <Typography sx={styles.valueTypography}>{result.qualityLevel}</Typography>
+                    </Stack>
+                )}
+            </>
+        );
+    };
+
+    return <>{renderStateEstimationStatusResult()}</>;
```
dbraquart, line 49: we could directly return the JSX code.

src/components/results/stateestimation/state-estimation-quality-result.tsx
```diff
@@ -0,0 +1,134 @@
...
+    const intl = useIntl();
+    const gridRef = useRef<AgGridReact>(null);
+
+    const tableNameFormatted = intl.formatMessage({ id: tableName });
+
+    const stateEstimationStatus = useSelector(
+        (state: AppState) => state.computingStatus[ComputingType.STATE_ESTIMATION]
+    );
+
+    const [isQualityReady, setIsQualityReady] = useState(false);
```
dbraquart, line 44: I dont understand this state. Seems to be always false ?

dbraquart (changes requested):
code: What to do with sort/filter if it does not work ? clean the code, fix later ?)
tests: ok (except sort/filter)

src/components/results/stateestimation/state-estimation-status-result.tsx
FranckLecuyer: Done

src/redux/actions.ts
FranckLecuyer: Sorting and filtering removed

dbraquart (approved):
code: ok

tests: ok

PR #2411, opened 2024-11-21
Fix NetworkMapTab: geodata and mapdata loading at start (manual mode)

sBouzols:
reactive geodata and mapdata loading at start in manual refresh mode.


src/components/network/network-map-tab.tsx
```diff
@@ -144,6 +144,8 @@ export const NetworkMapTab = ({
     const intlRef = useIntlRef();
     const [isRootNodeGeoDataLoaded, setIsRootNodeGeoDataLoaded] = useState(false);
     const [isInitialized, setInitialized] = useState(false);
+    const isInitializedRef = useRef<boolean>();
+    isInitializedRef.current = isInitialized;
```
klesaulnier, line 148: I think it's not used

src/components/network/network-map-tab.tsx
```diff
@@ -811,9 +812,21 @@ export const NetworkMapTab = ({
...
+            if (
+                isSameNode(previousCurrentNode, currentNode) && // must be the same node
+                !previousNodeStatus && // must pass from unbuilt ...
+                isCurrentNodeBuiltRef.current // to built
+            ) {
+                updateMapEquipmentsAndGeoData();
+            }
+            // return to avoid update in manual reload or to avoid double upate
+            return;
         }
```
klesaulnier, line 827:
```suggestion
        if (!(refIsMapManualRefreshEnabled.current && isInitialized
                isSameNode(previousCurrentNode, currentNode) && // must be the same node
                !previousNodeStatus && // must pass from unbuilt ...
                isCurrentNodeBuiltRef.current // to built)
        ) {
            return;
        }
```

I would make something like this to keep the same structure as the rest of the useEffect
The condition I've suggested might be wrong, it's mostly to illustrate my point
Of course, keep the comments as you did before

klesaulnier (commented):
will test

src/components/network/network-map-tab.tsx
sBouzols: removed

src/components/network/network-map-tab.tsx
sBouzols: I prefer the previous version to manage clearly the manual mode to avoid `updateMapEquipmentsAndGeoData` and it's exception inside when node is built.

klesaulnier (approved):

src/components/network/network-map-tab.tsx
```diff
@@ -811,9 +810,21 @@ export const NetworkMapTab = ({
...
+        if (refIsMapManualRefreshEnabled.current && isInitialized) {
+            // this reloads the mapEquipments when, in manual refresh mode, the current node is built.
+            if (
+                isSameNode(previousCurrentNode, currentNode) && // must be the same node
+                !previousNodeStatus && // must pass from unbuilt ...
+                isCurrentNodeBuiltRef.current // to built
+            ) {
+                updateMapEquipmentsAndGeoData();
+            }
+            // return to avoid update in manual reload or to avoid double upate
```
EstherDarkish, line 823:
```suggestion
            // return to avoid update in manual reload or to avoid double update
```

src/components/network/network-map-tab.tsx
```diff
@@ -811,9 +810,21 @@ export const NetworkMapTab = ({
...
         }
-        if (isRootNodeGeoDataLoaded && isMapEquipmentsInitialized) {
-            updateMapEquipmentsAndGeoData();
+        // manual reload
+        if (refIsMapManualRefreshEnabled.current && isInitialized) {
+            // this reloads the mapEquipments when, in manual refresh mode, the current node is built.
+            if (
+                isSameNode(previousCurrentNode, currentNode) && // must be the same node
+                !previousNodeStatus && // must pass from unbuilt ...
+                isCurrentNodeBuiltRef.current // to built
```
EstherDarkish, line 819:
```suggestion
                !previousNodeStatus && // must change from unbuilt ...
                isCurrentNodeBuiltRef.current // ... to built
```

EstherDarkish (approved):

PR #2416, opened 2024-11-25
Implement custom report tree for report viewer

Meklo:


src/components/report-viewer/treeview-item.tsx
```diff
@@ -0,0 +1,122 @@
...
+    data: TreeViewItemData;
+    index: number;
+    style: CSSProperties;
+}
+
+export const TreeviewItem: FunctionComponent<TreeViewItemProps> = (props) => {
+    const { data, index } = props;
+    const { nodes, onSelectedItem, onExpandItem, highlightedReportId } = data;
+    const currentNode = nodes[index];
+    const left = currentNode.depth * 12;
```
jonenst, line 98: name this constant

Meklo: named it `ITEM_DEPTH_OFFSET`

src/components/report-viewer/report-viewer.tsx
```diff
@@ -34,55 +38,54 @@ export default function ReportViewer({ report, reportType }: ReportViewerProps)
...
+        reportTreeData.current[item.id] = item;
+        for (let subReport of item.subReports) {
+            mapReportsById(subReport);
+        }
     }, []);
 
+    const toTreeNodes = useCallback(
+        (item: ReportTreeType, depth: number): ReportItem[] => {
+            const result: ReportItem[] = [];
+            const collapsed = !expandedTreeReports.includes(item.id);
```
jonenst, line 55: collapsed handling code should be moved to the virtaulized-treeview component

src/components/report-viewer/report-viewer.tsx
```diff
@@ -96,6 +99,30 @@ export default function ReportViewer({ report, reportType }: ReportViewerProps)
...
+                setSelectedReportId(report.id);
+                setSelectedReportType(reportTreeData.current[report.id].type);
+            }
+        },
+        [selectedReportId]
+    );
+
+    const handleExpandItem = useCallback(
+        (node: ReportItem) => {
+            if (node.collapsed) {
```
jonenst, line 114: collapsed handling code should be moved to the virtaulized-treeview component if it's not too complicated

src/components/report-viewer/report-viewer.tsx
```diff
@@ -34,68 +25,50 @@ export default function ReportViewer({ report, reportType }: ReportViewerProps)
...
-                labelIconColor={report.severity.colorName}
-                key={report.id}
-                sx={styles.treeItem}
-                nodeId={report.id}
-            >
-                {report.subReports.map((value: ReportTreeType) => initializeTreeDataAndComponent(value))}
-            </ReportItem>
-        );
+    const mapReportsById = useCallback((item: ReportTree) => {
+        reportTreeMap.current[item.id] = item;
```
jonenst, line 32: should maybe throw away the old map and assign a new empty object to reportTreeMap.current ?

src/components/report-viewer/report-viewer.tsx
```diff
@@ -34,68 +25,50 @@ export default function ReportViewer({ report, reportType }: ReportViewerProps)
...
     };
 
+    const handleSelectedItem = useCallback(
+        (report: ReportItem) => {
+            if (selectedReportId !== report.id) {
+                setSelectedReportId(report.id);
+                setSelectedReportType(reportTreeMap.current[report.id].type);
+            }
+        },
+        [selectedReportId]
```
jonenst, line 69: maybe in a separate PR we can try to refactor this code a bit. I find it weird for example that we go to the extend of using a ref for the map of ids to reports to avoid having it in the dependencies of this callback, but at the same time it has selectedReportId in its dependencies which presumably changes more often than the reportTree (and maybe even always changes when the reporttree is changed?).
Another weird thing is top use useCallback for callbacks for VirtualizedTreeview, but not for callbacks for LogTable (onLogRowClick is not a useCallback).. To be cleaned up later I guess

jonenst (approved):

PR #2434, opened 2024-12-04
Clean up custom-aggrid-header

Meklo:
Cleans up and deconstruct custom-aggrid-header component along the following lines :
- Typescript migration 
- Changes the way the handler of sort changes is pass down the header : before, onSortChanged was overriden in `SensitivityAnalysisResult` and `makeAgGridCustomHeaderColumn` to wrap the function so that the header only calls the method with the new SortWay, now it is passed as is and the formatting of the data is done inside useCustomAggridSort in order to simplify the typing of `sortParams`
- All logic related to filters and sorting has been put in custom hooks, respectively `useCustomAggridFilter` and `useCustomAggridSort` so that it clears up the components and ease up re-usability : for example the sort hook is called both from `custom-aggrid-header` and `CustomAggridSort` so that we can implement the sort behaviour when clicking on the header and on the sorting icon itself
- Icons and dedicated menus for sorting, filters and custom columns have been grouped inside their respective component : `CustomAggridSort`, `CustomAggridFilter` and `CustomMenu`
- The layout has been reworked to reduce duplication of code and ease up the placement of elements in the header : use of properties `flex` and `flexBasis` to assign weight to the elements in the row
- Each type of filter has been put in a dedicated component and all are managed from a filter wrapper called `CustomAggridFilter`, the wrapper instantiate a determined filter component in function of the defined tests : 
   - Autocomplete filters for enums
   - Boolean filters previously existed
   - Comparator filters for those which depend on firstly selecting a type of comparator (equal, not equal, contains...) before filling value which is either text based or duration base 
- Regroups all properties related to filtering, sorting and custom menu and better global readability and filters modularity
- Changes when `decimalAfterDot` is computed : before it was included in the handleFilterTextChange event handler, it has been moved to a useMemo because previously the value of decimalAfterDot persisted when the Popover was closed because its state was stored outside of it, with the new structure it has to be computed again when the Popover is opened again


ayolab (approved):

ayolab (approved):

src/components/custom-aggrid/custom-aggrid-filters/custom-aggrid-comparator-selecter.tsx
```diff
@@ -0,0 +1,43 @@
...
+    },
+};
+
+interface CustomAggridComparatorSelecterProps {
+    value: string;
+    onChange: (event: SelectChangeEvent) => void;
+    options: string[];
+}
+
+export const CustomAggridComparatorSelecter: React.FC<CustomAggridComparatorSelecterProps> = ({
```
TheMaskedTurtle, line 25: Selector ?

src/components/custom-aggrid/custom-aggrid-filters/custom-aggrid-autocomplete-filter.tsx
```diff
@@ -0,0 +1,61 @@
...
+    const { selectedFilterData, handleChangeFilterValue } = useCustomAggridFilter(field, filterParams);
+
+    const handleFilterAutoCompleteChange = (_: SyntheticEvent, data: string[]) => {
+        handleChangeFilterValue({ value: data, type: FILTER_TEXT_COMPARATORS.EQUALS });
+    };
+
+    const filterOption = useMemo(
+        () => (filterEnums?.hasOwnProperty(field) ? filterEnums[field] ?? [] : []),
+        [field, filterEnums]
+    );
```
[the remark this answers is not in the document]
Meklo: Done

src/components/custom-aggrid/custom-aggrid-filters/custom-aggrid-comparator-selecter.tsx
Meklo: Renamed

TheMaskedTurtle (approved):

PR #2444, opened 2024-12-11
Fixes an issue with Chromium's rendering of percentage based stroke-w…

EstherDarkish:
…idth polyline values.

To use with https://github.com/powsybl/powsybl-network-viewer/pull/98


EstherDarkish:
Another solution was merged for this issue. Closing the PR.

PR #2449, opened 2024-12-12
Use study-config-server to store network visualization parameters per study

dbraquart:


src/services/study/study-config.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright (c) 2023, RTE (http://www.rte-france.com)
```
ayolab, line 2:
```suggestion
 * Copyright (c) 2024, RTE (http://www.rte-france.com)
```

ayolab (approved):

PR #2469, opened 2024-12-24
feat: reference nodes in custom columns spreadsheet

AbdelHedhili:


AbdelHedhili:
Known possible improvement: 
for now, we redo the fetch to get the information of other nodes whenever we get to a tab that uses a custom column with a node because we don't know if the data is still up-to-date. We could use notification to only redo fetches when necessary.

src/components/spreadsheet/table-wrapper.tsx
```diff
@@ -494,8 +509,37 @@ const TableWrapper: FunctionComponent<TableWrapperProps> = ({
...
+            const nodeAlias: string = value[0];
+            const equipmentsToAdd = value[1][equipmentType];
+            if (equipmentsToAdd) {
+                equipmentsToAdd.forEach((equipmentToAdd) => {
+                    let matchingEquipmentIndex = equipmentsWithCustomColumnInfo.findIndex(
+                        (equipmentWithCustomColumnInfo) => equipmentWithCustomColumnInfo.id === equipmentToAdd.id
+                    );
+                    let matchingEquipment = equipmentsWithCustomColumnInfo[matchingEquipmentIndex];
+                    if (matchingEquipment) {
+                        let equipmentWithAddedInfo: any = { ...matchingEquipment };
```
Meklo, line 527: I reckon it could be typed as follow :
`export interface RecursiveIdentifiable extends Identifiable {
    [alias: string]: Identifiable | unknown;
}
`

src/components/spreadsheet/use-spreadsheet-equipments.ts
```diff
@@ -126,6 +133,15 @@ export const useSpreadsheetEquipments = (
         resetImpactedElementTypes,
     ]);
 
+    const toSpreadsheetEquipmentType = useCallback(
+        (tableName: string): SpreadsheetEquipmentType => {
+            return tablesDefinitions.filter(
```
Meklo, line 138: Same comment as for `toSpreadsheetEquipmentType` in `table-wrapper`

src/components/spreadsheet/use-spreadsheet-equipments.ts
```diff
@@ -137,14 +153,85 @@ export const useSpreadsheetEquipments = (
                         fetchedEquipments = formatFetchedEquipments(fetchedEquipments);
                     }
                     dispatch(loadEquipments(equipment.type, fetchedEquipments));
-                    setIsFetching(false);
```
Meklo, line 140: Why this removal ?

src/components/spreadsheet/custom-columns/use-custom-column.ts
```diff
@@ -73,8 +73,7 @@ export function useCustomColumn(tabIndex: number) {
                         acc[key] = typeof value === 'number' ? bignumber(value) : value;
                         return acc;
                     }, {} as Record<string, unknown>);
-
-                    return math.limitedEvaluate(colWithFormula.formula, scope);
+                    return math.limitedEvaluate(colWithFormula.formula ?? '', scope);
```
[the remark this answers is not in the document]
AbdelHedhili: its a remaining of a previous solution that i removed but forgot this

src/components/spreadsheet/table-wrapper.tsx
```diff
@@ -485,6 +491,15 @@ const TableWrapper: FunctionComponent<TableWrapperProps> = ({
         });
     }, [sortConfig, mergedColumnData]);
 
+    const toSpreadsheetEquipmentType = useCallback(
+        (tableName: string): SpreadsheetEquipmentType => {
+            return tablesDefinitions.filter(
```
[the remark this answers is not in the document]
AbdelHedhili: The thing is it's never supposed to be null since every equipment has a type but yeah if for whatever reason this happens it's better to not crash. I guess in this case we could just do nothing (not updating the rowdata)

src/components/spreadsheet/use-spreadsheet-equipments.ts
```diff
@@ -144,7 +159,80 @@ export const useSpreadsheetEquipments = (
...
+            if (spreadsheetEquipmentType) {
+                customColumns.forEach((column) => {
+                    const pattern: string = aliases.map((alias) => `\\b${alias}\\b\\.`).join('|'); // pattern to find the aliases in the expression
+                    const regex = new RegExp(pattern, 'g');
+                    const aliasesFoundInFormula = column.formula?.match(regex);
+                    aliasesFoundInFormula?.forEach((aliasWithDot) => {
+                        const alias = aliasWithDot.slice(0, -1); //because we searched for "alias." with the regex
+                        //To avoid fetching multiple time the same data because it's used in multiple column
+                        if (!aliasesAlreadyFound.includes(alias)) {
+                            aliasesAlreadyFound.push(alias);
```
TheMaskedTurtle, line 190: Why do you introduce all this complexity ? Why not just fetching all the data related to all the nodes defined in the customColumnsNodesAliases ?

src/components/spreadsheet/use-spreadsheet-equipments.ts
```diff
@@ -144,7 +159,80 @@ export const useSpreadsheetEquipments = (
...
+                                    }
+                                });
+                            }
+                        }
+                    });
+                });
+
+                //so we only dispatch once all the fetches are over
+                Promise.all(fetchers).then(() => {
+                    dispatch(addAdditionalEquipmentsByNodesForCustomColumns(additionalEquipmentsByNodes));
```
TheMaskedTurtle, line 220: Why do you create additional data in the redux for this ? Why not directly in the spreadsheetNetwork ?

src/components/spreadsheet/table-wrapper.tsx
```diff
@@ -485,17 +495,55 @@ const TableWrapper: FunctionComponent<TableWrapperProps> = ({
...
+        },
+        [tablesDefinitions]
+    );
+
     useEffect(() => {
         if (disabled || !equipments) {
             return;
         }
+
+        const equipmentType = toSpreadsheetEquipmentType(tablesNames[tabIndex]);
```
TheMaskedTurtle, line 511: This additionnal treatment and use of new selectors won't be necessary if you use spreadsheetNetwork directly

TheMaskedTurtle (changes requested):

src/components/spreadsheet/use-spreadsheet-equipments.ts
AbdelHedhili: but that's precisely what i'm trying to do here or i'm not understanding what you're saying ? I can't just fetch the data for every customcolumn without keeping trace of what i already fetch or i would be doing 50 time the same fetch

AbdelHedhili: Oh ok i got it wrong, I understand what you mean, but it's not great to do that, we would be fetching a lot of data we don't actually need if we fetch all the data of the nodes that are found in aliases

TheMaskedTurtle: And I'm not a big fan of parsing formulas like that I think for now we should avoid this.

TheMaskedTurtle: And for me you only have one fetch to do when you add a new node in the node aliases list, this is absolutely not related to custom columns or even "traditional" columns. It's just that you want to load additionnal data

src/components/spreadsheet/use-spreadsheet-equipments.ts
AbdelHedhili: I thought it's better to have the other nodes info in a seperate state because its probable that it won't behave exactly as the current data info

src/components/spreadsheet/use-spreadsheet-equipments.ts
```diff
@@ -126,6 +133,14 @@ export const useSpreadsheetEquipments = (
         resetImpactedElementTypes,
     ]);
 
+    const toSpreadsheetEquipmentType = useCallback(
+        (tableName: string): SpreadsheetEquipmentType | undefined => {
+            return tablesDefinitions.find((spreadsheetTabDefinition) => spreadsheetTabDefinition.name === tableName)
+                ?.type;
+        },
+        [tablesDefinitions]
+    );
```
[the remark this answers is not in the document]
AbdelHedhili: Yep, I missed it. Thanks

src/components/spreadsheet/use-spreadsheet-equipments.ts
AbdelHedhili: I don't know... So the user add 2 customs columns one "alias1.maxP" and "alias2.maxP" only for the generators and we would end up with 3 whole network loaded in the front. That seems so unreasonable, that's why i did it this way.

AbdelHedhili: > And for me you only have one fetch to do when you add a new node in the node aliases list, this is absolutely not related to custom columns or even "traditional" columns. It's just that you want to load additionnal data

But it is related, this node alias option is only here to define the aliases used in... the custom columns. It was not defined as a load the network with this id. It's a this alias = this node in the custom column

TheMaskedTurtle: For now, but in a really near future all columns will be "custom" columns, then it won't make any difference. And do we agree that the fetcher is loading all data for one type of equipment anyway ? It doesn't load just the data you need in the column ? So what you did just postpone the moment we're fetching the related equipments data of the new node (when you use it in a column) but it does not make a great difference if you fetch them node/alias declaration directly.

src/components/spreadsheet/use-spreadsheet-equipments.ts
TheMaskedTurtle: I think that's the whole point, it should behave exactly the same. So that, we just load more data initially and then nothing else changes in the code.

src/components/spreadsheet/use-spreadsheet-equipments.ts
TheMaskedTurtle: But yes we have to think about this conception considering what will be the final usage of the spreadsheet. All columns will be customizable, and we will be able to load several nodes of the network for one given equipment type. So we have to think about it like 2 distinct systems, the node/alias system enables to fill more available data in our spreadhseet, and the custom column system that enables the users to combine nodes data as they want.

src/components/spreadsheet/custom-columns/custom-columns-nodes-dialog.tsx
```diff
@@ -0,0 +1,111 @@
...
+};
+
+export default function CustomColumnNodesDialog({ open }: Readonly<CustomColumnNodesDialogProps>) {
+    const dispatch = useDispatch<AppDispatch>();
+
+    const intl = useIntl();
+
+    const customColumnsNodesAliases = useSelector((state: AppState) => state.customColumnsNodesAliases);
+
+    const formMethods = useForm({
```
TheMaskedTurtle, line 48:
```suggestion
    const formMethods = useForm<CustomColumnNodesForm>({
```

src/components/spreadsheet/custom-columns/custom-columns-nodes-dialog.tsx
```diff
@@ -0,0 +1,111 @@
...
+    });
+    const { reset, handleSubmit } = formMethods;
+
+    const onValidate = (data: any) => {
+        onClose();
+        const nodesAliases: NodeAlias[] = data.nodesAliases.map((nodeAlias: NodeAlias) => {
+            return { name: nodeAlias.name, alias: nodeAlias.alias };
+        });
+        dispatch(updateCustomColumnsNodesAliases(nodesAliases));
+    };
```
TheMaskedTurtle, line 60:
```suggestion
const onValidate = (data: CustomColumnNodesForm) => {
        onClose();
        dispatch(updateCustomColumnsNodesAliases(data.nodesAliases));
    };
```

src/components/spreadsheet/custom-columns/custom-columns-nodes-dialog.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    const clear = useCallback(() => {
+        reset(initialCustomColumnNodesForm);
+    }, [reset]);
+
+    const onClose = () => {
+        open.setFalse();
+        clear();
+    };
```
TheMaskedTurtle, line 69:
```suggestion
const onClose = () => {
        open.setFalse();
        reset(initialCustomColumnNodesForm);
    };
```

src/components/spreadsheet/table-wrapper.tsx
```diff
@@ -156,6 +157,10 @@ interface TableWrapperProps {
     disabled: boolean;
 }
 
+interface RecursiveIdentifiable extends Identifiable {
+    [alias: string]: Identifiable | unknown;
```
TheMaskedTurtle, line 161:
```suggestion
    [alias: string]: Identifiable | string;
```

src/components/spreadsheet/use-spreadsheet-equipments.ts
```diff
@@ -144,7 +150,79 @@ export const useSpreadsheetEquipments = (
...
+        customColumnsDefinitions,
+    ]);
+
+    useEffect(() => {
+        if (studyUuid && equipments != null && Object.keys(customColumnsDefinitions).length > 0) {
+            const aliases = customColumnsNodesAliases.map((nodeAlias) => nodeAlias.alias);
+            const aliasesAlreadyFound: string[] = [];
+            const currentTabCustomColumnsDef = customColumnsDefinitions[tablesNames[tabIndex]];
+            const customColumns = currentTabCustomColumnsDef.columns;
+            let fetchers: Promise<any>[] = [];
```
TheMaskedTurtle, line 169:
```suggestion
            let fetchers: Promise<unknown>[] = [];
```

TheMaskedTurtle (changes requested):

src/components/spreadsheet/use-spreadsheet-equipments.ts
TheMaskedTurtle: ok we'll wee that in another ticket

src/components/spreadsheet/table-wrapper.tsx
TheMaskedTurtle: Yes, we'll see that in another ticket

src/components/spreadsheet/use-spreadsheet-equipments.ts
```diff
@@ -144,7 +148,63 @@ export const useSpreadsheetEquipments = (
...
+    useEffect(() => {
+        if (studyUuid && equipments != null) {
+            const aliases = customColumnsNodesAliases.map((nodeAlias) => nodeAlias.alias);
+            let fetchers: Promise<unknown>[] = [];
+            let additionalEquipmentsByNodes: Record<string, Record<SpreadsheetEquipmentType, Identifiable[]>> = {};
+            if (equipment.type) {
+                aliases.forEach((alias) => {
+                    const nodeNameAssociatedToAlias = customColumnsNodesAliases.find(
+                        (nodeAlias) => nodeAlias.alias === alias
+                    )?.name;
```
TheMaskedTurtle, line 170: directly iterate through customColumnsNodesAliases to get the names ?

TheMaskedTurtle: Can't we store the node UUIDs directly and the names would be only used for displaying purposes in the autocomplete ? But behind the scene we only UUIDs ? I think it would be a bit more robust

TheMaskedTurtle (approved):

Meklo (approved):

PR #2476, opened 2025-01-02
Rendering optimization when we reorder nodes

SlimaneAmar:


src/components/network-modification-tree-pane.jsx
```diff
@@ -185,6 +189,24 @@ export const NetworkModificationTreePane = ({ studyUuid, studyMapTreeDisplay })
...
+        (parentNodeId, orderedChildrenNodeIds) => {
+            // We check if the current position is already correct
+            const children = treeModelRef.current.getChildren(parentNodeId);
+            if (orderedChildrenNodeIds.length !== children.length) {
+                snackWarning({
+                    messageId: 'ReorderSubtreeInvalidNotifInfo',
+                });
+                console.warn('Subtree order update cancelled : the size of the ordered children list is incompatible');
+                return;
+            }
```
EstherDarkish, line 202: This code is missing the update with your new needReorder function.

SlimaneAmar: Yes it'only from a notification !

src/components/network-modification-tree.jsx
```diff
@@ -80,20 +80,17 @@ const NetworkModificationTree = ({
...
-                window.requestAnimationFrame(() => fitView());
-            }
-        },
-        [fitView, setNodes, setEdges]
-    );
+    const updateNodePositions = useCallback(() => {
+        if (treeModel && treeModel.treeNodes?.length > 0) {
+            setNodes(getTreeNodesWithUpdatedPositions(treeModel.treeNodes));
+            setEdges([...treeModel.treeEdges]);
+            window.requestAnimationFrame(() => fitView());
```
EstherDarkish, line 87: This line should be removed as it changes the screen position when interacting with nodes.

EstherDarkish (approved):
test ok code ok

PR #2480, opened 2025-01-07
LogTable : remove direct report dependency and include severity fetching logic

ayolab:


TheMaskedTurtle (changes requested):
As discussed maybe move the severities fetching in report-viewer-tab.jsx and computation-report-viewer.tsx because we need to fetch that only once at initial loading.

src/components/results/common/computation-report-viewer.tsx
```diff
@@ -35,19 +37,27 @@ export const ComputationReportViewer: FunctionComponent<ComputationReportViewerP
             fetchReport().then((report) => {
                 if (report !== undefined) {
                     setReport(report);
+                    fetchReportSeverities?.(report?.id, report?.parentId ? ReportType.NODE : ReportType.GLOBAL)?.then(
```
TheMaskedTurtle, line 40: For computation it is always NODE no ?

TheMaskedTurtle (approved):

PR #2498, opened 2025-01-15
Cleans up report-viewer.tsx

Meklo:


src/components/report-viewer/report-viewer.tsx
```diff
@@ -8,104 +8,107 @@ import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
...
     const reportTree = useMemo(() => mapReportsTree(report), [report]);
     const reportTreeMap = useRef<Record<string, ReportTree>>({});
 
-    const mapReportsById = useCallback((item: ReportTree) => {
-        reportTreeMap.current[item.id] = item;
-        for (let subReport of item.subReports) {
-            mapReportsById(subReport);
-        }
-    }, []);
+    const computeReportTreeMap = useCallback(() => {
```
TheMaskedTurtle, line 35: Maybe just deifne reportTreeMap as a useMemo ? Depending only on reportTree ?

Meklo: No problem, I did so initially but reverted to keep callbacks dependency arrays empty.
In the end it recomputes only when the whole tree changes so it's no big deal

TheMaskedTurtle (approved):

jonenst (approved):

jonenst:
care with compound object as state, it doesn't have the behavior of primitives or string to compare by value and not rerender if not modified

jonenst:
useMemo for string concatenation doesn't do much but doesn't hurt

PR #2513, opened 2025-01-22
Fix missing twowindingstransformer distant regulation terminal field

jonenst:
regression from forgotten migration in 3e3ba6


ghazwarhili (approved):
Test => OK

PR #2516, opened 2025-01-22
Add ability to filter spreadsheet data from GS filters

Meklo:


src/components/spreadsheet/spreadsheet-gs-filter.tsx
```diff
@@ -0,0 +1,103 @@
...
+                        titleId="FiltersListsSelection"
+                        label="filter"
+                        elementType={ElementType.FILTER}
+                        equipmentTypes={[equipmentType]}
+                        labelRequiredFromContext={false}
+                    />
+                </CustomFormProvider>
+            </Popover>
+        </>
+    );
```
TheMaskedTurtle, line 102: Maybe we should only have the DirectoryItemsInput no ? I think it's a bit weird to have this multiselect input opened after a click on a button

src/components/spreadsheet/spreadsheet-gs-filter.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { spreadsheetStyles } from './utils/style';
+
+interface SpreadsheetGsFilterProps {
+    equipmentType: SpreadsheetEquipmentType;
+    applyGsFilter: (filter: ExpertFilter[]) => void;
+}
+
+export const SpreadsheetGsFilter = ({ equipmentType, applyGsFilter }: SpreadsheetGsFilterProps) => {
+    const dispatch = useDispatch();
+    const gsFilterSpreadsheetState = useSelector((state: AppState) => state.gsFilterSpreadsheetState);
```
TheMaskedTurtle, line 37: Maybe also move this in the hook ?

src/components/spreadsheet/utils/spreadsheet-gs-filter-utils.ts
```diff
@@ -0,0 +1,47 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import { ExpertFilter } from '../../../services/study/filter';
+import yup from '../../utils/yup-config';
+import { FILTERS, ID, NAME, SPREADSHEET_GS_FILTER } from '../../utils/field-constants';
+
+export type ExpertFilterForm = Omit<ExpertFilter, 'type' | 'equipmentType' | 'topologyKind' | 'rules'>;
+
+export const spreadsheetGsFilterFormSchema = yup.object({
```
TheMaskedTurtle, line 14: That's a bit too much to define all this... Isn't it possible to have an item component that just returns a list of IDs and that we don't have too much boilerplate code to write to use it ?

Meklo: Ideally we should have a `DirectoryItemsInput` which isn't coupled with RHF, I haven't add it because I judged it out of scope for this issue but that would be a good time to implement it

src/components/spreadsheet/spreadsheet-gs-filter.tsx
Meklo: I thought it would be cleaner and more homogeneous to add a button instead of a field directly in the spreadsheet top bar but I agree a popover only containing one field might be out of place in the application. I could check with Christine

TheMaskedTurtle: Yeah maybe you can see that with Christine :)

Meklo: She agreed with you I'll roll this back

src/components/spreadsheet/spreadsheet-gs-filter.tsx
Meklo: I moved the filter application logic inside the hook but had to keep the reference here to be able to reset the input values when switching equipment

TheMaskedTurtle (approved):

PR #2517, opened 2025-01-22
Fix: correctly update NAD when initNadWithGeoData param change

sBouzols:
fix(DiagramPane): call updateNAD when initNadWithGeoData changed


EstherDarkish (approved):

souissimai (approved):
it seems ok

PR #2521, opened 2025-01-24
Fix NAD double fetch from a certain depth

sBouzols:
fix(DiagramPane): use `previousNetworkAreaDiagramDepth` in test and Move its setting a little later in the code to avoid some close in time calls to `addOrReplaceNAD` which proc NAD fetch.


EstherDarkish (approved):
Code OK test OK

EstherDarkish:
Wait, I still can reproduce the bug in my local environment...

souissimai (approved):

PR #2531, opened 2025-01-28
violet voltage level chip

Mathieu-Deharbe:
various UI changes around computing results filters

### DIFF
--- a/src/components/results/common/results-global-filter.tsx
+++ b/src/components/results/common/results-global-filter.tsx
@@ -82,7 +82,7 @@ const styles = {
     },
     chipVoltageLevel: {
         '&.MuiChip-root, &.MuiChip-root[aria-selected="true"]': {
-            backgroundColor: '#1e88e5 !important',
+            backgroundColor: '#9c27b0 !important',
         },
         '& .MuiChip-deleteIcon': {
             color: '#bbdefb',


src/components/results/common/results-global-filter.tsx
```diff
@@ -224,22 +240,40 @@ const ResultsGlobalFilter: FunctionComponent<ResultsGlobalFilterProps> = ({ onCh
...
-                // Allows to find the translated countries (and not their countryCodes) when the user inputs a search value
-                filterOptions={(options: Filter[], state: FilterOptionsState<Filter>) =>
-                    options.filter((option) => {
-                        const labelToMatch =
-                            option.filterType === FilterType.COUNTRY ? translate(option.label) : option.label;
-                        return labelToMatch.toLowerCase().includes(state.inputValue.toLowerCase());
-                    })
-                }
+                filterOptions={(options: Filter[], state: FilterOptionsState<Filter>) => {
+                    let numByGroup: Map<string, number> = new Map();
```
basseche, line 252: use const instead of let

src/components/results/common/results-global-filter.tsx
```diff
@@ -224,22 +240,40 @@ const ResultsGlobalFilter: FunctionComponent<ResultsGlobalFilterProps> = ({ onCh
...
+                    let filteredOptions: Filter[] = options
```
basseche, line 253: same

src/components/results/common/results-global-filter.tsx
```diff
@@ -113,24 +115,29 @@ export interface ResultsGlobalFilterProps {
 }
 
 const emptyArray: Filter[] = [];
+const DEFAULT_NB_OPTIONS_DISPLAYED: number = 10;
```
basseche, line 118: is this constant going to be used somewhere else ? If you think so maybe it would be a good idea to put it inside utils/constants.ts

src/translations/en.json
```diff
@@ -493,8 +493,10 @@
     "LoadFlowResultsStatus": "Status",
     "ComputationResultsLogs": "Logs",
 
-    "results.globalFilter.fillerText": "Country or voltage level",
-    "results.globalFilter.recent": "Recent:",
+    "results.globalFilter.fillerText": "Filtering",
+    "results.globalFilter.recent": "Recents",
```
basseche, line 497: are you sûr that Recents should take an 's' ?

src/components/results/common/results-global-filter.tsx
```diff
@@ -41,66 +41,68 @@ const styles = {
...
         paddingLeft: 1,
     }),
     chip: {
         '&.MuiChip-root': {
+            borderRadius: '100px solid',
             margin: '4px 2px 4px 2px',
             padding: '0',
             color: 'white',
         },
+        '.MuiChip-deleteIcon': {
```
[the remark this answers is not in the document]
Mathieu-Deharbe: It doesn't work.
Probably because the `deleteIcon` is inside another anchor in another part of the very complicated hierachy. I don't see how to get the right adress. And I don't think it brings much given that this will only apply to the chip using that style anyway, which is fine.

src/components/results/common/results-global-filter.tsx
```diff
@@ -41,66 +41,68 @@ const styles = {
...
+            color: 'white',
+            opacity: 0.6,
+        },
         '.MuiChip-deleteIcon:hover': {
```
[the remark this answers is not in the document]
Mathieu-Deharbe: Same.

src/components/results/common/results-global-filter.tsx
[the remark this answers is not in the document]
basseche: ok

basseche (approved):

ghazwarhili:
Capture d’écran du 2025-02-17 13-32-22
to be fixed

ghazwarhili (approved):

PR #2534, opened 2025-01-28
Fix undisplayed error logs on root node

FranckLecuyer:


src/components/report-viewer/log-table.tsx
```diff
@@ -120,7 +120,7 @@ const LogTable = ({ selectedReport, reportType, severities, onRowClick }: LogTab
     }, [severityFilter, fetchReportLogs, selectedReport, messageFilter, resetSearch]);
 
     useEffect(() => {
-        if (filters?.length === 0 && severities && severities.length > 0) {
```
Meklo, line 123: Removing the condition altogether has some side effects in other use cases. it resets the displayed severities selection when switching selected nodes from the treeview

Meklo (changes requested):

src/components/report-viewer/log-table.tsx
FranckLecuyer: :+1: 
Fixed

Meklo (approved):

PR #2541, opened 2025-01-30
fix: dependencies form

TheMaskedTurtle:
- translation id mal-configured
- disable user from entering custom value
- not allow to define the column itself in its dependencies


src/components/spreadsheet/custom-columns/custom-columns-dialog.tsx
```diff
@@ -217,11 +217,16 @@ export default function CustomColumnDialog({
...
                                 name={COLUMN_DEPENDENCIES}
-                                options={customColumnsDefinitions?.map((definition) => definition.id) ?? []}
+                                options={
+                                    customColumnsDefinitions
+                                        ?.map((definition) => definition.id)
+                                        .filter((id) => id !== columnId) ?? []
+                                }
                                 disableClearable={false}
                                 disableCloseOnSelect
+                                allowNewValue={false}
```
achour94, line 229: This still does not fix the issue of adding random dependencies, the problem is in the handleOnBlur function in MultipleAutocompleteInput component, you have to add a boolean prop for exemple , to avoid adding the unsavedInut to the form data 
```
const handleOnBlur = () => {
        if (unsavedInput && !watchAutocompleteValues.includes(unsavedInput)) {
            append(unsavedInput);
        }
        setUnsavedInput('');
    };
```

achour94 (changes requested):

PR #2560, opened 2025-02-04
fix: simplify column state management and reduce quantity of custom states

TheMaskedTurtle:
…tates


src/components/spreadsheet/custom-columns/use-custom-column.ts
```diff
@@ -38,32 +37,34 @@ export function useCustomColumn(tabIndex: number) {
...
-                    sortParams: {
-                        table: SPREADSHEET_SORT_STORE,
-                        tab: tablesDefinitionIndexes.get(tabIndex)!.name,
-                    },
-                    menu: {
-                        Menu: CustomColumnMenu,
-                        menuParams: {
-                            tabIndex,
-                            customColumnName: colWithFormula.name,
+    const customColumns = useMemo(
```
achour94, line 40: Forgot to import useMemo

achour94 (changes requested):

src/components/spreadsheet/custom-columns/use-custom-column.ts
TheMaskedTurtle: fixed

src/components/custom-aggrid/custom-column-menu.tsx
```diff
@@ -60,9 +59,14 @@ export const CustomColumnMenu: React.FC<CustomColumnConfigProps> = ({
     const handleValidate = useCallback(() => {
         if (customColumnsDefinition?.id) {
```
achour94, line 60: currentColumnDefinition could be a useMemo, we don't really need a state

src/components/spreadsheet/equipment-tabs.tsx
```diff
@@ -22,22 +21,8 @@ interface EquipmentTabsProps {
...
     const developerMode = useSelector((state: AppState) => state[PARAM_DEVELOPER_MODE]);
-
-    const existsInTable = function (name: string) {
-        return TABLES_NAMES.findIndex((n) => n === name) !== -1;
-    };
-
-    const tablesNames = useMemo(() => {
-        return allTablesNames.map((tabName) => {
-            return existsInTable(tabName)
-                ? intl.formatMessage({
```
achour94, line 35: By removing this, we have an Error MISSING_TRANSLATION when we create a custom table

src/components/spreadsheet/custom-spreadsheet/custom-spreadsheet-dialog.tsx
```diff
@@ -97,7 +96,7 @@ export default function CustomSpreadsheetConfigDialog({
             if (selectedOption?.id === NEW_SPREADSHEET_CREATION_OPTIONS.EMPTY.id) {
```
achour94, line 96: You should update the columnsStates in redux when we submit , otherwise we have a bug when we open an imported/created new spreadsheet

achour94 (changes requested):
We have some bugs :)

src/components/spreadsheet/equipment-tabs.tsx
TheMaskedTurtle: I remove the translation because in the future tab names will be all customs too so the application cannot translate the labels dynamically

src/components/custom-aggrid/custom-column-menu.tsx
TheMaskedTurtle: done

src/components/spreadsheet/custom-spreadsheet/custom-spreadsheet-dialog.tsx
TheMaskedTurtle: fixed

achour94 (changes requested):
Bug: the backend request to save a custom table having a custom column fails

src/components/spreadsheet/table-wrapper.tsx
```diff
@@ -352,79 +294,26 @@ export const TableWrapper: FunctionComponent<TableWrapperProps> = ({
         manualTabSwitch,
         tabIndex,
         scrollToEquipmentIndex,
-        tablesDefinitionTypes,
+        tablesDefinitions,
     ]);
 
-    const handleRowDataUpdated = useCallback(() => {
```
achour94, line 358: We lose scrolling functionality in this scenario: If a spreadsheet has never been rendered and we attempt to visualize a line (for example) from the map in the spreadsheet, the scrolling does not work

achour94 (changes requested):

src/components/spreadsheet/equipment-tabs.tsx
```diff
@@ -55,9 +38,11 @@ export const EquipmentTabs: FunctionComponent<EquipmentTabsProps> = ({ tabIndex,
                     }}
                     aria-label="tables"
                 >
-                    {tablesNames.map((tabName) => (
-                        <Tab key={tabName} label={tabName} disabled={disabled} />
-                    ))}
+                    {tablesDefinitions
```
achour94, line 41: could be simplified with one map 
{tablesDefinitions.map((def) => (
                        <Tab key={def.name} label={def.name} disabled={disabled} />
                    ))}

src/components/spreadsheet/table-wrapper.tsx
TheMaskedTurtle: you're right, will revert

achour94 (approved):

PR #2562, opened 2025-02-04
fix: avoid refetch when data is already fetched

TheMaskedTurtle:


src/components/spreadsheet/use-spreadsheet-equipments.ts
```diff
@@ -210,6 +194,7 @@ export const useSpreadsheetEquipments = (
         treeModel,
```
achour94, line 194: do we need treeModel and equipments in the dependency list ?

AbdelHedhili (approved):

PR #2564, opened 2025-02-05
Add selection type for equipments selection in voltage init parameters

FranckLecuyer:


src/components/dialogs/parameters/voltageinit/equipment-selection-parameters.tsx
```diff
@@ -5,28 +5,87 @@
...
+        },
+    };
+};
+
 const EquipmentSelectionParameters = () => {
+    const initialEquipmentsSelectionType = useMemo(() => {
+        return intlEquipmentsSelectionType();
+    }, []);
+
     return (
```
ayolab, line 55:
```suggestion
const initialEquipmentsSelectionType = {
    ALL_EXCEPT: {
        id: 'ALL_EXCEPT',
        label: 'allExcept',
    },
    NONE_EXCEPT: {
        id: 'NONE_EXCEPT',
        label: 'noneExcept',
    },
};

const EquipmentSelectionParameters = () => {
    return (
```
there is no need for the useMemo here

ayolab (changes requested):

ayolab (approved):

PR #2570, opened 2025-02-06
Keep zoom pan on NAD and SLD when changing views

sBouzols:
feat(DiagramPane): set doNotBailOutOnEmptyChildren on Autosizer to avoid to unmount children component when Map view and diagrams are hidden.


EstherDarkish (approved):
Test OK Code OK

PR #2574, opened 2025-02-10
refacto spreadsheet redux storage

AbdelHedhili:


TheMaskedTurtle (commented):
Tests ok, it functionnaly works except the small bug of initial sorting :) 
And the redux state looks much better. I still feel like the id matching to build final data in table-wrapper is too eager but we can see that later.

TheMaskedTurtle (approved):

PR #2575, opened 2025-02-10
use all columns as custom

achour94:


src/redux/reducer.ts
```diff
@@ -1737,24 +1733,44 @@ export const reducer = createReducer(initialState, (builder) => {
...
+    builder.addCase(UPDATE_COLUMNS_DEFINITION, (state, action: UpdateColumnsDefinitionsAction) => {
+        const { colData } = action;
+
+        // Retrieve the table definition by index
+        const tableDefinitionIndex = state.tables.definitions.findIndex((def) => def.index === colData.index);
+
+        if (tableDefinitionIndex !== -1) {
+            const existingColumnIndex = state.tables.definitions[tableDefinitionIndex].columns.findIndex(
+                (col) => col.id === colData.value.id
+            );
```
TheMaskedTurtle, line 1745:
```suggestion
        const tableDefinition = state.tables.definitions[colData.index];
        if (tableDefinition) {
            const existingColumnIndex = tableDefinition.columns.findIndex((col) => col.id === colData.value.id);
```
Why not this directly ?

src/redux/reducer.ts
```diff
@@ -1737,24 +1733,44 @@ export const reducer = createReducer(initialState, (builder) => {
...
+
+        if (tableDefinitionIndex !== -1) {
+            state.tables.definitions[tableDefinitionIndex].columns = state.tables.definitions[
+                tableDefinitionIndex
+            ].columns.filter((col) => col.id !== value);
+            // remove column from columnsStates
+            state.tables.columnsStates[tableDefinitionIndex] = state.tables.columnsStates[tableDefinitionIndex].filter(
+                (col) => col.colId !== value
+            );
+        }
```
TheMaskedTurtle, line 1773:
```suggestion
    builder.addCase(REMOVE_COLUMN_DEFINITION, (state, action: RemoveColumnDefinitionAction) => {
        const { index, value } = action.definition;
        const tableDefinition = state.tables.definitions[index];

        if (tableDefinition) {
            tableDefinition.columns = state.tables.definitions[index].columns.filter((col) => col.id !== value);
            // remove column from columnsStates
            state.tables.columnsStates[index] = state.tables.columnsStates[index].filter((col) => col.colId !== value);
        }
    });
```
No ?

TheMaskedTurtle (changes requested):
You should harmonize id/name across equipment types. For example some are id:ID and name:ID and some others are id:id and name:ID. I think we should do everywhere ids in lower case and names with upper case, but be careful then in the initial sort in the reducer you should put id and not ID. Currently some tabs are broken because of that.

TheMaskedTurtle (approved):

PR #2579, opened 2025-02-10
Block params tab when computation is in progress

ayolab:


src/components/parameters-tabs.tsx
```diff
@@ -258,7 +247,7 @@ const ParametersTabs: FunctionComponent<OwnProps> = (props) => {
         });
     }, [enableDeveloperMode]);
 
-    const loadFlowStatus = useSelector((state: AppState) => state.computingStatus[ComputingType.LOAD_FLOW]);
+    const computationStatus = useSelector((state: AppState) => state.computingStatus[tabValue as ComputingType]);
```
FranckLecuyer, line 250: for SHORT_CIRCUIT_ONE_BUS computation, this does not work, because here, we have 2 distinct computation type (SHORT_CIRCUIT and SHORT_CIRCUIT_ONE_BUS) for just one parameter tab value (TAB_VALUES.shortCircuitParamsTabValue)

FranckLecuyer (changes requested):
Code Review: OK
Tests: pb with SHORT_CIRCUIT_ONE_BUS computation, where we can still modify the parameters during computation ...

FranckLecuyer (approved):

PR #2592, opened 2025-02-17
Fix get short circuit result for one bus

SlimaneAmar:


antoinebhs (changes requested):
There is an error if we switch between a root network + node with SC results for one bus and a root network node unbuilt
The error is: "We should have one and only one fault for one bus SCA results, or we found 0 or more than 1"
And there is the snackbar: "An error occurred while retrieving short circuit analysis result"

SlimaneAmar:
> There is an error if we switch between a root network + node with SC results for one bus and a root network node unbuilt The error is: "We should have one and only one fault for one bus SCA results, or we found 0 or more than 1" And there is the snackbar: "An error occurred while retrieving short circuit analysis result"

This bug is already present
I create a ticket

antoinebhs (approved):

PR #2599, opened 2025-02-18
Add save spreadsheet collection

achour94:


src/components/spreadsheet/custom-spreadsheet/spreadsheet-collection-save-dialog.tsx
```diff
@@ -0,0 +1,248 @@
...
+    type: SpreadsheetEquipmentType;
+    selected: boolean;
+    index: number;
+}
+
+export const SpreadsheetCollectionSaveDialog: FunctionComponent<SpreadsheetCollectionSaveDialogProps> = ({ open }) => {
+    const { snackError, snackInfo } = useSnackMessage();
+    const intl = useIntl();
+    const tables = useSelector((state: AppState) => state.tables.definitions);
+    const columnsStates = useSelector((state: AppState) => state.tables.columnsStates);
```
dbraquart, line 54: I dont get why columnsStates is used in this component (see below)

src/components/spreadsheet/custom-spreadsheet/spreadsheet-collection-save-dialog.tsx
```diff
@@ -0,0 +1,248 @@
...
+            });
+        }
+    }, []);
+
+    const getReorderedColumns = useCallback(
+        (tableIndex: number): ColumnDefinitionDto[] => {
+            const table = tables[tableIndex];
+            const columnStates = columnsStates[tableIndex];
+
+            return columnStates
```
dbraquart, line 113: why not directly use:

_return table.columns.map((column) => {
                const dto: ColumnDefinitionDto = {
                    uuid: column.uuid ?? uuid4(),
etc etc_

src/components/spreadsheet/custom-spreadsheet/spreadsheet-collection-save-dialog.tsx
```diff
@@ -0,0 +1,248 @@
...
+            setLocalTablesState((prev) => {
+                const reordered = [...prev];
+                const [removed] = reordered.splice(source.index, 1);
+                reordered.splice(destination.index, 0, removed);
+                return reordered;
+            });
+        }
+    }, []);
+
+    const getReorderedColumns = useCallback(
```
dbraquart, line 108:
```suggestion
    const getTableColumns = useCallback(
```
Why call it "reordered" ? We return the natural columns order ...

dbraquart (changes requested):
code: a few remarks to be discussed
tests: ok, but:

I think we shoud not allow saving an empty collection element. To be discuss with the PO ?

src/components/spreadsheet/custom-spreadsheet/spreadsheet-collection-save-dialog.tsx
achour94: because the order of the columns is stored in culomns state, and we want to get the order of the columns as we set it in the app

dbraquart: currently we cannot change column order, so I did not see the difference.

dbraquart (approved):

src/components/spreadsheet/custom-spreadsheet/spreadsheet-collection-save-dialog.tsx
achour94: we can change it with the button 'Show/hide columns'

PR #2605, opened 2025-02-20
[BUG] Fix bug post reindex-all when a study not yet indexed

thangqp:
Fix regression from [2457](https://github.com/gridsuite/gridstudy-app/pull/2457)

Should not use previous, current value at the same time in dependencies


src/components/study-container.jsx
```diff
@@ -677,10 +675,7 @@ export function StudyContainer({ view, onChangeTab }) {
     );
 
     useEffect(() => {
-        if (
-            (studyUuid && currentRootNetworkUuid && !isStudyNetworkFound) ||
-            (currentRootNetworkRef.current && currentRootNetworkRef.current !== currentRootNetworkUuid)
-        ) {
```
sBouzols, line 683: this is an error.
If `currentRootNetworkUuid` change, you must check network existence, check study indexation for this root network etc... then you must keep this

src/components/study-container.jsx
```diff
@@ -729,22 +724,18 @@ export function StudyContainer({ view, onChangeTab }) {
         }
         // A modification has been added to the currentNode and this one has been built incrementally.
         // No need to load the network because reloadImpactedSubstationsEquipments will be executed in the notification useEffect.
-        if (
-            previousCurrentRootNetwork === currentRootNetworkUuid &&
-            isSameNode(previousCurrentNode, currentNode) &&
-            isNodeBuilt(previousCurrentNode)
-        ) {
```
sBouzols, line 736: This is an error,
If currentRootNetwork changes then we must reset equipments in spreadsheets.

src/components/study-container.jsx
```diff
@@ -537,7 +535,7 @@ export function StudyContainer({ view, onChangeTab }) {
                         });
 
                     // Check if the current root network has changed and if there's a current selected node
-                    if (previousCurrentRootNetwork !== currentRootNetworkUuid && currentNode) {
+                    if (currentNode) {
```
sBouzols, line 538: this is an error.
We must override the currentNode when rootNetwork change because even if the tree is the same currentNode could have different build status in a root network or another.

src/components/study-container.jsx
thangqp: OK! I reverted condition but using previous value in ref as in other useEffect

src/components/study-container.jsx
thangqp: OK! I reverted while keeping condition with previous value in ref

src/components/study-container.jsx
thangqp: I keep the condition as before but using previous value in ref.. but i think it not really necessary because in the useEffect we have already a condition to trigger the call of loadTree

PR #2607, opened 2025-02-20
Fix StudyContainer double call to `checkNetworkExistenceAndRecreateIfNotFound` at start

sBouzols:
Remove error prone `previousCurrentRootNetwork` which already exist with the `currentRootNetworkRef`
Adapt break conditions in the related `useEffect`

Use `setIsStudyNetworkFound` consistently to apply waitingLoader at rootNetwork check existency, (when changing root networks for example).


src/components/study-container.jsx
```diff
@@ -718,7 +718,7 @@ export function StudyContainer({ view, onChangeTab }) {
         }
         let previousCurrentNode = currentNodeRef.current;
         currentNodeRef.current = currentNode;
-
+        let previousCurrentRootNetwork = currentRootNetworkRef.current;
```
thangqp, line 721:
```suggestion
        const previousCurrentRootNetwork = currentRootNetworkRef.current;
```

src/components/study-container.jsx
```diff
@@ -242,6 +242,8 @@ export function StudyContainer({ view, onChangeTab }) {
     const [errorMessage, setErrorMessage] = useState(undefined);
 
     const [isStudyNetworkFound, setIsStudyNetworkFound] = useState(false);
+    const isStudyNetworkFoundRef = useRef(false);
+    isStudyNetworkFoundRef.current = isStudyNetworkFound;
```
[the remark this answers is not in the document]
sBouzols: it's to avoid using `isStudyNetworkFound` state as a trigger dependency in the corresponding `useEffect`, classic workaround.

src/components/study-container.jsx
sBouzols: true, will fix it in the following refactor PR because it's also true for `previousCurrentNode` variable here

thangqp (approved):
Test OK!
Code to validate by your teams

thangqp (approved):
Tested again, OK 

Just see a warning when switch to second root network:

Serverside warning: Something went wrong, Study equipments aren't indexed anymore. Please refresh the page (F5)

souissimai (approved):
Tested, Code OK

PR #2612, opened 2025-02-21
Add manual node refresh in spreadsheet

dbraquart:


achour94 (approved):

src/components/spreadsheet/data-fetching/use-load-equipment.ts
```diff
@@ -0,0 +1,118 @@
...
+import { isStatusBuilt } from '../../graph/util/model-functions';
+import { AppState } from '../../../redux/reducer';
+import { loadEquipments } from '../../../redux/actions';
+import { getFetcher } from './fetchers';
+
+export const useLoadEquipment = (
+    type: SpreadsheetEquipmentType,
+    studyUuid: UUID | null,
+    currentRootNetworkUuid: UUID | null,
+    currentNodeUuid: UUID | undefined,
```
TheMaskedTurtle, line 23: This shouldn't be passed, it should use useSelector in the hook directly

src/translations/spreadsheet-en.ts
```diff
@@ -32,6 +32,10 @@ const spreadsheetEn = {
     'spreadsheet/custom_column/column_name_already_exist': 'Column name already exists',
     'spreadsheet/custom_column/dependencies': 'The formula is linked to other columns',
     'spreadsheet/custom_column/nodes': 'Nodes',
+    'spreadsheet/custom_column/option/parameter': 'Parameter',
```
TheMaskedTurtle, line 35: nit: configure ?

src/components/spreadsheet/data-fetching/use-spreadsheet-equipments.ts
```diff
@@ -177,50 +156,19 @@ export const useSpreadsheetEquipments = (
...
-                    if (formatFetchedEquipments) {
-                        fetchedEquipments = formatFetchedEquipments(fetchedEquipments);
-                        spreadsheetEquipmentsByNodes.equipmentsByNodeId[nodeId] = fetchedEquipments;
-                    }
-                });
-            });
-
-            Promise.all(fetchers).then(() => {
-                dispatch(loadEquipments(type, spreadsheetEquipmentsByNodes));
+            const promises = loadEquipmentData(nodeIdsToFetch);
```
TheMaskedTurtle, line 166: it's kind of weird to call loadequipmentData(nodeIdsToFetch) when both comes from the hook, we want to call only loadEquipmentData() from the hook

src/components/spreadsheet/custom-columns/custom-columns-nodes-config.tsx
```diff
@@ -19,25 +25,130 @@ const styles = {
...
+        studyUuid,
+        currentRootNetworkUuid,
+        currentNode?.id,
+        undefined
+    );
+
+    const nodesToReload = useMemo(() => {
+        // Get all aliased nodes ids, except for Root and current node (both are always up-to-date)
+        return nodesAliases.filter((node) => node.id !== currentNode?.id && node.name !== ROOT_NODE_LABEL);
+    }, [currentNode?.id, nodesAliases]);
```
TheMaskedTurtle, line 68: I think this logic can be put in the hook too so that in the end we just call a simple loadEquipmentData() method

src/components/spreadsheet/custom-columns/custom-columns-nodes-config.tsx
```diff
@@ -19,25 +25,130 @@ const styles = {
...
+    }, []);
+
+    const handleRefresh = useCallback(() => {
+        if (nodesToReload?.length) {
+            const promises = loadEquipmentData(nodesToReload.map((n) => n.id as string));
+            Promise.all(promises).then(() => {
+                console.debug(`Refresh done for ${promises.length} nodes among ${nodesToReload.length} aliased nodes`);
+            });
+        }
+    }, [loadEquipmentData, nodesToReload]);
```
TheMaskedTurtle, line 85: Same goes for this

TheMaskedTurtle (changes requested):

src/translations/spreadsheet-en.ts
dbraquart: I wanted to be consistent with the dialog title:

'spreadsheet/custom_column/parameter_nodes': 'Parameter nodes aliases',

src/components/spreadsheet/custom-columns/custom-columns-nodes-config.tsx
dbraquart: the reason I return the promises to the caller : he can perform a specific action when all promises end. In the case above, it's only a console log (we could remove it). In the case of the useLoadEquipment we reset a state (cf _setIsFetching(false)_).

Maybe I could add a callback _onFetchingDone_ to allow specific action from the caller ?

src/components/spreadsheet/data-fetching/use-load-equipment.ts
dbraquart: yes, done

src/translations/spreadsheet-en.ts
TheMaskedTurtle: Yes but maybe we should change them all, parameter is not a verb in english ^^

src/components/spreadsheet/custom-columns/custom-columns-nodes-config.tsx
TheMaskedTurtle: Yes could be nice :)

src/components/spreadsheet/data-fetching/use-spreadsheet-equipments.ts
dbraquart: yes

src/components/spreadsheet/custom-columns/custom-columns-nodes-config.tsx
dbraquart: the nodes list to load/reload are not the same, thats why I pass them to the hook method.

If we want to hide this, I think I need two distinct methods:

- loadEquipmentData()
- reloadEquipmentData()

?

src/components/spreadsheet/data-fetching/use-load-equipment.ts
dbraquart: hook deleted in the meantime

src/components/spreadsheet/data-fetching/use-spreadsheet-equipments.ts
dbraquart: outdated

src/components/spreadsheet/custom-columns/custom-columns-nodes-config.tsx
dbraquart: I use redux action now

src/components/spreadsheet/custom-columns/use-node-aliases.ts
```diff
@@ -14,7 +14,8 @@ import { NodeAlias } from './node-alias.type';
 
 export const useNodeAliases = () => {
     const studyUuid = useSelector((state: AppState) => state.studyUuid);
-    const [nodeAliases, setNodeAliases] = useState<NodeAlias[]>([]);
+    // init value is undefined until we have successfully made a fetch
+    const [nodeAliases, setNodeAliases] = useState<NodeAlias[] | undefined>(undefined);
```
TheMaskedTurtle, line 18:
```suggestion
    const [nodeAliases, setNodeAliases] = useState<NodeAlias[]>();
```
this is equivalent

src/components/spreadsheet/custom-columns/custom-columns-nodes-config.tsx
```diff
@@ -20,31 +27,125 @@ const styles = {
...
+        () => ({
+            [NodesOptionId.CONFIG]: {
+                id: NodesOptionId.CONFIG,
+                label: 'spreadsheet/custom_column/option/parameter',
+                action: dialogOpen.setTrue,
+            },
+            [NodesOptionId.REFRESH]: {
+                id: NodesOptionId.REFRESH,
+                label: 'spreadsheet/custom_column/option/refresh',
+                action: () => handleRefresh(),
```
TheMaskedTurtle, line 94:
```suggestion
                action: handleRefresh,
```

src/components/spreadsheet/custom-columns/custom-columns-nodes-config.tsx
```diff
@@ -20,31 +27,125 @@ const styles = {
...
-                onClick={dialogOpen.setTrue}
-                disabled={disabled}
-            >
+            <Button sx={spreadsheetStyles.spreadsheetButton} size={'small'} onClick={handleClick} disabled={disabled}>
                 <BuildIcon sx={styles.icon} />
                 <FormattedMessage id="spreadsheet/custom_column/nodes" />
             </Button>
-
+            <Menu anchorEl={anchorEl} open={Boolean(anchorEl)} onClose={handleClose}>
+                {Object.values(nodesOptions).map(renderMenuEntry)}
```
TheMaskedTurtle, line 147:
```suggestion
               <MenuItem key={NodesOptionId.CONFIG} onClick={dialogOpen.setTrue}>
                    <FormattedMessage id={'spreadsheet/custom_column/option/parameter'} />
                </MenuItem>
                <Tooltip
                    title={
                        <FormattedMessage
                            id={'spreadsheet/custom_column/option/refresh/tooltip'}
                            values={{
                                aliases: nodesToReload?.map((node) => node.alias).join(', '),
                            }}
                        />
                    }
                >
                    <MenuItem
                        key={NodesOptionId.REFRESH}
                        onClick={handleRefresh}
                        disabled={nodesToReload ? nodesToReload.length === 0 : true}
                    >
                        <FormattedMessage id={'spreadsheet/custom_column/option/refresh/tooltip'} />
                    </MenuItem>
                </Tooltip>
```
Can we just define the components directly, all the dynamic thing is verbose and maybe not really useful

src/components/spreadsheet/data-fetching/use-fetch-equipment.ts
```diff
@@ -0,0 +1,104 @@
...
+    const isBuilt = useCallback(
+        (nodeId: string) =>
+            treeNodes?.find(
+                (node) =>
+                    node.id === nodeId && (node.type === NodeType.ROOT || isStatusBuilt(node.data?.globalBuildStatus))
+            ) !== undefined,
+        [treeNodes]
+    );
+
+    const fetchNodesEquipementData = useCallback(
```
TheMaskedTurtle, line 47: typo: equipment

src/components/spreadsheet/data-fetching/use-fetch-equipment.ts
```diff
@@ -0,0 +1,104 @@
...
+                        console.debug('Equipment data fetching and dispatch NOT done');
+                        errorMessage = err.message;
+                    })
+                    .finally(() => {
+                        if (onFetchingDone) {
+                            onFetchingDone({
+                                errorMessage: errorMessage,
+                            });
+                        }
+                    });
```
TheMaskedTurtle, line 97:
```suggestion
                    .catch((err: unknown) => {
                        console.debug('Equipment data fetching and dispatch NOT done');
                        if (onError) {
                            catchErrorHandler(err, onError);
                        }
                    });
```
why not something like that instead ?

TheMaskedTurtle (changes requested):

src/components/spreadsheet/custom-columns/custom-columns-nodes-config.tsx
dbraquart: yes for 2 items it does not worth it

src/components/spreadsheet/data-fetching/use-fetch-equipment.ts
```diff
@@ -0,0 +1,100 @@
...
+                    .catch((err) => {
+                        console.debug('Equipment data fetching and dispatch NOT done');
+                        snackError({
+                            messageTxt: err.message,
+                            headerId: 'SpreadsheetFetchError',
+                        });
+                    })
+                    .finally(() => {
+                        onFetchingDone && onFetchingDone();
+                    });
```
TheMaskedTurtle, line 93: Maybe it's not useful anymore ? We can just call `setIsFetching(false);` after the fetch in use-spreadsheet-equipments

src/components/spreadsheet/table-wrapper.tsx
```diff
@@ -487,10 +469,7 @@ export const TableWrapper: FunctionComponent<TableWrapperProps> = ({
                         currentNode={currentNode}
                         rowData={rowData}
                         columnData={reorderedColsDefs}
-                        fetched={
-                            equipments?.nodesId.find((nodeId) => nodeId === currentNode.id) !== undefined ||
-                            !!errorMessage
-                        }
+                        fetched={equipments?.nodesId.find((nodeId) => nodeId === currentNode.id) !== undefined}
```
TheMaskedTurtle, line 472:
```suggestion
                        fetched={!isFetching}
```
maybe it should simply be that ?

src/components/spreadsheet/data-fetching/use-fetch-equipment.ts
dbraquart: when use-spreadsheet-equipments calls fetchNodesEquipmentData(), this is async: we return asap while the promises are going on. The Promises.all.finally may occurs a long time after we return.

If we want to manage correctly the fetching/loading status, we need this callback onFetchingDone.

src/components/spreadsheet/table-wrapper.tsx
dbraquart: Yes it sounds equivalent

src/components/spreadsheet/data-fetching/use-fetch-equipment.ts
TheMaskedTurtle: You're right ! Good ! Then can you just replace `(nodeIds: Set<string>, onFetchingDone: undefined | (() => void))` by `(nodeIds: Set<string>, onFetchingDone?: () => void)` so we don't have to explicitely set undefined if we don't want to add a method

TheMaskedTurtle (approved):

src/components/spreadsheet/data-fetching/use-fetch-equipment.ts
dbraquart: yes better.

PR #2615, opened 2025-02-23
Dynamic spreadsheets with backend persistence

ayolab:


dbraquart (changes requested):
code: a few more remarks

src/translations/en.json
```diff
@@ -1097,6 +1097,8 @@
 
     "LoadStudyAndParentsInfoError": "An error occurred while loading current study's name and path",
     "InvalidNode": "Node not built",
+    "NoSpreadsheets": "You have no spreadsheets, please create one",
+    "spreadsheetConfigCollectionRetrievingError": "An error occurred while retrieving the spreadsheet configuration collection",
```
dbraquart, line 1101: to remove ?

dbraquart (changes requested):
code: a fex remarks
tests: WIP

dbraquart (changes requested):
tests: 2 issues found

dbraquart (changes requested):
code: ok
tests: ok. One small issue: we should remove sort/filter on column deletion
- add a sort + filter on a col
- remove the col
- re-create the col with same name/id/formula : the sort/filter is still here

dbraquart (approved):

PR #2640, opened 2025-03-03
corrects permanent limits validation

Mathieu-Deharbe:


src/components/dialogs/limits/limits-pane-utils.js
```diff
@@ -46,13 +46,13 @@ const temporaryLimitsValidationSchema = () => {
...
 
-const currentLimitsValidationSchema = () => ({
+const currentLimitsValidationSchema = (isModification = false) => ({
     [PERMANENT_LIMIT]: yup
         .number()
         .nullable()
         .positive('permanentCurrentLimitMustBeGreaterThanZero')
         .when([TEMPORARY_LIMITS], {
-            is: (temporaryLimits) => temporaryLimits.length > 0,
+            is: (temporaryLimits) => temporaryLimits && temporaryLimits.length > 0 && !isModification,
```
basseche, line 55:
```suggestion
            is: (temporaryLimits) => temporaryLimits?.length > 0 && !isModification,
```

basseche:
if Active limit set has no limit and you add limit without adding IST no warning is dislayed 

Capture d’écran du 2025-03-03 16-04-28

src/components/dialogs/limits/limits-pane-utils.js
basseche: i think test is not good for modification :  you don't take into account the case where the active limit set is empty and you add  a temporary limit.

PR #2645, opened 2025-03-04
Add reordering columns backend persistence

ayolab:


src/components/spreadsheet/columns-config.tsx
```diff
@@ -74,14 +77,26 @@ export const ColumnsConfig: FunctionComponent<ColumnsConfigProps> = ({ tabIndex,
...
+            localColumns.map((col) => col.uuid)
+        )
+            .then(() => {
+                dispatch(
+                    updateTableDefinition({
+                        ...tableDefinition,
+                        columns: localColumns,
+                    })
+                );
             })
```
dbraquart, line 91: When we are here (after validate) we may have changed a combinaison of lock, visible, or columns reorder. It's a pitie we always sent a back-end request to update the database even if we havent change the columns order. Could we avoid this ?

dbraquart (changes requested):
code: I request a change if possible
tests: ok

dbraquart (approved):
tests OK

PR #2660, opened 2025-03-05
Add Tool tip and Icon that indicates node status on previous value

basseche:
needs : https://github.com/gridsuite/commons-ui/pull/709 
to work properly


package.json
```diff
@@ -11,7 +11,7 @@
     "dependencies": {
         "@emotion/react": "^11.14.0",
         "@emotion/styled": "^11.14.0",
-        "@gridsuite/commons-ui": "0.88.1",
+        "@gridsuite/commons-ui": "file:../commons-ui/gridsuite-commons-ui-0.88.1.tgz",
```
basseche, line 14: will be replaced before merge

Mathieu-Deharbe (changes requested):
1. I don't think that the specs are respected here : 
I created a new load modification in a built node so I think I am in this case : 
-nœud courant réalisé et la modification réseau est en train d'être créée : pas d'icône ni de tooltip à côté de la valeur
But I have tooltips : 
image

2. The tooltip should appear when hovering the icon. For now it only appears when the text is hovered. This is what is usually done in that case and what is asked in the specs.

3. The tooltip looks different that in the specs : 
- on the bottom instead of the right
- tour text is left aligned instead of centered.
I prefer your version so it is fine for me, but you may confirm with Stéphane.

basseche:
> 1. I don't think that the specs are respected here :
>        I created a new load modification in a built node so I think I am in this case :
>        -nœud courant réalisé et la modification réseau est en train d'être créée : pas d'icône ni de tooltip à côté de la valeur
>        But I have tooltips :
>        image
> 
>     2. The tooltip should appear when hovering the icon. This is what is usually done in that case and what is asked in the specs. For now it only appears when the text is hovered.
> 
> 
> Not real problems, just to be checked :
> 
>     3. The tooltip looks different that in the specs :
> 
> 
>     * on the bottom instead of the right
> 
>     * tour text is left aligned instead of centered.
>       I prefer your version so it is fine for me, but you may confirm with Stéphane.
> 
> 
>     4. The ticket was only about  "modifications de poste / consommation / groupe / batterie / MCS". You added the tooltip on all the modifications, which is probably better. Was there any reason to ask for only half of them ?

1/ done
2/ done 
3/ done
4/No problem with that says Thibault (it's a good think)

thangqp (approved):

package.json
Mathieu-Deharbe: Ok but I will approve after you remove it.

Mathieu-Deharbe (approved):

PR #2664, opened 2025-03-06
Change mapping to post and pass ids (substations/voltage levels) in request body

antoinebhs:
Change get mappings for NAD and geo-data lines and substations fetching to post mapping with ids in the request body (it avoids oversized headers).


EstherDarkish (approved):
Code OK Test OK

PR #2667, opened 2025-03-07
Persists node aliases

Meklo:


FranckLecuyer (changes requested):
Tests : when defining different aliases for the same node, only the first alias can be used in a column formula definition ...

Meklo:
> Tests : when defining different aliases for the same node, only the first alias can be used in a column formula definition ...

Seen with PO will be dedicated to another issue

FranckLecuyer (approved):
Code review: OK
Tests: OK

PR #2677, opened 2025-03-11
Fix design: align RR form for case selection

sBouzols:


AAJELLAL (approved):
Test and code are OK.

souissimai (approved):

PR #2684, opened 2025-03-12
Add row index column and calculation functionality to spreadsheets

ayolab:


src/components/spreadsheet/config/common-column-definitions.ts
```diff
@@ -169,3 +174,39 @@ export const booleanColumnDefinition = (displayName: string, tab: string): ColDe
...
+
+            if (!params.node) {
+                return null;
+            }
+
+            // Get the actual displayed row index
+            const displayedRowIndex = params.api.getDisplayedRowCount() > 0 ? params.node.rowIndex : null;
+
+            // Add 1 to convert from 0-based to 1-based indexing
+            return displayedRowIndex !== null ? displayedRowIndex + 1 : null;
```
TheMaskedTurtle, line 196:
```suggestion
            return null;
```
What's the point of having these inedexes ?

src/components/spreadsheet/utils/cell-renderers.tsx
```diff
@@ -223,3 +232,115 @@ export const ContingencyCellRenderer = ({ value }: { value: { cellValue: ReactNo
         </Box>
     );
 };
+
+export const RowIndexCellRenderer = (props: any) => {
```
TheMaskedTurtle, line 236:
```suggestion
export const RowIndexCellRenderer = (props: CustomCellRendererProps) => {
```

src/components/spreadsheet/table-wrapper.tsx
```diff
@@ -173,15 +184,30 @@ export const TableWrapper: FunctionComponent<TableWrapperProps> = ({
...
+                ?.map((column) => {
+                    const s: ColumnState = {
+                        colId: column.id ?? '',
+                        pinned: 'left',
+                    };
+                    return s;
+                }) || [];
+
+        // Combine both sets of columns
+        lockedColumnsConfig.push(...userLockedColumns);
```
TheMaskedTurtle, line 208: you can just put `[lockedColumnsConfig, ...userLockedColumns]` in the apply

src/components/spreadsheet/table-wrapper.tsx
```diff
@@ -116,6 +118,10 @@ export const TableWrapper: FunctionComponent<TableWrapperProps> = ({
     const [rowData, setRowData] = useState<Identifiable[]>([]);
     const [equipmentToUpdateId, setEquipmentToUpdateId] = useState<string | null>(null);
 
+    const calculationSelections = useSelector((state: AppState) =>
+        activeTabUuid ? state.calculationSelections?.[activeTabUuid] || [] : []
+    );
```
TheMaskedTurtle, line 123: This is only used in the calculation hook, prefer to define it there

TheMaskedTurtle (changes requested):

src/hooks/use-grid-calculations.ts
```diff
@@ -0,0 +1,68 @@
...
+
+        // Update grid directly
+        const api = gridRef?.current?.api;
+        if (api) {
+            api.setGridOption('pinnedBottomRowData', newRows);
+        }
+    }, [calculationSelections, columnDefs, gridRef, hasData]);
+
+    // Debounce the update to prevent multiple unnecessary calculations
+    const debouncedUpdateCalculations = useDebounce(updateCalculations, 5);
```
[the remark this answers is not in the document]
ayolab: The goal here is mainly to avoid unnecessary calculations because the onModelUpdated event is triggered multiple times in succession. Doing so will add 500 milliseconds delay even if you are on the same tab and are just checking another type of calculation.

TheMaskedTurtle (approved):

achour94 (commented):
Tests: OK
Code: each time you test for calculation rows you use params.node?.rowPinned, i suggest to check rowType (CalculationRowType) params that we have in data

achour94 (approved):

PR #2704, opened 2025-03-18
Network modification panel with activation status by root network

souissimai:


antoinebhs:
org.gridsuite.modification.NetworkModificationException: MOVE_MODIFICATION_ERROR when moving modifications on a built node

antoinebhs:
When you change the global applicability, there is a double PUT request

antoinebhs:
Why when we change global activation status, we reset everything and get all statuses from computation etc but not when we change local activation status? (you can see that in the request that are sent) 
What's the difference between both?

antoinebhs:
image
Do we want "Aucune donnée" in addition to "Aucune modification" maybe to do in another PR?

src/components/spreadsheet/utils/chip-cell-renderer.tsx
```diff
@@ -0,0 +1,34 @@
+/**
```
antoinebhs, line 1: Not sure why this file is in spreadsheet/utils? Is it related to spreadsheet?

src/components/spreadsheet/utils/chip-cell-renderer.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+    return (
+        <Chip
+            label={label}
+            deleteIcon={isActivated ? <CheckCircleOutlineIcon /> : <CancelIcon />}
+            color="primary"
+            size="small"
+            variant={isActivated ? 'filled' : 'outlined'}
+            onDelete={onClick}
+            onClick={onClick}
```
antoinebhs, line 30: Same action ondelete/onclik invert modif activation status?

src/components/graph/menus/root-network-node-editor.tsx
```diff
@@ -153,6 +154,7 @@ const RootNetworkNodeEditor = () => {
             fetchRootNetworks(studyUuid)
                 .then((res: RootNetworkMetadata[]) => {
                     updateSelectedItems(res);
+                    dispatch(setRootNetworks(res));
```
antoinebhs, line 157: root networks are fetched twice when we open the application. Can we avoid that by using redux in both cases?
see "root-networks" request

antoinebhs:
In the case where the local applicability is set to false for both networks (in developer mode) and we change from developer mode to user mode, then the switch is ON but the modification is not applied, do we want to deal with this case?

src/components/graph/menus/network-modifications/cell-renderer-switch.tsx
```diff
@@ -0,0 +1,92 @@
...
+            };
+
+            updateModification(newStatus);
+            return newModifications;
+        });
+        api.stopEditing();
+    }, [modificationUuid, updateModification, setModifications, api]);
+
+    return (
+        <Tooltip title={<FormattedMessage id={modificationActivated ? 'disable' : 'enable'} />} arrow>
```
antoinebhs, line 79: I see that we don't have similar tooltip for local activation/deletion, is it expected?

src/components/graph/menus/network-modifications/chip-root-network-cell-renderer.tsx
```diff
@@ -0,0 +1,101 @@
...
+    const studyUuid = useSelector((state: AppState) => state.studyUuid);
+    const currentNode = useSelector((state: AppState) => state.currentTreeNode);
+    const [isLoading, setIsLoading] = useState(false);
+    const isAnyNodeBuilding = useIsAnyNodeBuilding();
+    const mapDataLoading = useSelector((state: AppState) => state.mapDataLoading);
+
+    const { snackError } = useSnackMessage();
+    const modificationUuid = data?.modificationInfos.uuid;
+
+    const modificationactivatedByRootNetwork = useMemo(
```
antoinebhs, line 34:
```suggestion
    const modificationActivatedByRootNetwork = useMemo(
```
?

src/components/graph/menus/network-modifications/chip-root-network-cell-renderer.tsx
```diff
@@ -0,0 +1,101 @@
+/**
```
antoinebhs, line 1: should we name these files local applicability switch and global applicability switch to make it clearer what it does? I'm not sure

antoinebhs:
I think there is an issue when you are applying a modification, you can drag modification to change their order which throw errors.

src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
```diff
@@ -748,155 +767,43 @@ const NetworkModificationNodeEditor = () => {
...
-                divider
+            <NetworkModificationsTable
+                handleCellClick={handleCellClick}
+                modifications={modifications}
+                setModifications={setModifications}
+                onRowDragStart={onRowDragStart}
+                onRowDragEnd={onRowDragEnd}
+                onRowSelected={handleRowSelected}
+                isDragging
+                isRowDragEnabled={!isImpactedByNotification() && !isAnyNodeBuilding && !mapDataLoading}
```
antoinebhs, line 799: Not sure it works, see my comment. Try to apply a generator dispatch and see that you can drag modification to change order

souissimai:
> When you change the global applicability, there is a double PUT request

fixed

src/components/spreadsheet/utils/chip-cell-renderer.tsx
souissimai: ondelete when clicking on icon , onclick when clicking on the chip , it's the same action yes

souissimai:
> org.gridsuite.modification.NetworkModificationException: MOVE_MODIFICATION_ERROR when moving modifications on a built node

fixed

souissimai:
> image Do we want "Aucune donnée" in addition to "Aucune modification" maybe to do in another PR?

To hide the "No Data" overlay in AG Grid , fixed

souissimai:
> ng and get all statuses fro

souissimai:
> Why when we change global activation status, we reset everything and get all statuses from computation etc but not when we change local activation status? (you can see that in the request that are sent) What's the difference between both? I think I saw cases where changing local applicability did not invalidate the node but I can't reproduce it 100%


- When you change the global activation status, everything resets, and all statuses are recalculated.

- When you change the local activation status:
    If it's for the current root network, everything resets immediately.
    If it's for another root network, the change is applied only when switching to it (no need to process it before).

souissimai:
> In the case where the local applicability is set to false for both networks (in developer mode) and we change from developer mode to user mode, then the switch is ON but the modification is not applied, do we want to deal with this case? It's simple to have only the global switch for the user mode but it does not always reflect the reality.

This case will not be handled in this ticket.
Right now, we only need to hide the RN columns when not in developer mode. Since it's not mentioned in the ticket, handling the modification issue should be considered in a separate ticket if needed.

src/components/graph/menus/root-network-node-editor.tsx
souissimai: fixed

src/components/graph/menus/network-modifications/chip-root-network-cell-renderer.tsx
souissimai: RootNetworkChipCellRenderer generates a chip for each modification per root network. I think this name is more generic

src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
souissimai: fixed , missing condition added it

antoinebhs:
When adding a new root network, modifications are not applied to it while on refresh they're applied as expected

antoinebhs:
> > Why when we change global activation status, we reset everything and get all statuses from computation etc but not when we change local activation status? (you can see that in the request that are sent) What's the difference between both? I think I saw cases where changing local applicability did not invalidate the node but I can't reproduce it 100%
> 
>     * When you change the global activation status, everything resets, and all statuses are recalculated.
> 
>     * When you change the local activation status:
>       If it's for the current root network, everything resets immediately.
>       If it's for another root network, the change is applied only when switching to it (no need to process it before).

From what I see, if I modify the local applicability of a modification on the selected root network, statuses are not fetched. Can you check?

antoinebhs:
Peek 26-03-2025 10-38
There seems to be a bug when dragging before last modification to last: sometimes its position is set to the first modification

src/components/graph/menus/network-modifications/network-modifications-table.tsx
```diff
@@ -0,0 +1,189 @@
...
+            );
+        },
+        [computeLabel, intl]
+    );
+
+    const columnDefs = useMemo(() => {
+        const staticColumns: ColDef<NetworkModificationInfos>[] = [
+            {
+                colId: 'modificationName',
+                rowDrag: !isRowDragDisabled || !isDragging,
```
antoinebhs, line 94: I'm not sure we need this isDragging that you added. Also I'm not sure you mean || but instead && ?

src/components/spreadsheet/utils/chip-cell-renderer.tsx
antoinebhs: Will be done in a separate PR: to move to commons-ui

src/components/spreadsheet/utils/chip-cell-renderer.tsx
```diff
@@ -0,0 +1,38 @@
...
+    label: string;
+    onClick: () => void;
+    isDisabled?: boolean;
+}
+
+export const ChipCellRenderer = (props: ChipCellRendererProps) => {
+    const { isActivated, label, onClick, isDisabled } = props;
+
+    return (
+        <Tooltip title={label} arrow>
```
antoinebhs, line 23: I'm not sure the title should be the label here but instead "Disable/Enable" depending on whether or not it's activated?

antoinebhs (approved):
Code OK
Test OK
:+1:

PR #2713, opened 2025-03-19
Fix infinite loader at root network deletion

sBouzols:
fix(RootNetworkNodeEditor): do not update the current root network uuid if deleted root networks don't contains it.


klesaulnier (approved):
Code: OK
Test: OK

AAJELLAL (approved):
Tests and code are OK.

PR #2723, opened 2025-03-20
Add regular expression support to custom columns

Meklo:


src/components/spreadsheet/custom-columns/math.ts
```diff
@@ -36,6 +36,9 @@ instance.import(
             // == instead of === to be able to compare strings to numbers
             return a === b;
         },
+        match: function (expr: string, variable: string) {
+            return RegExp(expr).exec(variable) !== null;
+        },
```
achour94, line 41:
```suggestion
        match: function (expr: string, variable: string, flags: string = '') {
            const escapedRegex = expr.replace(/\\/g, '\\\\');
            return RegExp(escapedRegex, flags).test(variable);
        },
```
- add flags option, and ask PO if we put i flag as defaut to be Case-insensitive
- replace exec with test
- escape '\' inside the function since it's related only to this one

Meklo: Seen with PO agreed with first point, minor adjustment that said. 
The default flag would be g (case sensitive) and i flag could be passed to ease up the use of the regex. I think it feels more natural for normie users to pass a flag that explicitely enables case insensitivity whereas if it's enabled by default they need to pass g flag to disable it which in itself has nothing to do with case sensitivity and isn't even relevant in this use case
Ok with your second point, it will be more efficient
Your third point touch on something a little bit trickier, I noticed putting a backslash in any formula will break it. So escaping it for all formulas is in my opinion better since it also works as a fix, I could test for its presence before calling replace but I don't know if it would be more efficient

achour94 (approved):

PR #2727, opened 2025-03-20
extra securities on null temporary limits

Mathieu-Deharbe:


src/components/dialogs/limits/limits-pane-utils.js
```diff
@@ -179,15 +179,17 @@ export const getAllLimitsFormData = (
 export const sanitizeLimitsGroups = (limitsGroups /*: OperationalLimitsGroup[]*/) =>
     limitsGroups.map(({ currentLimits, ...baseData }) => ({
         ...baseData,
-        currentLimits: {
+        currentLimits: currentLimits && {
```
basseche: it seems that if currentLimits is undefined or null currentLimits will be undefined maybe ternary with null in this case

Mathieu-Deharbe: Agreed, updated.

basseche (approved):

basseche:
Test OK

src/components/dialogs/limits/limits-pane-utils.ts
```diff
@@ -40,7 +42,7 @@ const temporaryLimitsValidationSchema = () => {
             .string()
             .nullable()
             .when([TEMPORARY_LIMIT_VALUE, TEMPORARY_LIMIT_DURATION], {
-                is: (limitValue, limitDuration) => limitValue || limitDuration,
+                is: (limitValue: number | null, limitDuration: number | null) => limitValue || limitDuration,
```
basseche, line 45: as sonar said : prefer nullish operator ?? instead ||

PR #2747, opened 2025-03-27
Add type predicates for notification types narrowing

sBouzols:


src/components/dialogs/parameters/common/computation-parameters-util.ts
```diff
@@ -5,16 +5,19 @@
...
-import { NotificationType, StudyUpdated, StudyUpdatedEventData } from '../../../../redux/reducer';
+import { NotificationType, StudyUpdated } from '../../../../redux/reducer';
 import { ComputingType, isValidComputingType } from '../../../computing-status/computing-type';
 
 export const isComputationParametersUpdated = (type: ComputingType, studyUpdated: StudyUpdated) => {
-    const studyUpdatedEventData = studyUpdated?.eventData as StudyUpdatedEventData;
+    // TODO fix my type
+    if (studyUpdated.type !== undefined) {
+        return false;
+    }
```
thangqp, line 15: we can use isComputationParametersUpdatedNotification(studyUpdated.eventData) ??

src/redux/reducer.ts
```diff
@@ -428,7 +473,7 @@ type RootNetworkUpdatedNotification = {
 
 type RootNetworkUpdateFailedNotification = {
     type: NotificationType.ROOT_NETWORKS_UPDATE_FAILED;
-    eventData: RootNetworkModifiedEventData;
+    eventData: RootNetworkUpdateFailedEventData;
```
thangqp, line 476: I think all we need is a type StudyUpdated in which we have eventData : un union type od all XXXEventData

src/components/graph/menus/root-network-node-editor.tsx
```diff
@@ -176,24 +176,18 @@ const RootNetworkNodeEditor = () => {
 
     const rootNetworkModifiedNotification = useCallback(
         (event: MessageEvent<string>) => {
-            const parsedEventData: unknown = JSON.parse(event.data);
-            const eventData = parsedEventData as RootNetworkModifiedEventData;
-            const updateTypeHeader = eventData.headers.updateType;
-            if (
-                updateTypeHeader === NotificationType.ROOT_NETWORK_MODIFIED ||
-                updateTypeHeader === NotificationType.ROOT_NETWORK_UPDATED
```
thangqp, line 184: I seems we have two updateType for this eventData, so the updateType of this event data should be an union of NotificationType.ROOT_NETWORK_MODIFIED | NotificationType.ROOT_NETWORK_UPDATED

src/redux/reducer.ts
```diff
@@ -347,19 +347,26 @@ export interface StudyUpdatedEventDataHeader {
...
     rootNetwork: UUID; // todo rename rootNetworkUuid in back as well
-    updateType: string;
+    updateType: NotificationType.LOADFLOW_RESULT;
 }
 
 interface RootNetworkModifiedEventDataHeaders {
     studyUuid: UUID;
     rootNetwork: UUID; // todo rename rootNetworkUuid in back as well
-    updateType: string;
+    updateType: NotificationType.ROOT_NETWORK_MODIFIED;
```
thangqp, line 362: not right here, but in case an XXXEventData having several updateType, we should use union of all supported NotificationType for field _updateType_

PR #2750, opened 2025-03-27
Migrate to Prettier v3

Tristan-WorkGH:


.prettierignore
```diff
@@ -1 +0,0 @@
-build
```
Meklo, line 1: Has it been translated into prettier.config.ts or was it useless ?

Tristan-WorkGH: Prettier read `.gitignore` and `.prettierignore` for ignored files, and the build folder is already in the .`gitignore` file.

Meklo (approved):

PR #2756, opened 2025-03-31
Fix node aliases configuration on node deletion

FranckLecuyer:


src/components/spreadsheet/custom-columns/use-node-aliases.ts
```diff
@@ -36,7 +38,7 @@ export const useNodeAliases = () => {
         } else {
             setNodeAliases(undefined);
         }
-    }, [snackError, studyUuid]);
+    }, [snackError, studyUuid, treeNodes]);
```
dbraquart, line 41: watching treeNodes is a bit too much for our needs. 

ex: any node built => we call 2 times GET /node-aliases for nothing + some effects are triggered

maybe we should use a new notification ALIAS_UPDATE ?

FranckLecuyer: The existing notifications on node deletion or node renaming are now used to update the node aliases.
So, these aliases are now updated only when a node is removed or when a node is renamed ...

dbraquart (changes requested):
code: ok
tests: I found 1 issue:
- in the tree we have a subtree N2 -> N3
- we remove the subtree N2 (incl. N3)
- in the aliases dialog, only N2 has been resetted

FranckLecuyer:
> code: ok tests: I found 1 issue:
> 
>     * in the tree we have a subtree N2 -> N3
> 
>     * we remove the subtree N2 (incl. N3)
> 
>     * in the aliases dialog, only N2 has been resetted

Now fixed in PR : https://github.com/gridsuite/study-server/pull/727

dbraquart (approved):

dbraquart (changes requested):
tests: on going

dbraquart (approved):

PR #2760, opened 2025-04-01
Change dependencies to avoid multiple effects in spreadsheet fetches

dbraquart:
To fix a specific case: a spreadsheet is open on node N1, we unbuild and build N1 again : all fetches are made twice in the spreadsheet view.


FranckLecuyer (approved):
Code review: OK
Tests: OK

PR #2762, opened 2025-04-02
Add measurements to load modification and update load dialogs to tabbed structure

ayolab:


FranckLecuyer (approved):
Tests: OK
Code review: OK

PR #2763, opened 2025-04-02
Fix the node name change : do not send a request to the backend if th…

EstherDarkish:
…e trimmed name is equal to its original value.


src/components/graph/menus/unique-check-name-input.tsx
```diff
@@ -109,7 +111,7 @@ export function UniqueCheckNameInput({
         const trimmedValue = value.trim();
 
         // if the name is unchanged, we don't do custom validation
-        if (!isDirty) {
+        if (!isDirty || defaultFieldValue.trim() === value.trim()) {
```
sBouzols, line 114: use `trimmedValue` instead or trim a second time `value` ?

EstherDarkish: done

sBouzols (approved):
Code Review OK
Tests OK
Console warning check OK

PR #2776, opened 2025-04-08
Migrate react-beautiful-dnd to @hello-pangea/dnd

achour94:
This migration was necessary to support React 19 and maintain compatibility with modern React versions.
react-beautiful-dnd is no longer actively maintained and does not officially support React 19. @hello-pangea/dnd is a direct fork that is actively maintained and has full support for React 19.


FranckLecuyer (approved):
Tests: OK

PR #2779, opened 2025-04-08
 index status by root network for study

souissimai:


antoinebhs (commented):
I think this PR should also rename all:
studyIndexationStatus
=> rootNetworkIndexationStatus 
that we have in gridstudy-app for consistency and readability

antoinebhs (commented):
// notification that the study is not indexed anymore then ask to refresh
=>             // notification that the root network is not indexed anymore then ask to refresh

klesaulnier (approved):

antoinebhs (approved):
Code OK
Test OK

PR #2783, opened 2025-04-10
Improve filters responsiveness

Meklo:
- Put onChange handlers in callback to prevent unnecessary computing
- Don't trigger redux state update if the comparator changes while there's no associated value
- Prioritize selected comparator UI value over the one in redux state to prevent the comparator selector rolling back 
- Trigger filter update immediately after setting a new selector since there's no apparent reason to debounce a selector based input
- Increase debounce timeout values to prevent filters update while filling options

Those aims at a less sluggish feeling while configuring comparator filters

Further optimization available : don't update all spreadsheet columns filters redux state when update one filter state ? Getting filters state out of redux ?


dbraquart (commented):
I could not really see perf improvments when we change a text filter on a big tab (ex: Line on french case).

- when we apply a filter, updateFilters (aggrid-filters-utils.ts) is always called a second time after the dispatch.
It does not seems to cost, but maybe we could avoid it. We could also add a checking function to not ask aggrid to filter
if the same filter is already applied.
- the debounce does not work well. Sometimes I strike pretty fast but intermediate filters are applied, freezing the
filter text input. This is maybe what we should work on.

PR #2739, opened 2025-03-26
new results filtering component

Mathieu-Deharbe:

PR #2791, opened 2025-04-12
Using context to manage global filter component

thangqp:
Following to PR: https://github.com/gridsuite/gridstudy-app/pull/2739

This PR refactors the previous PR to correct the bug in which selecting an option in the dropdown paper, the position of scrollbar is lost

This bug is due to the fact that the paper component is unmounted, by using context we can avoid unexpected unmount


src/components/results/common/global-filter/global-filter-paper.tsx
```diff
@@ -0,0 +1,201 @@
...
+        [intl, selectedGlobalFilters.length]
+    );
+
+    const addSelectedFilters = useCallback(
+        (values: TreeViewFinderNodeProps[] | undefined) => {
+            if (!values) {
+                return;
+            }
+
+            fetchElementsInfos(values.map((value) => value.id as UUID)).then((elements: ElementAttributes[]) => {
```
Mathieu-Deharbe, line 76: Not needed : id IS a UUID anyway : 
```suggestion
            fetchElementsInfos(values.map((value) => value.id)).then((elements: ElementAttributes[]) => {
```

thangqp: DONE in https://github.com/gridsuite/gridstudy-app/pull/2791/commits/be94b006d2c6523fcc413269097da148fa102383

PR #2795, opened 2025-04-15
Fix spreadsheet filters

achour94:


Tristan-WorkGH (changes requested):

Tristan-WorkGH (approved):
Well okay for the cast, but we will need to found a better way later to not have the cast.

PR #2797, opened 2025-04-16
Add spreadsheet notifications

dbraquart:
Now any spreadsheet change triggers a dedicated notification, consumed by all other open spreadsheets. In this case the other spreadsheets reload their data. 
All REST end-points that modify the spreadsheet have to pass througth the study-server for that notification purpose.


FranckLecuyer (approved):
Code review : OK
Tests: when we open the node aliases configuration dialog on the same study in 2 differents navigator tabs, a modification done in this dialog on one tab is not automatically synchronized in the other tab dialog : we have to close and reopen the dialog to see the modification ...

PR #2798, opened 2025-04-16
Root network panel minimized

souissimai:


src/components/root-network-panel-header.tsx
```diff
@@ -0,0 +1,143 @@
...
+                    headerId: 'createRootNetworksError',
+                });
+                setIsRootNetworksProcessing(false);
+            });
+    };
+    const minimizeRootNetworkPanel = useCallback(() => {
+        setIsRootNetworkPanelMinimized((prev) => !prev);
+    }, [setIsRootNetworkPanelMinimized]);
+    const LeftPanelIconComponent = isRootNetworkPanelMinimized ? LeftPanelOpen : LeftPanelClose;
+    const leftPanelIconColor = theme === LIGHT_THEME ? 'rgba(0, 0, 0, 0.54)' : '#fff';
```
klesaulnier, line 107: I think we should export this to commons-ui
I believe the safe way to use icons from other library than material UI is to make it in commons-ui

Something looking like this
```
const LeftPanelOpenIcon = (style) => {
    const theme = useTheme();
    
    return (
        <LeftPanelOpen style={{ width: 24, height: 24, fill: theme.iconColor, ...style }}/>
    );
}
```

With this, developpers could use this icon without having to define size and colors depending on theme

src/components/graph/menus/root-network-minimized-panel-content.tsx
```diff
@@ -0,0 +1,61 @@
...
+        flexGrow: 1,
+        fontWeight: 'bold',
+        marginLeft: theme.spacing(2),
+    }),
+    icon: {
+        display: 'flex',
+        justifyContent: 'center',
+        alignItems: 'center',
+    },
+};
```
klesaulnier, line 35:
```suggestion
const styles = {
    header: (theme: Theme) => ({
        display: 'flex',
        alignItems: 'center',
        padding: theme.spacing(1),
    }),
    minimizedPanel: (theme: Theme) => ({
        flexGrow: 1,
        marginLeft: theme.spacing(2),
    }),
    icon: {
        display: 'flex',
        justifyContent: 'center',
        alignItems: 'center',
    },
};
```

klesaulnier (changes requested):
Some adjustements

src/components/root-network-panel-header.tsx
souissimai: you  're right

src/components/root-network-panel.tsx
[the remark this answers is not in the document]
souissimai: ok

package.json
```diff
@@ -14,7 +14,7 @@
         "@gridsuite/commons-ui": "0.95.0",
         "@hello-pangea/dnd": "^18.0.1",
         "@hookform/resolvers": "^4.0.0",
-        "@mui/icons-material": "^5.16.14",
+       "@mui/icons-material": "^5.16.14",
```
klesaulnier, line 17:
```suggestion
        "@mui/icons-material": "^5.16.14",
```

src/components/app-wrapper.jsx
```diff
@@ -118,6 +118,9 @@ const lightTheme = createTheme({
     link: {
         color: 'blue',
     },
+    icon: {
+        color: '#757575',
+    },
```
klesaulnier, line 123: as stated in commons-ui PR, no need to create new theme properties I think

klesaulnier (approved):

PR #2800, opened 2025-04-16
[Dynamic simulation] Button close of plot no longer works

thangqp:
Corrections in this PR:

- Button Close and Zoom of each plot now handled by onMouseDown event instead of onClick due to the onClick event is not well captured from the usage of gridlayout library
- Using Grid system instead of Box for plot tabs header which is more compatible to our layout system

### DIFF
--- a/src/components/results/dynamicsimulation/dynamic-simulation-result-time-series.tsx
+++ b/src/components/results/dynamicsimulation/dynamic-simulation-result-time-series.tsx
@@ -7,7 +7,7 @@
 
 import CloseIcon from '@mui/icons-material/Close';
 import AddIcon from '@mui/icons-material/Add';
-import { Box, LinearProgress, Theme } from '@mui/material';
+import { Box, Grid, LinearProgress, Theme } from '@mui/material';
 import DynamicSimulationResultChart from './timeseries/dynamic-simulation-result-chart';
 import { memo, SyntheticEvent, useMemo, useState } from 'react';
 import { useIntl } from 'react-intl';
@@ -27,10 +27,11 @@ import { DropResult } from '@hello-pangea/dnd';
 const styles = {
     root: {
         height: '100%',
+        maxWidth: '100vw',
     },
     addButton: (theme: Theme) => ({
         borderRadius: '50%',
-        marginRight: theme.spacing(10),
+        marginRight: theme.spacing(2),
         color: theme.palette.primary.main,
     }),
     loader: {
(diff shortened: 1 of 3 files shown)


src/components/results/dynamicsimulation/dynamic-simulation-result-time-series.tsx
```diff
@@ -123,61 +124,65 @@ const DynamicSimulationResultTimeSeries = memo(function ({
...
-                                                        id: 'DynamicSimulationResultTab',
-                                                    })} ${tab.id}`}
-                                                    <TooltipIconButton
-                                                        tooltip={intl.formatMessage({
-                                                            id: 'DynamicSimulationCloseTab',
-                                                        })}
-                                                        size="small"
-                                                        component="span"
-                                                        onClick={handleClose(index)}
+                <Grid container sx={styles.root}>
```
thangqp, line 127: Grid configured with container make the layout does not work, i will revert to Box

thangqp: DONE in https://github.com/gridsuite/gridstudy-app/pull/2800/commits/b3db4997a5fe1e66b71ae04f2924e317635e6208

src/components/results/dynamicsimulation/timeseries/dynamic-simulation-result-series-chart.tsx
```diff
@@ -99,37 +99,45 @@ function DynamicSimulationResultSeriesChart({
...
-                                <Tooltip
-                                    title={intl.formatMessage({
-                                        id: 'DynamicSimulationPlotScaleEnable',
-                                    })}
-                                >
-                                    <FitScreenSharpIcon />
-                                </Tooltip>
-                            )}
-                        </ToggleButton>
+                        {plotScale ? (
```
Mathieu-Deharbe, line 102: I think "plotScale" is misleading. Isn't it a focus on the plot which then takes the whole viewport ? I think this is clearer : 
```suggestion
                        {plotForegrounded ? (
```
(or simply plotFullViewport)

Mathieu-Deharbe (approved):
OK.

But the drag and drop is still buggy.
When you try to drag to the top right, it "slides" towars the left and vice versa.

thangqp:
> OK.
> 
> But the drag and drop is still buggy. When you try to drag to the top right, it "slides" towars the left and vice versa.

If you drag quickly, it should work correctly..

PR #2804, opened 2025-04-18
[Dynamic Simulation] Correction for Sync of zoom between plots

thangqp:
**Observation:** a plot is not synced with a zoom action from another plot unless we manipulate (zoom, etc.) at least once on this plot.

**Reason:** the default layout for Xaxis is configured with field 'autorange' = true..

**Solution:** when sync from other plot, we force 'autorange' = false


Mathieu-Deharbe (approved):
Fine but just in case : 
it works with the zooming and the horizontal panning, but not with the vertical panning. They are never synchronized. Nor with the autoscale button.

thangqp:
> Fine but just in case : it works with the zooming and the horizontal panning, but not with the vertical panning. They are never synchronized. Nor with the autoscale button.

sync is only for horizontal and not for vertical, it is no sense when synchronizing in vertical

PR #2808, opened 2025-04-18
Update EquipmentTable to configure rowBuffer for better performance

ayolab:


TheMaskedTurtle (commented):
Ok not sure it improves much, we'll see with Christine

Meklo (approved):

PR #2809, opened 2025-04-18
Add visual feedback when at least one node alias is configured

ayolab:


TheMaskedTurtle (changes requested):
Be careful there is already a badge when there is an alias with no associated node, and if both conditions happen they overlap

dbraquart (approved):

PR #2810, opened 2025-04-18
[Dynamic Simulation] Prevent plots unmounted when switching tabs

thangqp:
It seems a regression due to the migration of library autosizer

**Observation:** when we have two tabs of plots, and zooming on a plot of the first tab. Switching to the second tab then back to the first => plot is reset and previous zooming is lost.

**Reason:** when AutoSizer computes width and height from a hidden parent component, width and height may be zero and AutoSizer ignore rendering children (regression from new version autosizer)

**Solution:** fortunatly, the recent new version of AutoSizer provides a props to override this behaviour, i.e. doNotBailOutOnEmptyChildren

**Enhancement:** when providing width and height for the GridLayout, we provide a filter to take only valid computed dimensions, otherwise, the previous valid dimensions are taken into account.. This is usefull when a tab content is waked up from display=hidden to display=flex and we do not see a flash that GridLayout reorganizes plots


Mathieu-Deharbe (approved):
Nice

PR #2812, opened 2025-04-18
Update formula validation to accept numbers for ENUM column type

ayolab:


src/components/spreadsheet/custom-columns/formula-validator.ts
```diff
@@ -27,7 +27,7 @@ export const validateFormulaResult = (value: any, type: COLUMN_TYPES): Validatio
             };
         case COLUMN_TYPES.ENUM:
             return {
-                isValid: typeof value === 'string',
+                isValid: typeof value === 'string' || typeof value === 'number',
```
TheMaskedTurtle, line 30: Do you think we can order propositions in the filter when it's numbers ?

ayolab: Done

TheMaskedTurtle (approved):

PR #2828, opened 2025-04-28
Fix logs filter initialization

ayolab:


FranckLecuyer (approved):
Code review: OK
Tests: OK

PR #2845, opened 2025-05-02
Add option for etalab geographical map tiles

jonenst:
needs network diagram viewer release


jonenst:
done in https://github.com/powsybl/powsybl-network-viewer/pull/194 and https://github.com/gridsuite/commons-ui/pull/795

PR #2814, opened 2025-04-23
implement creation voltage level topology modification

ghazwarhili:

PR #2865, opened 2025-05-14
Fix Open vs Close in modifying voltage level topology

thangqp:
Following to PR: https://github.com/gridsuite/gridstudy-app/pull/2814


src/components/dialogs/network-modifications/voltage-level-topology-modification/voltage-level-topology-modification-dialog.tsx
```diff
@@ -42,8 +42,8 @@ const formSchema = yup.object().shape({
         .of(
             yup.object().shape({
                 [SWITCH_ID]: yup.string(),
-                [PREV_CONNECTION_STATUS]: yup.string(),
-                [CURRENT_CONNECTION_STATUS]: yup.boolean().nullable(),
+                [PREV_CONNECTION_STATUS]: yup.boolean().nullable(), // presents 'open'
```
etiennehomer, line 45: These statuses do not present the same value ?

etiennehomer (approved):

PR #2867, opened 2025-05-15
refactor chips

AbdelHedhili:


FranckLecuyer (changes requested):
Tests: in some cases, we can't delete a chip and we do not see the '+' chip after the last chip displayed ...

FranckLecuyer (approved):

PR #2889, opened 2025-05-22
generic filters always at the end of the menus

Mathieu-Deharbe:


src/components/results/common/global-filter/global-filter-paper.tsx
```diff
@@ -66,10 +66,13 @@ function GlobalFilterPaper({ children, autocompleteRef }: Readonly<GlobalFilterP
...
     useEffect(() => {
         fetchSubstationPropertiesGlobalFilters().then(({ substationPropertiesGlobalFilters }) => {
-            setCategories([
+            const sortedCategories = [
                 ...standardCategories,
                 ...(substationPropertiesGlobalFilters ? Array.from(substationPropertiesGlobalFilters.keys()) : []),
-            ]);
+            ]
+                // generic filters always at the end of the menus
+                .sort((a) => (a === FilterType.GENERIC_FILTER ? 1 : -1));
```
etiennehomer, line 74: Are you sure that the rest of the array wont be moved ?

It's looks more common to use something like : 
sortedCategories.push(sortedCategories.splice(sortedCategories.indexOf(FilterType.GENERIC_FILTER), 1)[0]);

Mathieu-Deharbe: Ok it is clearer and more cautious, but I securised it just in case.

etiennehomer (approved):

PR #2933, opened 2025-06-10
Fix diagram grid layout titles

sBouzols:


src/components/diagrams/hooks/use-diagram-title.ts
```diff
@@ -0,0 +1,42 @@
...
+                return `${diagram.voltageLevelIds.join(', ')}`;
+            } else if (diagram.type === DiagramType.NAD_FROM_CONFIG) {
+                return `${diagram.nadName}`;
+            }
+            return `diagram type unknown`;
+        },
+        [translate]
+    );
+    return getDiagramTitle;
+};
```
EstherDarkish, line 42:
```suggestion
export const useDiagramTitle = () => {
    const { translate } = useLocalizedCountries();

    return useCallback(
        (diagram: Diagram, svgData: Svg): string => {
            const getCountrySuffix = (): string => {
                const country = (svgData as SldSvg).additionalMetadata?.country;
                return country ? ` - ${translate(country)}` : '';
            };

            switch (diagram.type) {
                case DiagramType.VOLTAGE_LEVEL:
                    return `${diagram.voltageLevelId}${getCountrySuffix()}`;

                case DiagramType.SUBSTATION:
                    return `${diagram.substationId}${getCountrySuffix()}`;

                case DiagramType.NETWORK_AREA_DIAGRAM:
                    return diagram.voltageLevelIds.join(', ');

                case DiagramType.NAD_FROM_CONFIG:
                    return diagram.nadName;

                default:
                    return 'diagram type unknown';
            }
        },
        [translate]
    );
};
```
Code cleaning

PR #2935, opened 2025-06-10
use the same validation schema for creation and modification dialog

ghazwarhili:
This PR covers the following scenarios:

1) When "En charge" is checked and "Mode de réglage" is set to "Voltage Regulation" and make the following fields mandatory:

    The voltage setpoint field is mandatory

    The regulation mode :

        If it is set to Local then the regulating side is required

        If it is set to Remote then the voltage level  and the equipment are required
        
2) "OnLoad" checkbox can only be toggled by clicking directly on it (CSS fix)


EtienneLt (approved):
Test OK

PR #2963, opened 2025-06-18
Update some tabular modifications

dbraquart:


AbdelHedhili (changes requested):
test : connected1 and connected2 fields are not applied

dbraquart:
> test : connected1 and connected2 fields are not applied

yes!
in fact i did not add connected1/2 cause it was already there. But it does not work on main branch. Fixed in commit dbd534a

AbdelHedhili (approved):

PR #2965, opened 2025-06-19
feat: csv format depend on local

AbdelHedhili:


dbraquart (approved):
code: maybe some renaming
tests: ok

PR #2976, opened 2025-06-23
Add reset confirmation dialog for parameter resets

achour94:


sBouzols (approved):
Code Review OK
Tests OK
Console warning check : full of errors and warnings but not created by this PR then OK...

PR #2981, opened 2025-06-24
clean clipboard after copy self subtree

souissimai:


src/components/network-modification-tree-pane.jsx
```diff
@@ -385,7 +408,7 @@ export const NetworkModificationTreePane = ({ studyUuid, studyMapTreeDisplay, cu
                     }
                 );
                 //Do not wait for the response, after the first CUT / PASTE operation, we can't paste anymore
-                dispatch(setNodeSelectionForCopy(noNodeSelectionForCopy));
+                // dispatch(setNodeSelectionForCopy(noNodeSelectionForCopy));
```
klesaulnier, line 411: To revert I think

src/components/network-modification-tree-pane.jsx
```diff
@@ -258,6 +268,12 @@ export const NetworkModificationTreePane = ({ studyUuid, studyMapTreeDisplay, cu
                         )
                     );
                 });
+                if (
+                    studyUpdatedForce.eventData.headers.movedNode === nodeSelectionForCopyRef.current.nodeId ||
```
klesaulnier, line 272: I think this check is already made in "isSubtreeImpactedByMovedNode" ?

src/components/network-modification-tree-pane.jsx
```diff
@@ -266,6 +282,12 @@ export const NetworkModificationTreePane = ({ studyUuid, studyMapTreeDisplay, cu
                         );
                     }
                 );
+                if (
+                    studyUpdatedForce.eventData.headers.movedNode === nodeSelectionForCopyRef.current.nodeId ||
```
klesaulnier, line 286: same here ?

PR #2986, opened 2025-06-26
Remove global warning panel on update dialogbox and invalid node.

ipirog:
Remove global warning panel on modification dialogbox when the current node is invalid.


Mathieu-Deharbe (approved):
Code OK, test OK.

PR #3009, opened 2025-07-07
Fix root network chip : update on notification reception

ayolab:
Refresh excluded modifications when notifications are received.


souissimai (approved):
tested

PR #3010, opened 2025-07-07
save the result of any filter's combination as a new explicit naming filter (list of equipment id's)

benrejebmoh:


dbraquart (commented):
code: just 1 remark
tests: filter creation is OK but:
- action should be set in the 3-dot menu for each tab ?
- filter ids are not sorted as in the spreadsheet

PR #3022, opened 2025-07-10
Fix root network panel : update on notification reception

ayolab:


AAJELLAL (approved):
Tests and code are ok.

PR #3024, opened 2025-07-11
Add limit sets tabular modification

Meklo:
Requires https://github.com/gridsuite/commons-ui/pull/807

### DIFF
--- a/src/components/dialogs/network-modifications/limit-sets/limit-sets-tabular-modification-form.tsx
+++ b/src/components/dialogs/network-modifications/limit-sets/limit-sets-tabular-modification-form.tsx
@@ -176,6 +176,12 @@ export function LimitSetsTabularModificationForm({ dataFetching }: Readonly<Tabu
         language: language,
     });
 
+    const handleChange = useCallback(() => {
+        setTypeChangedTrigger(!typeChangedTrigger);
+        clearErrors(MODIFICATIONS_TABLE);
+        setValue(MODIFICATIONS_TABLE, []);
+    }, [clearErrors, setValue, typeChangedTrigger]);
+
     const watchTable = useWatch({
         name: MODIFICATIONS_TABLE,
     });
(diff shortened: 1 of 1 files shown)


src/components/dialogs/network-modifications/limit-sets/limit-sets-modification-dialog.tsx
```diff
@@ -0,0 +1,140 @@
...
+                let modification = formatModification(row);
+                Object.keys(modification).forEach((key) => {
+                    modification[key] = row[key];
+                });
+                if (modification[SIDE] === BranchSide.ONE) {
+                    modification.operationalLimitsGroup1 = [
+                        formatOperationalLimitGroupsFrontToBack(modification, amountMaxTemporaryLimits, BranchSide.ONE),
+                    ];
+                }
+                if (modification[SIDE] === BranchSide.TWO) {
```
FranckLecuyer, line 82: else if  ?

src/components/dialogs/network-modifications/limit-sets/limit-sets-tabular-modification-form.tsx
```diff
@@ -0,0 +1,268 @@
...
+                <GridItem>
+                    <CsvDownloader
+                        columns={csvColumns}
+                        datas={commentLines}
+                        filename={equipmentType + '_modification_template'}
+                        disabled={!csvColumns}
+                        separator={language === LANG_FRENCH ? ';' : ','}
+                    >
+                        <Button variant="contained" disabled={!csvColumns}>
+                            <FormattedMessage id="GenerateSkeleton" />
```
FranckLecuyer, line 243: The generate skeleton action is triggered even if the mouse is outside right of the button, when clicking ...

src/components/dialogs/network-modifications/limit-sets/limit-sets-tabular-modification-form.tsx
```diff
@@ -0,0 +1,268 @@
...
+    });
+
+    useEffect(() => {
+        computeRepeatableColumns();
+    }, [computeRepeatableColumns]);
+
+    useEffect(() => {
+        setIsFetching(dataFetching);
+    }, [dataFetching]);
+
```
FranckLecuyer, line 149: I think we must check, on import, if the different fields in the csv are of the right type, and possibly of the expected value

side should be ONE or TWO
permanentLimit should be a numeric
modificationType should be ADDED or MODIFIED or REPLACED
temporaryLimitsModificationType should ADDED or MODIFIED or DELETED or REPLACED
acceptableDuration should be a numeric
value should be a numeric

For the enum type, the skeleton should also contain these expected values

This is to be consistent with the checks that have recently been added on the other tabular creation/modification

FranckLecuyer (changes requested):

PR #3028, opened 2025-07-16
create security node and adjust creation rules for network modificati…

souissimai:
…on nodes


AAJELLAL (approved):
Test and code are OK.

PR #3040, opened 2025-07-18
First step in adding global filter on shortcircuit analysis panel

ipirog:


src/components/results/shortcircuit/shortcircuit-analysis-result-tab.tsx
```diff
@@ -145,6 +166,13 @@ export const ShortCircuitAnalysisResultTab: FunctionComponent<ShortCircuitAnalys
                     <Tab label={<FormattedMessage id={'Results'} />} />
                     <Tab label={<FormattedMessage id={'ComputationResultsLogs'} />} />
                 </Tabs>
+                {resultOrLogIndex === RESULTS_TAB_INDEX && tabIndex === ShortCircuitAnalysisResultTabs.ALL_BUSES && (
+                    <GlobalFilterSelector
+                        onChange={handleGlobalFilterChange}
+                        filters={globalFilterOptions}
+                        filterableEquipmentTypes={filterableEquipmentTypes}
```
Mathieu-Deharbe, line 173: You should disable the selection of invalid filters, like in the other results pages : 
```suggestion
                        filterableEquipmentTypes={filterableEquipmentTypes}
                        genericFiltersStrictMode={true}
```
(will hide filters types that are not part of filterableEquipmentTypes)

Mathieu-Deharbe (commented):
No matter the global voltage filter that I pick I always get all the results (same with the countries). Is it normal ?
always same result

src/components/results/shortcircuit/shortcircuit-analysis-result-tab.tsx
ipirog: If the global filter is not applied, you may have made the mistake to use an old study-server?

ipirog: image
The global filter is not always near the tabs

Mathieu-Deharbe: Yes but this is a bad thing. It should be harmonized according to Stephane decision.

ipirog: image

ipirog: It's okay now :p

src/components/results/shortcircuit/shortcircuit-analysis-result-tab.tsx
```diff
@@ -128,6 +135,20 @@ export const ShortCircuitAnalysisResultTab: FunctionComponent<ShortCircuitAnalys
         }
     }, []);
 
+    const filterableEquipmentTypes: EQUIPMENT_TYPES[] = useMemo(() => {
+        return [EQUIPMENT_TYPES.VOLTAGE_LEVEL];
```
Mathieu-Deharbe, line 139: Shouldn't `SUBSTATION` be included as well ? If I read the JIRA correctly it states : 

> Il faut pouvoir utiliser un filtre générique listant des sites ou des postes pour filtrer les buses.

Mathieu-Deharbe (approved):

PR #3071, opened 2025-07-28
Refactor column state management in AG Grid to fix scroll bar behavior

achour94:


dbraquart (approved):
code: nice refacto
tests: ok

PR #3073, opened 2025-07-29
Refactor column state management in AG Grid to fix scroll bar behavior

achour94:


dbraquart (approved):
code: ok, nice refacto
test: I could not make a valid test changing "onLoad", but it seems unrelated to this PR

PR #3083, opened 2025-07-31
Fix NetworkMapTab update after receiving a `loadflowResultNotification`

sBouzols:
fix(NetworkMapTab): Add nodeUuid check in `loadflowResultNotification` and don't call `reloadMapEquipments` directly but manage it throught `reloadMapNeeded` for both auto/manual modes.

remove useless dependency/trigger in the useEffect


EstherDarkish (approved):
Code OK Test OK

PR #3087, opened 2025-07-31
Fix NetworkMapTab update at study impacts

sBouzols:
fix(useGetStudyImpacts): Add nodeUuid check for notification consumption and `setReloadMapNeeded` if there are some impacts


EstherDarkish (approved):
Code OK Test OK

PR #3104, opened 2025-08-06
Add CSV generation from filters

achour94:


dbraquart (changes requested):
code: a suggestion to remove Sonar issues
tests: the feature works fine!
- found 1 issue about comment lines (pushed 1 commit for that point)
- found 1 regression: global filter does not work anymore in spreadsheet (err http 500)

+ remark:  when a filter does not match any equipment, nothing happens (no template downloaded, no feedback): it's a bit disturbing.

PR #3105, opened 2025-08-06
Grid enhancements : consistent cards order and proportions

ayolab:


AAJELLAL (approved):
Tests and code are OK.

PR #3108, opened 2025-08-07
Short Circuit debug mode

thangqp:


src/services/short-circuit-analysis.ts
```diff
@@ -0,0 +1,27 @@
...
+import { UUID } from 'crypto';
+import { backendFetch } from './utils';
+
+const PREFIX_SHORT_CIRCUIT_ANALYSIS_SERVER_QUERIES = import.meta.env.VITE_API_GATEWAY + '/shortcircuit';
+
+function getShotCircuitAnalysisUrl() {
+    return `${PREFIX_SHORT_CIRCUIT_ANALYSIS_SERVER_QUERIES}/v1/`;
+}
+
+export function downloadDebugFileShortCircuitAnalysis(resultUuid: UUID): Promise<Response> {
```
etiennehomer, line 17: We could have a generic method for all computation nope ?

etiennehomer (approved):

src/services/short-circuit-analysis.ts
thangqp: IMO, not really usefull to generalize this method.. the method is an API method which refers to a server via a specfic url..
In gridsuite, we do not have any generalized API method for different servers

PR #3113, opened 2025-08-08
One bus short circuit debug mode

thangqp:


etiennehomer (approved):

thangqp:
> The loader in the SLD for "one bus shortcircuit" computation is broken
Not reproduced when testing at my local machine.

PR #3114, opened 2025-08-11
Fix evaluation pb when using multiple generic filters in spreadsheet global filter

FranckLecuyer:


dbraquart (approved):
code: ok
tests: ok

PR #3120, opened 2025-08-13
Resizable panels for Tree tab

ayolab:


AAJELLAL (approved):
Test and Code are OK.

PR #3125, opened 2025-08-13
Modify HVDC through SLD and network map context menu

Meklo:


FranckLecuyer (approved):
Tests: OK
Code review: just one remark

PR #3157, opened 2025-08-25
Add new fields in loadflow results

klesaulnier:


souissimai (approved):

antoinebhs:
I forced the merge because we considered the duplication was not a problem here (few lines modified, no other improvements to reduce it).

PR #3187, opened 2025-08-29
Add dev mode management to mask bolt/lock icons

ipirog:


src/components/diagrams/diagram-common.ts
```diff
@@ -61,13 +61,18 @@ export const styles = {
...
         '& .sld-flash, .sld-lock': {
-            stroke: 'none',
             fill: theme.palette.text.primary,
+            stroke: 'none',
         },
         '& .arrow': {
             fill: theme.palette.text.primary,
         },
     }),
+    divSingleLineDiagramDevModeDisabled: (theme: Theme) => ({
```
Mathieu-Deharbe, line 71: I think that here the naming should be independent from the devMode and describe more specifically what it does : 
```suggestion
    divSingleLineDiagramHideLock: (theme: Theme) => ({
```

Mathieu-Deharbe (approved):
It works and seems fine, especially for a temporary fix. I didn't look for a way to do the job on the map.

PR #3203, opened 2025-09-02
fix order for lines creation catalog

EtienneLt:


src/components/dialogs/line-types-catalog/line-types-catalog-selector-dialog.tsx
```diff
@@ -168,7 +168,8 @@ export default function LineTypesCatalogSelectorDialog({
             return [];
         }
 
-        const uniqueAreas = [...new Set(limitsData.map((limit) => limit.area))];
+        let uniqueAreas = [...new Set(limitsData.map((limit) => limit.area))];
+        uniqueAreas = uniqueAreas.toSorted((a, b) => a.localeCompare(b, undefined, { numeric: true }));
```
ghazwarhili, line 172: return [...new Set(limitsData.map((limit) => limit.area))]
    .toSorted((a, b) => a.localeCompare(b, undefined, { numeric: true }))
    .map((area) => ({ id: area, label: area }));

PR #3204, opened 2025-09-02
Update name in front

ipirog:


Mathieu-Deharbe (changes requested):
I am almost sure that in `two-windings-transformer-creation-dialog.jsx` you should change those functions as well : 
```
const getRegulationTypeForEdit = (twt, tap) => {
        return tap?.regulatingTerminalId != null
            ? tap?.regulatingTerminalId === twt.equipmentId
                ? REGULATION_TYPES.LOCAL.id
                : REGULATION_TYPES.DISTANT.id
            : null;
    };
```
```
const getTapSideForEdit = (twt, tap) => {
        return tap?.regulatingTerminalId === twt.equipmentId
            ? tap?.regulatingTerminalVlId === twt?.voltageLevelId1
                ? SIDE.SIDE1.id
                : SIDE.SIDE2.id
            : null;
    };
```
Sadly they are not typescripted yet.

PR #3212, opened 2025-09-04
Fix filter TypeScript definitions

Tristan-WorkGH:
Fix incomplete TypeScript definitions of filters DTO.  
Types and enum create from _filter-lib_ library sources.

No code/logic modified, only typescript definitions modifications, but `rules.add(...)` directly to have tsc infer and check Expert rules content.


antoinebhs:
Outdated

PR #3218, opened 2025-09-04
Fix : add tooltip maxWidth style to card header

ayolab:


sBouzols (approved):
Code Review : there is an underlying bug in the OverflowableText.
We should be able to use percentage to set the maxWidth of the tooltip but it breaks alignment today. Let's accept this for now but it's not really robust.

PR #3225, opened 2025-09-05
Add new fields to NmK security analysis results

klesaulnier:


src/components/results/securityanalysis/security-analysis-result-utils.ts
```diff
@@ -608,28 +689,65 @@ export const securityAnalysisTableNmKConstraintsColumnsDefinition = (
...
+                filterComponentParams: {
+                    filterParams: {
+                        dataType: FILTER_DATA_TYPES.NUMBER,
+                        comparators: Object.values(FILTER_NUMBER_COMPARATORS),
+                        ...filterParams,
+                    },
+                },
+            },
+        }),
+        makeAgGridCustomHeaderColumn({
```
antoinebhs, line 729: wrapper to avoid duplication (?)

src/translations/fr.json
```diff
@@ -376,7 +376,6 @@
     "ViolationType": "Type de contrainte",
     "Bus": "Noeud électrique",
     "LimitSide": "Côté",
-    "LimitName": "Nom du seuil",
```
souissimai, line 379: **LimitName was deleted in fr.json but not in en.json; it should be removed there as well

antoinebhs (approved):
Code OK
Rest of the duplication will be treated separately

souissimai (approved):

PR #3244, opened 2025-09-09
Save the label positions in NAD

EstherDarkish:


src/components/grid-layout/cards/diagrams/diagram-card.tsx
```diff
@@ -131,6 +131,21 @@ export const DiagramCard = forwardRef((props: DiagramCardProps, ref: Ref<HTMLDiv
...
+                );
+
+                updateDiagramPositions({
+                    ...diagram,
+                    positions: updatedPositions,
+                });
+            }
+        },
+        [diagram, updateDiagramPositions]
+    );
```
antoinebhs, line 148: Almost a duplicate from above method, can we improve this?

PR #3272, opened 2025-09-15
Fix Export CSV Button success status for SA,SE,CC at context change

sBouzols:
fix(SA,SE,CC): reset `exportSuccessful` states when context like study, node or rootNetwork change as well as resultType, computation type or tab change.

refactor(SE): move export CSV Button in a separeted component like for other computations


FranckLecuyer (changes requested):
Tests: 
no reset is done on sensi result column sort change
no reset is done when switching to another tab (tree or spreadsheet tab), and returning to the currently result tab

PR #3282, opened 2025-09-17
Add callback to populate unset regulation attributes with previous values

flomillot:
When the user enable onload we autofill the regulation attributes with the previous values of the equipment

### DIFF
--- a/src/components/dialogs/network-modifications/two-windings-transformer/tap-changer-pane/ratio-tap-changer-pane/ratio-tap-changer-pane.jsx
+++ b/src/components/dialogs/network-modifications/two-windings-transformer/tap-changer-pane/ratio-tap-changer-pane/ratio-tap-changer-pane.jsx
@@ -16,7 +16,7 @@ import {
     TARGET_DEADBAND,
     TARGET_V,
 } from 'components/utils/field-constants';
-import { useEffect, useMemo } from 'react';
+import { useCallback, useEffect, useMemo } from 'react';
 import { useFormContext, useWatch } from 'react-hook-form';
 import { useIntl } from 'react-intl';
 import { VoltageAdornment } from '../../../../dialog-utils';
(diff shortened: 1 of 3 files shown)


Mathieu-Deharbe (changes requested):
This is a bit different when "Type de réglage" is "distant". Then it is loaded but `poste` and `ouvrage` are not : 
image

And in this case validation is impossible : 
image

src/components/dialogs/network-modifications/two-windings-transformer/tap-changer-pane/ratio-tap-changer-pane/ratio-tap-changer-pane.jsx
```diff
@@ -90,6 +97,59 @@ const RatioTapChangerPane = ({
...
+                        console.error('Voltage level not found:', prevVl);
+                    }
+                }
+                if (curRatioTapChanger[EQUIPMENT] === null) {
+                    const prevEquipmentId = previousValues?.ratioTapChanger?.regulatingTerminalConnectableId;
+                    if (prevEquipmentId) {
+                        const prevEquipmentType = previousValues?.ratioTapChanger?.regulatingTerminalConnectableType;
+                        const newEquipment = {
+                            id: prevEquipmentId,
+                            label: prevEquipmentType,
```
Mathieu-Deharbe, line 142: I didn't spot any bug but I am a bit surprised that the label is `prevEquipmentType`.Is it wanted ?

Mathieu-Deharbe (approved):
Code and test OK. 
But you will have to obey to GitHub Actions about the missing dependency: 'voltageLevelOptions' !

src/components/dialogs/network-modifications/two-windings-transformer/tap-changer-pane/ratio-tap-changer-pane/ratio-tap-changer-pane.jsx
flomillot: Took it from here 
https://github.com/gridsuite/gridstudy-app/blob/985dc851f320ece0472bf761cda9b0bec00748fb/src/components/dialogs/regulating-terminal/regulating-terminal-form.tsx#L82-L88

PR #3286, opened 2025-09-17
Disable spreadsheet filter export on branch model

dbraquart:


src/components/spreadsheet-view/spreadsheet/spreadsheet-toolbar/save/save-spreadsheet-button.tsx
```diff
@@ -90,12 +90,7 @@ export default function SaveSpreadsheetButton({
                 id: SpreadsheetSaveOptionId.SAVE_FILTER,
                 label: 'spreadsheet/save/options/filter',
                 action: saveFilterDialogOpen.setTrue,
-                disabled:
-                    dataSize === 0 ||
-                    (tableDefinition.type === SpreadsheetEquipmentType.BRANCH
-                        ? !FILTER_EQUIPMENTS[EquipmentType.LINE] ||
-                          !FILTER_EQUIPMENTS[EquipmentType.TWO_WINDINGS_TRANSFORMER]
-                        : !FILTER_EQUIPMENTS[tableDefinition.type as unknown as EquipmentType]),
+                disabled: dataSize === 0 || !FILTER_EQUIPMENTS[tableDefinition.type as unknown as EquipmentType],
```
FranckLecuyer, line 93: 'dataSize' property is always undefined, so the 'Save as a filter' option is enabled even in case of no data (for example in 'Battery' tab when there is no battery in the network)

I don't understand how the cast 'tableDefinition.type as unknown as EquipmentType' function for branch, because the value BRANCH in SpreadsheetEquipmentType does not appear in EquipmentType ? but, indeed, the 'Save as a filter' option is disabled for branch ...

PR #3297, opened 2025-09-19
Fix the mouse cursors on NADs

EstherDarkish:


sBouzols (approved):
Code Review OK
Tests OK
`grabbing` cursor would be a must have during dragging event
Hard to understant as user that the voltageLevel is left clickable in edit mode.
Console warning check OK

PR #3306, opened 2025-09-23
Edit olg switch

Mathieu-Deharbe:


src/components/dialogs/limits/limits-pane-utils.ts
```diff
@@ -220,6 +225,21 @@ export const updateTemporaryLimits = (
     return updatedTemporaryLimits;
 };
 
+export const mapServerOLGtoOperationalLimitsGroupsFormInfos = (currentLimits: CurrentLimits[]) => {
```
EtienneLt, line 228: well for the naming, there is OLG and OperationalLimitsGroups in the method name

basseche:
In Visualization mode we can edit active limit sets but we should not according to specs!

basseche (changes requested):
In Visualization mode we can edit active limit sets but we should not according to specs!

EtienneLt (approved):
Test Ok Code Ok

PR #3313, opened 2025-09-24
Fix Tooltip display when resistance or reactance are purely 0

sBouzols:
fix(EquipmentPopover): do not test `equipmentInfo.r` or `equipmentInfo.x` before rendering the raw because if it's strict 0 (not decimal) then it doesn't render the table raw at all


souissimai (approved):
tested ok

PR #3690, opened 2026-01-23
Decouple log pagination state from computation type

thangqp:
## PR Summary
Complement of the quick fix in https://github.com/gridsuite/gridstudy-app/pull/3689
to avoid regressions when introducing new computations


thangqp:
Expired

PR #3825, opened 2026-03-20
Refactor column filters

flomillot:
## PR Summary
<!-- Briefly summarize the changes: what it was before, what it is after, and any important notes for reviewers. -->
Simplifying and securing column filters state updates :
  - Unified all column filters into state.tableFilters.columnsFilters as Record<string, Record<string, FilterConfig[]>>, keyed by TableType then by tab ID      
  - Stop updating the same things multiple times in different callbacks, data flow for results computation and spreadsheet is update backend database -> notification -> update redux state -> useEffect -> update hook state (no backend for logs)                                                                                                                                                                                                                    
  - Simplified pagination reset directly in the reducer when filters change using Immer direct mutation (.page = 0) instead of creating new objects via dispatch actions and callbacks                                                                                                                                                                                                                                                        
  - handleChangeComparator now consistently uses debouncedUpdateFilter for all branches instead of mixing direct and debounced calls, also fixed a bug by keeping the debounce when unmounting                                                                                             
  - Stop updating AG Grid filters when it's not necessary (not used anymore for computation result and already did in a use effect for spreadsheet)                   
  - Onle reset current tab and not all tabs when filters change
  - Global code cleaning


AbdelHedhili (commented):
just some empty lines to remove. Also prettier error to fix.

AbdelHedhili (approved):
test : ok 
code : ok for me but waiting on @ghazwarhili

src/components/custom-aggrid/custom-aggrid-filters/hooks/use-custom-aggrid-column-filter.ts
```diff
@@ -77,43 +97,40 @@ export const useCustomAggridFilter = (
...
+
+            // Data flow for logs table is: update redux state -> useEffect -> update hook state
+            if (type === TableType.Logs) {
+                dispatch(updateColumnFiltersAction(TableType.Logs, tab, updatedFilters));
+            }
+            // Data flow for spreadsheet / computation tables is: update backend database -> notification -> update redux state -> useEffect -> update hook state
+            else if (type === TableType.Spreadsheet) {
+                persistSpreadsheetColumnFilters(studyUuid, tab as UUID, updatedFilters, colDef, onError);
+            } else {
+                persistComputationColumnFilters(updatedFilters, colId, studyUuid, type, tab, onError);
```
ghazwarhili: studyUuid, tableType, filterSubType, colId) are the keys and updatedFilters is the data => Inverting the order makes the signature read more naturally and aligns with the existing persistSpreadsheetColumnFilters convention (studyUuid first)

src/components/results/common/column-filter/update-computation-columns-filters.ts
```diff
@@ -6,23 +6,20 @@
...
-export const updateComputationColumnsFilters = (
-    agGridApi?: GridApi,
+export const persistComputationColumnFilters = (
     filters?: FilterConfig[],
     colId?: string,
     studyUuid?: UUID,
     tableType?: TableType,
     filterSubType?: string,
-    onBeforePersist?: () => void
+    onError?: (error: unknown) => void
```
ghazwarhili: studyUuid, tableType, filterSubType, colId are all effectively required
    studyUuid: UUID,
    tableType: TableType,
    filterSubType: string,
    colId: string,
    filters?: FilterConfig[],
    onError?: (error: unknown) => void

src/components/custom-aggrid/custom-aggrid-filters/custom-aggrid-duration-filter.tsx
```diff
@@ -37,11 +37,11 @@ const styles = {
     },
 } as const satisfies MuiStyles;
 
-const CustomAggridDurationFilter: FunctionComponent<CustomAggridFilterParams> = ({ api, colId, filterParams }) => {
+const CustomAggridDurationFilter: FunctionComponent<CustomAggridFilterParams> = ({ colId, filterParams }) => {
```
ghazwarhili: the same remark to avoid FC to be aligned with our code

src/components/report-viewer/log-table.tsx
```diff
@@ -29,15 +29,17 @@ import { COMPUTING_AND_NETWORK_MODIFICATION_TYPE } from 'utils/report/report.con
...
+import { getColumnFiltersFromState } from '../../redux/selectors/filter-selectors';
 import { AGGRID_LOCALES } from '../../translations/not-intl/aggrid-locales';
 import CustomTablePagination from 'components/utils/custom-table-pagination';
 import { reportStyles } from './report.styles';
 import { useLogsPagination } from './use-logs-pagination';
 import { useStableComputedArray } from '../../hooks/use-stable-computed-array';
+import { updateAgGridFilters } from '../custom-aggrid/custom-aggrid-filters/utils/aggrid-filters-utils';
 
-const getColumnFilterValue = (array: FilterConfig[] | null, columnName: string): any => {
+const getColumnFilterValue = (array: FilterConfig[] | null | undefined, columnName: string): any => {
```
ghazwarhili: why this changes to add undefined ?

src/components/custom-aggrid/custom-aggrid-filters/hooks/use-custom-aggrid-column-filter.ts
flomillot: I kept the original order  to keep the history,  but you're right it's better to have of the same order everywhere, I will update both

src/redux/reducer.ts
```diff
@@ -1658,10 +1640,23 @@ export const reducer = createReducer(initialState, (builder) => {
...
+            case TableType.SensitivityAnalysis:
+                state[SENSITIVITY_ANALYSIS_PAGINATION_STORE_FIELD][filterSubType as SensitivityAnalysisTab].page = 0;
+                break;
+            case TableType.ShortcircuitAnalysis:
+                state[SHORTCIRCUIT_ANALYSIS_PAGINATION_STORE_FIELD][filterSubType as ShortcircuitAnalysisTab].page = 0;
+                break;
+            case TableType.PccMin:
+                state[PCCMIN_ANALYSIS_PAGINATION_STORE_FIELD][filterSubType as PccminTab].page = 0;
+                break;
+        }
```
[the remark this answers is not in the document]
ghazwarhili: @etiennehomer An opinion?

src/components/results/common/column-filter/persist-computation-column-filters.ts
```diff
@@ -1,28 +1,22 @@
...
-    filters?: FilterConfig[],
-    colId?: string,
-    studyUuid?: UUID,
-    tableType?: TableType,
-    filterSubType?: string,
-    onBeforePersist?: () => void
+export const persistComputationColumnFilters = (
+    studyUuid: UUID,
+    tableType: TableType,
+    tableId: string,
```
ghazwarhili: tableId: UUID,

src/components/custom-aggrid/custom-aggrid-filters/hooks/use-custom-aggrid-column-filter.ts
```diff
@@ -77,43 +97,40 @@ export const useCustomAggridFilter = (
...
-            dispatchFilters(updatedFilters);
+
+            const onError = (error: unknown) => snackWithFallback(snackError, error);
+
+            // Data flow for logs table is: update redux state -> useEffect -> update hook state
+            if (type === TableType.Logs) {
+                dispatch(updateColumnFiltersAction(TableType.Logs, tab, updatedFilters));
+            }
+            // Data flow for spreadsheet / computation tables is: update backend database -> notification -> update redux state -> useEffect -> update hook state
+            else if (type === TableType.Spreadsheet) {
```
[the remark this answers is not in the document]
flomillot: Tbh I prefer the current code than what you propose. If I remove the else it's the same code as you but with less code in the useEffect.
But reading the code again, you're right that the 2 functions are very close, and I can mutualize a bit.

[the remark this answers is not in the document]
flomillot: I tried but considering that I can't remove persistSpreadsheetColumnFilters and that the column DTO for spreadsheet and for computation results are not exactly the same (especially the ID), I think it's better like it is currently.

src/components/custom-aggrid/custom-aggrid-filters/custom-aggrid-duration-filter.tsx
etiennehomer: At a moment, we said on gridsuite that FunctionComponent are recommended. So there are plenty of it in the code.
To me, it's clearly ok to use it in gridstudy. 
For the new front we gonna start soon, maybe we'll avoid it (not really recommenced by react now)

ghazwarhili (approved):
test => OK
code => OK

src/components/results/common/column-filter/persist-computation-column-filters.ts
flomillot: No its' not

image
image
image
image

flomillot: I will rename it

PR #3856, opened 2026-03-31
Moving the Modification by filter assignment form gridstudy to commons-ui

benrejebmoh:


dbraquart (changes requested):
Code: 
remove moved field constants and use FieldConstants.*
remove moved translations

Tests:
- regression: when I edit an existing "VL" modification, and turn the type into "Site", I dont have the confirmation pop-up I can see on main branch

dbraquart (approved):

PR #3869, opened 2026-04-09
Fix : Remove useless snackbar on current tab before unload

ayolab:


sBouzols:
as discussed we should fix the case precisely to keep other usages:
then change signature and implementation of `cleanClipboard` functions like this : 
```typescript
const cleanClipboard = useCallback(
        (showSnackInfoInCurrentTab = true, showSnackInfoInOtherTabs = true, snackInfoMessage?: string) => {
```
and fix the only concerned case here : 
```typescript
    window.addEventListener('beforeunload', () => {
            cleanClipboard(false, true, 'copiedNodeInvalidationMsgFromStudyClosure');
        });
```

SlimaneAmar (approved):

PR #3872, opened 2026-04-13
Fix spreadsheet showing 'No data' on rapid node switch

flomillot:
## Summary
- ~~**Race condition fix**: Added a version counter in `useFetchEquipment` to discard stale fetch responses when the selected node changes before a previous fetch completes.~~ moved in https://github.com/gridsuite/gridstudy-app/pull/3879
- **isFetching preservation**: `REMOVE_NODE_DATA` no longer resets `isFetching` to `false`, preventing the component from showing "No data" while a fetch is still in flight.
- **isInitialized flag**: Added `isInitialized` to `SpreadsheetEquipmentsByNodes` to distinguish "never loaded" types from "temporarily empty" ones. This prevents `useSpreadsheetEquipments` from skipping fetches for types whose data was cleared during a rapid node switch.


ghazwarhili:
The fix did not resolve the bug
Capture d’écran du 2026-04-14 13-31-37

Capture d’écran du 2026-04-14 13-30-42

src/components/spreadsheet-view/hooks/use-fetch-equipment.ts
```diff
@@ -22,9 +22,16 @@ export function useFetchEquipment() {
...
 
+    // Track the latest fetch version per equipment type to ignore stale responses
+    const fetchVersionRef = useRef<Map<SpreadsheetEquipmentType, number>>(new Map());
+
     const fetchNodesEquipmentData = useCallback(
         (type: SpreadsheetEquipmentType, nodesIds: Set<UUID>) => {
             if (studyUuid && currentRootNetworkUuid) {
+                // Increment version for this type — only the latest version will dispatch results
+                const currentVersion = (fetchVersionRef.current.get(type) ?? 0) + 1;
+                fetchVersionRef.current.set(type, currentVersion);
```
ghazwarhili: The kind of solution based on versioning is generally not reliable, and it seems to me that it was AI that suggested it to you, but you should be cautious !

ghazwarhili: @etiennehomer what do you think ?

flomillot:
> The fix did not resolve the bug Capture d’écran du 2026-04-14 13-31-37
> 
> Capture d’écran du 2026-04-14 13-30-42

As discussed there is just no data for three winding transformers. Have you got another example ?

src/components/spreadsheet-view/hooks/use-fetch-equipment.ts
flomillot: Can you explain why it's not reliable ? You have an example in mind ? Because for me it's the best / easiest way to cover every cases.

etiennehomer: To me, the solution looks ok. But it is weird to have it in only one treatment in the front apps. Let's wait for a more global decision in the team to manage such problems

flomillot: Extracted in https://github.com/gridsuite/gridstudy-app/pull/3879

etiennehomer (approved):

ghazwarhili (approved):
code => OK 
test => OK

PR #3876, opened 2026-04-14
exclude inexistant filter from fetch

ghazwarhili:
## PR Summary
exclude inexistant filter from fetch


src/redux/actions.ts
```diff
@@ -961,16 +961,19 @@ export type MarkNotFoundGlobalFiltersAsDeletedAction = Readonly<
...
     globalFilters: GlobalFilter[];
     tableId: string;
+    tableType: string;
 };
 
 export function markNotFoundGlobalFiltersAsDeleted(
     globalFilters: GlobalFilter[],
-    tableId: string
+    tableId: string,
+    tableType: string
```
flomillot: As say rabbit, it's not a string. The other comment on the index is also accurate.

src/redux/reducer.ts
```diff
@@ -1702,12 +1702,13 @@ export const reducer = createReducer(initialState, (builder) => {
...
         (state, action: MarkNotFoundGlobalFiltersAsDeletedAction) => {
-            const { globalFilters, tableId } = action;
+            const { globalFilters, tableId, tableType } = action;
             const ids = new Set(globalFilters.map((f) => f.id));
             state.globalFilterOptions.forEach((globalFilter) => {
                 if (ids.has(globalFilter.id)) globalFilter.deleted = true;
             });
-            const tableState = state.tableFilters.globalFilters[tableId];
+            const index = tableId ?? tableType;
+            const tableState = state.tableFilters.globalFilters[index];
```
flomillot: ?

flomillot (changes requested):

src/redux/actions.ts
ghazwarhili: TableType

src/redux/reducer.ts
ghazwarhili: fixed

flomillot (approved):

PR #3878, opened 2026-04-15
fix selected report viwer

ghazwarhili:
## PR Summary
fix selected report viwer


flomillot (changes requested):
~~I don't understand how this code can change something... You return exactly the same value, no ?~~
~~The issue is not in GridStudy IMO.~~ I did not read correctly mb
However, did you reproduce exactly the steps described in the ticket? It's not fixed for me. 

Image

Image

ghazwarhili:
https://github.com/gridsuite/gridstudy-app/pull/3878#pullrequestreview-4114862317
You’re right — it’s not the root cause of the bug.

PR #3881, opened 2026-04-15
Fix segment insertion order

basseche:
## PR Summary

Keep same order for line segments insertion


src/components/dialogs/line-types-catalog/line-type-segment-form.tsx
```diff
@@ -230,10 +230,11 @@ export const LineTypeSegmentForm: FunctionComponent<LineTypeSegmentFormProps> =
         }
         arrayRef.current?.replaceItems([]);
         const updateSegmentsLimits = async () => {
-            const promises = editData.map((segment) => appendSegmentLimits(segment));
-
             try {
-                await Promise.all(promises);
```
flomillot, line 236: Keep the promise all which is important in terms of performance. You can reorganize the order.

flomillot (changes requested):

src/components/dialogs/line-types-catalog/line-type-segment-form.tsx
basseche: How ?

flomillot (approved):
Code looks OK

PR #3885, opened 2026-04-16
Limit composites nested imbrication

Meklo:


Meklo:
Moved to https://github.com/gridsuite/commons-ui/pull/1103

PR #979, opened 2023-01-03
Load creation dialog, migration to react hook form and yup

archendel:

PR #3897, opened 2026-04-17
report use uuid, parentOrder as key instead of having an extra v7 ordered uuid - #979

jonenst:
Prompted opus4.6
"look at report-server/ the diffs from main I changed the PK from uuid to (uuid, order). This changes the API. adapt the code in gridstudy-app to use the new APIs "


jonenst:
Not planned for now, can come back to this if needed

PR #3898, opened 2026-04-17
prevent app refetch on token silent renew

ghazwarhili:
## PR Summary
Each oidc silent renew dispatched a user action carrying a fresh User instance (new access_token, id_token, expires_at) while the identity (profile.sub) was unchanged. 

Two issues cascaded from this:

1. The reducer replaced state.user unconditionally, changing its reference on every renew. All components selecting state.user re-rendered, and effects depending on user re-ran, triggering redundant fetches (network visualization parameters, spreadsheet config, optional services, etc.).

1. The WebSocket URLs generated by useNotificationsUrlGenerator embedded the token as a query parameter, so token rotation changed the URL objects. NotificationsProvider then tore down and recreated all WebSockets, and each new onopen broadcast caused listeners (useNodeAliases, network elements, etc.) to resync their data.

Changes:

1. Split state.user: state.user keeps a stable reference across silent renews (only changes when profile.sub changes), and state.userToken holds the rotating tokens used by getUserToken and the WS layer.

2. useNotificationsUrlGenerator now emits token-free URLs based on an isAuthenticated boolean, so the urls object stays stable when the token rotates.

1. The fresh token is appended at WebSocket connection/reconnection time via getUserToken(), preserving reconnection behavior for real disconnects.

needs https://github.com/gridsuite/commons-ui/pull/1102


src/services/utils.ts
```diff
@@ -91,14 +90,6 @@ export const getQueryParamsList = (params: string[] | number[] | null | undefine
     return '';
 };
 
-export function getUrlWithToken(baseUrl: string) {
```
flomillot, line 94: Maybe keep it with (baseUrl, tokenId) props

src/redux/reducer.ts
```diff
@@ -1019,7 +1020,26 @@ export const reducer = createReducer(initialState, (builder) => {
     });
 
     builder.addCase(USER, (state, action: UserAction) => {
-        state.user = action.user;
+        const newUser = action.user;
```
flomillot: So that mean state.user will contain a deprecated token after the first one ?
IMO it's a bit weird, it's not possible to identify and fix the problematic useEffect instead ?
I'm not a fan of duplicating the token id in the store to avoid bad triggers...

flomillot (commented):
I would rather keep the previous code of `useNotificationsUrlGenerator` and improving it a bit, by injecting tokenId as prop to getUrlWithToken instead of fetching it again as it's alredy present in the component

src/services/utils.ts
ghazwarhili: reverted and tokenId added to getUrlWithToken props

src/redux/user-store.ts
```diff
@@ -41,5 +46,5 @@ export function setUserStore(store: UserStore): void {
 
 //TODO use the one from commons-ui instead when exported in next version
 export function getUserToken() {
-    return userStore?.getState().user?.id_token ?? undefined;
+    return userStore?.getState().userToken?.id_token ?? undefined;
```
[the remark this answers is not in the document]
ghazwarhili: code not needed anymore

src/redux/reducer.ts
ghazwarhili: Yes fair point, with my fix the id_token in state.user, I agree it's a bit weird
identifying the problematic useEffect : the thing is, it's not really one specific effect we can fix. Any useSelector(state => state.user) or selector returning a sub-object of user triggers a re-render on every renew because state.user's reference changes. We have this pattern all over the place (App, AppTopBar, plus commons-ui  like TopBar, AnnouncementNotification). It's not one place to fix, it's a pattern that's everywhere.

duplicate the token to have a dedicated source for websockets. It's uglier, but localized and low-risk

the perfect solution for later is clean up all the useSelector(state.user) calls to use more precise primitive selectors (state.user?.profile?.sub, etc.), and adapt commons-ui components to accept primitives instead of the full user object. That's the clean refactor, but it's bigger and touches commons-ui

src/services/utils.ts
```diff
@@ -91,11 +89,11 @@ export const getQueryParamsList = (params: string[] | number[] | null | undefine
...
+export function getUrlWithToken(baseUrl: string, tokenId: string) {
```
flomillot: Not used anymore, you can delete it

flomillot (changes requested):

src/services/utils.ts
ghazwarhili: should be deleted 
done

src/redux/reducer.ts
ghazwarhili: outdated code

src/services/utils.ts
ghazwarhili: outdated code

flomillot (approved):

src/components/app.jsx
```diff
@@ -89,7 +89,7 @@ const App = () => {
             .catch(() => cleanupStaleStudyData());
     }, []);
 
-    const user = useSelector((state) => state.user);
+    const userProfile = useSelector((state) => state.user?.profile ?? null, shallowEqual);
```
flomillot, line 92: @coderabbitai is shallowEqual the appropriate equal method here ?

sBouzols (approved):
Code Review OK
Tests OK
Console warning check OK

PR #3908, opened 2026-04-22
Fix double update column call when updating column

flomillot:
## Summary
- Drop the filter-clearing logic from `ColumnCreationDialog`: omitting `columnFilterInfos` on update is now what triggers the backend to reset the filter. **It also removes an unnecessary HTTP call and fix a potential bug reseting the previous column informations.**
- Rename `persistSpreadsheetColumnFilters` / `persistComputationColumnFilters` to their singular form and pass a single `FilterConfig` instead of the full filters array.
- Make `mapColDefToDto` emit an undefined `columnFilterInfos` when no filter is provided.

### DIFF
--- a/src/components/results/common/column-filter/persist-computation-column-filter.ts
+++ b/src/components/results/common/column-filter/persist-computation-column-filter.ts
@@ -13,17 +13,17 @@ export const persistComputationColumnFilter = (
     computationType: TableType,
     computationSubType: string,
     colId: string,
-    filter: FilterConfig | undefined,
+    colFilter: FilterConfig | undefined,
     onError: (error: unknown) => void
 ) => {
     const columnFilterInfos = {
         columnId: colId,
-        columnFilterInfos: filter
+        columnFilterInfos: colFilter
             ? {
-                  filterDataType: filter?.dataType,
-                  filterType: filter?.type,
-                  filterValue: JSON.stringify(filter?.value),
-                  filterTolerance: filter?.tolerance,
+                  filterDataType: colFilter?.dataType,
+                  filterType: colFilter?.type,
+                  filterValue: JSON.stringify(colFilter?.value),
+                  filterTolerance: colFilter?.tolerance,
               }
             : null,
     };
--- a/src/components/results/common/column-filter/persist-computation-column-filter.ts
+++ b/src/components/results/common/column-filter/persist-computation-column-filter.ts
@@ -8,15 +8,14 @@ import { UUID } from 'node:crypto';
 import { FilterConfig, TableType } from '../../../../types/custom-aggrid-types';
 import { updateComputationResultFiltersColumn } from '../../../../services/study/study-config';
 
-export const persistComputationColumnFilters = (
+export const persistComputationColumnFilter = (
     studyUuid: UUID,
     computationType: TableType,
     computationSubType: string,
     colId: string,
-    filters: FilterConfig[],
+    filter: FilterConfig | undefined,
     onError: (error: unknown) => void
 ) => {
-    const filter = filters.find((f) => f.column === colId);
     const columnFilterInfos = {
         columnId: colId,
         columnFilterInfos: filter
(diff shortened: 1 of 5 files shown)


src/components/results/common/column-filter/persist-computation-column-filter.ts
```diff
@@ -8,15 +8,14 @@ import { UUID } from 'node:crypto';
...
 import { updateComputationResultFiltersColumn } from '../../../../services/study/study-config';
 
-export const persistComputationColumnFilters = (
+export const persistComputationColumnFilter = (
     studyUuid: UUID,
     computationType: TableType,
     computationSubType: string,
     colId: string,
-    filters: FilterConfig[],
+    filter: FilterConfig | undefined,
```
ghazwarhili: rename to colFilter ?

and if you changed the type here, you need to update the rest so that everything stays consistent

columnFilterInfos: filter
            ? {
                  filterDataType: filter?.dataType,
                  filterType: filter?.type,
                  filterValue: JSON.stringify(filter?.value),
                  filterTolerance: filter?.tolerance,
              }
            : undefined,

Meklo (approved):

src/components/spreadsheet-view/columns/column-creation-dialog.tsx
```diff
@@ -294,22 +280,15 @@ export default function ColumnCreationDialog({
...
+            // We don't include the column filter, so it will be removed when we update the column.
             const formattedParams = {
                 ...newParams,
                 dependencies: newParams.dependencies?.length ? JSON.stringify(newParams.dependencies) : undefined,
             };
 
             const updateOrCreateColumn =
-                isUpdate && columnDefinition
+                existingColumn && columnDefinition
                     ? updateSpreadsheetColumn(studyUuid, spreadsheetConfigUuid, columnDefinition.uuid, formattedParams)
```
ghazwarhili: I know it is not your code
Missing await for an async function call

src/components/spreadsheet-view/columns/utils/persist-spreadsheet-column-filter.ts
```diff
@@ -10,18 +10,17 @@ import { ColumnDefinition } from '../../types/spreadsheet.type';
 import { updateSpreadsheetColumn } from 'services/study/study-config';
 import { mapColDefToDto } from '../../add-spreadsheet/dialogs/add-spreadsheet-utils';
 
-export const persistSpreadsheetColumnFilters = (
+export const persistSpreadsheetColumnFilter = (
     studyUuid: UUID,
     tabUuid: UUID,
     colDef: ColumnDefinition | undefined,
-    filters: FilterConfig[],
+    filter: FilterConfig | undefined,
```
ghazwarhili: rename to colFilter ?

src/components/results/common/column-filter/persist-computation-column-filter.ts
flomillot: I didn't change it on purpose to preserve the Git history, and null or undefined will be the same in the backend
For the name, I agree if it was a new development, but as it's existing code, I prefer let it like that.

src/components/spreadsheet-view/columns/utils/persist-spreadsheet-column-filter.ts
flomillot: Well okay for this one.

src/components/spreadsheet-view/columns/column-creation-dialog.tsx
flomillot: Be careful, await will change the behavior of the code. I added the missing catch instead.

flomillot: I also fixed some incoherences with the code.

src/components/results/common/column-filter/persist-computation-column-filter.ts
flomillot: Finally I will update it to be coherent with persistSpreadsheetColumnFilters

ghazwarhili (approved):
code => OK

PR #3909, opened 2026-04-22
add extended equipment type for equipment deletion

EtienneLt:
## PR Summary
add ExtendedEquipmentType when fetching element ids for deletion


Mathieu-Deharbe (changes requested):
When we reload a hvdc line deletion the equipment type is displayed as HVDC instead of Ligne HVDC (VSC) or Ligne HVDC (LCC) (which are in the list).

Image

Mathieu-Deharbe:
When I reload my VSC et LCC deletion the lists are identical :

rechargement

because the `equipmentType` is HVDC_LINE in both case : 

image

PR #3919, opened 2026-04-24
Fix filtering on global filters when only invalid filters are left

basseche:
## PR Summary

- When we have 2 generic filters.
- One of them becomes invalid (deleted)
- When we remove the remaining valid filter we should have all equipments.


flomillot (approved):

ghazwarhili (approved):
code => OK
test => OK

PR #3926, opened 2026-04-29
fix(Estim): active weights parameters must be positive

dbraquart:
## PR Summary

With powsybl 2026.0.0, active weights parameters must be positive (rather than negative before).


KoloMenek (approved):

KoloMenek:
Tested locally, it works and is a valid change after changes to estim parameters, but even with this dev, if a study has state estimation parameters from previous PowSyBl version, this dev won't fix them, the user needs to reset parameters of the study to take into account the new weights.

dbraquart:
exact: this PR fixes only edition issue with grid-study. A database update has to be done to fix existing data.

PR #3932, opened 2026-05-05
rename composite modif

souissimai:


Mathieu-Deharbe (commented):
This is not a bug and probably not a big deal but it should be checked with the PO : the renaming doesn't unbuild the nodes. But the composite name is written in the logs. Therefore when a composite is renamed, the logs don't display the right name anymore : 

Image

package-lock.json
```diff
@@ -173,6 +173,7 @@
             "integrity": "sha512-2BCOP7TN8M+gVDj7/ht3hsaO/B/n5oDbiAyyvnRlNOs+u1o+JWNYTQrmpuNp1/Wq2gcFrI01JAW+paEKDMx/CA==",
             "dev": true,
             "license": "MIT",
+            "peer": true,
```
Mathieu-Deharbe: Don't merge this.

src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
```diff
@@ -805,6 +806,27 @@ const NetworkModificationNodeEditor = () => {
...
+    );
+    const handleCellEdit = useCallback(
+        async (modification: ComposedModificationMetadata, newName?: string) => {
+            console.log('rename →', modification, 'newName:', newName);
+
+            if (!newName || newName.trim() === '') {
+                return;
+            }
+
+            await updateModification(modification as ComposedModificationMetadata, newName.trim());
```
Mathieu-Deharbe: Agreed with coderabbit : 
```suggestion
            await updateModification(modification, newName.trim());
```
`modification` is already typed as a `ComposedModificationMetadata`.

src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
```diff
@@ -805,6 +806,27 @@ const NetworkModificationNodeEditor = () => {
...
+                name: newName,
+                type: modif?.type,
+            }).finally(() => {});
+        },
+        [studyUuid, currentNode?.id]
+    );
+    const handleCellEdit = useCallback(
+        async (modification: ComposedModificationMetadata, newName?: string) => {
+            console.log('rename →', modification, 'newName:', newName);
+
```
Mathieu-Deharbe: Forgotten ?
```suggestion
```

src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
```diff
@@ -802,6 +803,27 @@ const NetworkModificationNodeEditor = () => {
...
 
+    const updateModification = useCallback(
+        async (modif: ComposedModificationMetadata, newName: string) => {
+            return setModificationMetadata(studyUuid, currentNode?.id, modif.uuid, {
+                name: newName,
+                type: modif?.type,
+            }).finally(() => {});
+        },
+        [studyUuid, currentNode?.id]
+    );
```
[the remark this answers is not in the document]
Mathieu-Deharbe: I quite agree : I am not sure that the snackbar is needed, but the finally definitely is not and should be removed.

Mathieu-Deharbe (commented):
I didn't try it myself but I think almost all this code can be put directly into commons-ui (setModificationMetadata would need to move too).

It will be done soon because we will have to make it work from gridexplore.

So even if this is not directly asked by the JIRA ticket it might be done here and save some time later (plus it would be the first edition available from gridexplore!)

src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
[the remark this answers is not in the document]
souissimai: done

src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
souissimai: ok

souissimai:
> I didn't try it myself but I think almost all this code can be put directly into commons-ui (setModificationMetadata would need to move too).
> 
> It will be done soon because we will have to make it work from gridexplore.
> 
> So even if this is not directly asked by the JIRA ticket it might be done here and save some time later (plus it would be the first edition available from gridexplore!)

I’d prefer to handle this separately for now. It may still contain some GridStudy-specific behavior, and if we later need the same functionality in GridExplore, we can adapt and generalize it at that point.

souissimai:
> This is not a bug and probably not a big deal but it should be checked with the PO : the renaming doesn't unbuild the nodes. But the composite name is written in the logs. Therefore when a composite is renamed, the logs don't display the right name anymore :
> Image

==> fixed , unbuild node in this case , discussed with PO

Mathieu-Deharbe (approved):

src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
```diff
@@ -817,12 +817,33 @@ const NetworkModificationNodeEditor = () => {
     );
     const handleCellEdit = useCallback(
         async (modification: ComposedModificationMetadata, newName?: string) => {
-
             if (!newName || newName.trim() === '') {
                 return;
             }
+            const trimmed = newName.trim();
+
+            // Optimistic update immédiat
```
Mathieu-Deharbe:
```suggestion
            // Optimistic immediate update
```

Mathieu-Deharbe (changes requested):

src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
```diff
@@ -805,6 +806,47 @@ const NetworkModificationNodeEditor = () => {
...
+                    } catch {
+                        return m;
+                    }
+                })
+            );
+
+            try {
+                await updateModification(modification, trimmed);
+            } catch {
+                // Rollback en cas d'erreur
```
Mathieu-Deharbe:
```suggestion
                // Rollback in case of an error
```

Mathieu-Deharbe (changes requested):

Mathieu-Deharbe (approved):

PR #3942, opened 2026-05-11
init yup config before creating form schemas

khouadrired:
remove duplicate translations to use the ones in commons-ui


src/components/app-wrapper.jsx
```diff
@@ -6,6 +6,8 @@
  */
 
 import { useMemo } from 'react';
+// need to configuer yup before creating the form schemas
```
klesaulnier:
```suggestion
// need to configure yup before creating any form schema
```
It could be nice to add a little explanation of why we need to do so ?
I feel like we could forget why we did that in the future otherwise

thangqp: IMO, we should mention at both sides, usage context and inside the imported module if we consider yupConfig.ts as a provided object exposed by commons-ui.

src/components/app-wrapper.jsx
```diff
@@ -6,6 +6,10 @@
  */
 
 import { useMemo } from 'react';
+// need to configure yup before creating any form schema as we have an issue where the form schema
+// is created before configuring the translation in yup, in that case it will take the default
+// beahaviour of yup
```
klesaulnier:
```suggestion
// behaviour of yup
```

klesaulnier (approved):

thangqp (approved):

src/components/app-wrapper.jsx
thangqp: @khouadrired commit direct?

PR #3946, opened 2026-05-15
[FIX] Variation id in sensi result is not enriched with element names

thangqp:
## PR Summary
This is only a proposal solution which enrich element names for variation id after fetching the result DTO.
**However, there is not any solution at frontend to export sensi result with enriched variation id.**


thangqp:
Expired

PR #3958, opened 2026-05-25
import modification dialog : composite selection

souissimai:


src/components/dialogs/import-modification-dialog.tsx
```diff
@@ -9,42 +9,55 @@ import { FormattedMessage, useIntl } from 'react-intl';
...
+    useFormContext,
+    useWatch,
+} from 'react-hook-form';
 import { yupResolver } from '@hookform/resolvers/yup';
 import yup from 'components/utils/yup-config';
-import { ModificationDialog } from './commons/modificationDialog';
-import { ACTION, COMPOSITE_NAMES, SELECTED_MODIFICATIONS } from 'components/utils/field-constants';
-import GridItem from './commons/grid-item';
+import { ACTION, SELECTED_MODIFICATIONS } from 'components/utils/field-constants';
 import { UUID } from 'node:crypto';
```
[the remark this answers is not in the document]
Mathieu-Deharbe: Not your fault, but rabbit is right so while we are here : 
```suggestion
import type { UUID } from 'node:crypto';
```

Mathieu-Deharbe (changes requested):
The jira ticket states that the size of the window shouldn't change between the 2 steps (La taille de la fenêtre ne change pas entre les deux étapes.)

This is not the case here : 
Step 1 : 

Image

Step 2 : 

Image

src/translations/fr.json
```diff
@@ -439,12 +439,14 @@
     "CreateCompositeModificationLabel": "Créer une nouvelle modification composite",
     "UpdateCompositeModificationLabel": "Remplacer une modification composite existante",
     "importComposites.title": "Importer des modifications composites",
-    "importComposites.label": "Organisation des modifications : ",
+    "importComposites.organizationLabel": "Organisation des modifications : ",
     "importComposites.action.split": "Unitaires seulement",
```
Mathieu-Deharbe:
```suggestion
    "importComposites.action.split": "Identiques",
```
cf JIRA : 

> On renomme le radioButton "Unitaires seulement" en Same / Identiques

src/translations/en.json
```diff
@@ -440,12 +440,14 @@
     "CreateCompositeModificationLabel": "Create a new composite modification",
     "UpdateCompositeModificationLabel": "Replace an existing composite modification",
     "importComposites.title": "Import composite modifications",
-    "importComposites.label": "Modifications organisation: ",
+    "importComposites.organizationLabel": "Modifications organisation: ",
     "importComposites.action.split": "Only unitary",
```
Mathieu-Deharbe:
```suggestion
    "importComposites.action.split": "Same",
```
cf JIRA : 

> On renomme le radioButton "Unitaires seulement" en Same / Identiques

Mathieu-Deharbe (commented):
I find it a bit strange that "next" is inside a blue box : 

Image

While the "previous" button is not : 

Image

In the mockup everything is into boxes including "cancel".
But I have no idea if this is important. To be checked with the ux designer.

Mathieu-Deharbe (commented):
According to JIRA the step text is supposed to be under the numbers : 

Image

For now it is not

Image

(I don't know if this is important, to be checked with the ux designer)

Mathieu-Deharbe (commented):
I think the "shared" checkbox should be aligned vertically with the text box : 

Image

src/components/dialogs/import-modification-dialog.tsx
```diff
@@ -53,68 +66,110 @@ interface ImportModificationDialogProps {
...
+    const isShared: boolean = useWatch({
+        name: `${SELECTED_MODIFICATIONS}.${rowIndex}.isShared`,
+    });
+
+    return (
+        <FormControlLabel
+            control={<Checkbox size="small" disabled checked={!!isShared} sx={{ p: 0, ml: 1 }} />}
+            label={
+                <Typography variant="body2" color="text.disabled" sx={{ ml: 0.5 }}>
+                    <FormattedMessage id="importComposites.shared" defaultMessage="Partagé" />
```
Mathieu-Deharbe: `importComposites.shared` desn't exist in the en.jons and fr.json files.

(plus defaultMessage shouldn't include an accent and should probably be removed)

src/components/dialogs/import-modification-dialog.tsx
```diff
@@ -53,68 +66,110 @@ interface ImportModificationDialogProps {
...
+}
+
+function SharedCell({ rowIndex }: Readonly<SharedCellProps>) {
+    const isShared: boolean = useWatch({
+        name: `${SELECTED_MODIFICATIONS}.${rowIndex}.isShared`,
+    });
+
+    return (
+        <FormControlLabel
+            control={<Checkbox size="small" disabled checked={!!isShared} sx={{ p: 0, ml: 1 }} />}
```
Mathieu-Deharbe: Why not just : ?
```suggestion
            control={<Checkbox size="small" disabled checked={isShared} sx={{ p: 0, ml: 1 }} />}
```

src/components/dialogs/import-modification-dialog.tsx
```diff
@@ -127,155 +182,271 @@ const ImportModificationDialog: ({ open, onClose }: Readonly<ImportModificationD
...
-    const [isDeveloperMode] = useParameterState(PARAM_DEVELOPER_MODE);
+
+    // useFieldArray — consumed by DndTable
+    const useFieldArrayOutput = useFieldArray({
+        control,
+        name: SELECTED_MODIFICATIONS,
+    }) as unknown as UseFieldArrayReturn;
+
+    // Column definitions
+    // Shared column — same in both modes, shows disabled checkbox only when isShared=true
```
Mathieu-Deharbe: I don't understand that comment : it is always disabled (for now) and it has nothing to do with being isShared=true.

src/components/dialogs/import-modification-dialog.tsx
```diff
@@ -53,68 +66,110 @@ interface ImportModificationDialogProps {
...
+    return (
+        <Box sx={{ display: 'flex', flexDirection: 'column', width: '100%' }}>
+            <Controller
+                name={`${SELECTED_MODIFICATIONS}.${rowIndex}.name`}
+                control={control}
+                render={({ field, fieldState }) => (
+                    <TextField {...field} size="small" fullWidth error={!!fieldState.error} />
+                )}
+            />
+            <Typography variant="caption" color="text.disabled" sx={{ px: 1, mt: 0.25 }}>
```
Mathieu-Deharbe: This would preserve MUI uniformity (we already use it for the same purpose so I don't think we should personalise a Typography) : 
```suggestion
            <FormHelperText sx={{ px: 1, mt: 0.25 }}>
```

src/translations/en.json
```diff
@@ -1544,5 +1546,8 @@
     "deactivateModification": "Deactivate network modification",
     "deselectModification": "Deselect modification",
     "selectModification": "Select modification",
-    "moveModification": "Move modification"
+    "moveModification": "Move modification",
+    "button.cancel": "Cancel",
```
Mathieu-Deharbe: I think you should use this label from commons-ui : 
```
export const commonButtonFr = {
    cancel: 'Annuler',
    validate: 'Valider',
    'button.changeType': 'Modifier',
    'button.changeOperator': 'Modifier',
    'button.restore': 'Restaurer',
};
```

src/components/dialogs/import-modification-dialog.tsx
```diff
@@ -127,155 +182,271 @@ const ImportModificationDialog: ({ open, onClose }: Readonly<ImportModificationD
...
+                    <Box>
+                        {activeStep === STEP_ORGANIZATION && (
+                            <Button onClick={handlePrevious}>
+                                <FormattedMessage id="button.previous" />
+                            </Button>
+                        )}
+                    </Box>
+                    <Box sx={{ display: 'flex', gap: 1 }}>
+                        <Button onClick={handleClose}>
+                            <FormattedMessage id="button.cancel" />
```
Mathieu-Deharbe: This is in commons-ui : 
```suggestion
                            <FormattedMessage id="cancel" />
```

src/components/dialogs/import-modification-dialog.tsx
```diff
@@ -9,42 +9,55 @@ import { FormattedMessage, useIntl } from 'react-intl';
...
+import { ACTION, SELECTED_MODIFICATIONS } from 'components/utils/field-constants';
 import { UUID } from 'node:crypto';
 import { useParameterState } from './parameters/use-parameters-state';
-import FormControlLabel from '@mui/material/FormControlLabel';
 
-/**
- * Dialog to select composite network modifications and append them to the current node.
- * - SPLIT mode: extracts and inserts individual modifications contained in the composites.
- * - INSERT mode: inserts the composite modifications as-is; a name per composite is required.
- *
```
Mathieu-Deharbe, line 38: Why do you remove those comments ?

src/translations/en.json
souissimai: ok

souissimai:
> the step text is supposed to be under the numbers :

right! changed it

src/components/dialogs/import-modification-dialog.tsx
souissimai: yes

src/translations/en.json
souissimai: yes

src/components/dialogs/import-modification-dialog.tsx
souissimai: ok

souissimai:
> I find it a bit strange that "next" is inside a blue box :
> Image
> 
> While the "previous" button is not :
> Image
> 
> In the mockup everything is into boxes including "cancel". But I have no idea if this is important. To be checked with the ux designer.

I just followed what was in the mockup. I will recheck with the UX designer.

src/components/dialogs/import-modification-dialog.tsx
souissimai: it's outdated , updated it and added it

gabriellevy (changes requested):
I tried it without the developper mode and if I select a composite the "next" button remains disabled and unclickable : 

Image

Mathieu-Deharbe (changes requested):
I tried it without the developper mode and if I select a composite the "next" button remains disabled and unclickable : 

Image

souissimai:
> dev

I m not able to reproduce it please recheck from your side , 
image
image

src/components/dialogs/import-modification-dialog.tsx
```diff
@@ -127,155 +194,290 @@ const ImportModificationDialog: ({ open, onClose }: Readonly<ImportModificationD
...
             }
 
-            setHasConfirmedFirstSelection(true);
-            const newSelections = selectedElements.map((e) => ({ id: e.id, name: e.name }));
-            setValue(SELECTED_MODIFICATIONS, newSelections, { shouldValidate: true, shouldDirty: true });
+            const newRows: SelectedComposite[] = selectedElements.map((e) => ({
+                id: e.id as UUID,
+                name: e.name,
+                originalName: e.name,
+                isShared: false, // actually is false, to be computed later
```
Mathieu-Deharbe, line 308: Actually means "en réalité", it doesn't mean "actuellement".
```suggestion
                isShared: false, // currently always false, to be computed later when shared modifications are added
```

src/components/dialogs/import-modification-dialog.tsx
```diff
@@ -127,155 +182,271 @@ const ImportModificationDialog: ({ open, onClose }: Readonly<ImportModificationD
...
+
+        insertCompositeModifications(
+            studyUuid,
+            currentNode.id,
+            modificationsToInsert,
+            formMethods.getValues(ACTION)
+        ).catch((error) => snackWithFallback(snackError, error, { headerId: 'importComposites.error' }));
+
+        handleClose();
+    }, [studyUuid, currentNode, isValid, formMethods, snackError, handleClose]);
```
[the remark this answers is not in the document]
Mathieu-Deharbe: No need. The async is handled in the network modification table with a fine error message.

Mathieu-Deharbe (approved):
Code and tests ok. I left a few secondary remarks, do as you wish with it.

gabriellevy (approved):

PR #3964, opened 2026-05-28
Fix run button state management and prevent concurrent computations

ayolab:


Mathieu-Deharbe (commented):
This is not because of your dev but very close and I encountered this while testing : I start a security analysis, then I stop it quickly. Then I try to unbuild the node and I can't : the node is stuck : 

Image

This is a quite big problem. If you don't know where it is coming from maybe we should create another anomaly ticket.

(The bug is on dev opf as well)

src/components/run-button.jsx
```diff
@@ -30,6 +30,16 @@ const RunButton = ({ runnables, activeRunnables, getStatus, computationStopped,
     );
     const [selectedRunnable, setSelectedRunnable] = useState(activeRunnables[0]);
 
+    const runningRunnable = useMemo(
+        () => activeRunnables.find((r) => getStatus(r) === RunningStatus.RUNNING) ?? null,
+        [activeRunnables, getStatus]
+    );
```
Mathieu-Deharbe, line 36: Isn't there a risk that this will return an array ?

src/components/utils/split-button.tsx
```diff
@@ -219,10 +217,6 @@ const SplitButton = ({
         return text.split('\n').map((text, i) => (i ? [<br />, text] : text));
     };
 
-    const disabledOption =
-        computationStarting || // disable if fetch starting a computation is pending
-        (runningStatus === RunningStatus.RUNNING && computationStopped); // disable if already stopped once
-
```
Mathieu-Deharbe, line 225: I thought that code was here in order to prevent a restart of a computation before a stop is handled in the back but it looks like I was wrong because even with that code I have this bug : 

Image

1. start a computation
2. cancel it immediately
3. restart the same computation quickly
4. fails (and then "succeeds" sometimes because the computation finished in the back and its result is returned)

So it looks like you are right to remove this, but I am a bit lost. One more bug ?

Mathieu-Deharbe (commented):
Your code looks good and corrects the bug so I should approve it, but those 2 bugs are worrying me.

Mathieu-Deharbe (approved):

PR #3983, opened 2026-06-03
Desactivate validation for limits in visualization mode for line modification

basseche:


flomillot (approved):
LGTM, need to check how the back-end manage empty array

PR #4000, opened 2026-06-05
Fix error on fetching segment limits when opening line creation modification

basseche:
## PR Summary

When we create a Line creation modification and we try to open line catalog we have a error.

### DIFF
--- a/src/components/dialogs/line-types-catalog/line-type-segment-form.tsx
+++ b/src/components/dialogs/line-types-catalog/line-type-segment-form.tsx
@@ -231,10 +231,10 @@ export const LineTypeSegmentForm: FunctionComponent<LineTypeSegmentFormProps> = 
     }, []);
 
     useEffect(() => {
-        if (!editData || editData[LINE_SEGMENTS]?.length) {
+        if (!editData?.[LINE_SEGMENTS]?.length) {
             return;
         }
-        arrayRef?.current?.replaceItems([]);
+        arrayRef.current?.replaceItems([]);
         const updateSegmentsLimits = async () => {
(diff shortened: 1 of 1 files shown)


src/components/dialogs/line-types-catalog/line-type-segment-form.tsx
```diff
@@ -234,14 +234,16 @@ export const LineTypeSegmentForm: FunctionComponent<LineTypeSegmentFormProps> =
         if (!editData || editData[LINE_SEGMENTS]?.length === 0) {
             return;
         }
-        arrayRef.current?.replaceItems([]);
+        arrayRef?.current?.replaceItems([]);
```
flomillot: How is it posssible to have null on arrayRef ?

src/components/dialogs/line-types-catalog/line-type-segment-form.tsx
```diff
@@ -234,14 +234,16 @@ export const LineTypeSegmentForm: FunctionComponent<LineTypeSegmentFormProps> =
         if (!editData || editData[LINE_SEGMENTS]?.length === 0) {
```
flomillot:
```suggestion
        if (!editData || editData[LINE_SEGMENTS]?.length) {
```
This should be enough to fix your problem. You don't need the code below.

flomillot (changes requested):

src/components/dialogs/line-types-catalog/line-type-segment-form.tsx
basseche: since it is initialized with null but maybe it is not necessary.

flomillot (approved):

PR #4005, opened 2026-06-08
use user Profile instead of user

ghazwarhili:
### Context
On every OIDC silent renew, oidc-client-ts loads `/silent-renew-callback` inside a
hidden iframe. Because `index.tsx` always rendered `<AppWrapper />`, the entire
application was bootstrapped inside that iframe just to end up calling
`signinSilentCallback()`: Redux store, `IntlProvider` and all translations,
AG Grid registration, MUI theme, `SnackbarProvider` and `NotificationsProvider`.
`NotificationsProvider` additionally fetched `env.json` and opened a notification
WebSocket on each renewal — all of it pure overhead repeated at every token refresh.

### Changes
- `index.tsx`: branch on the pathname before rendering. On `/silent-renew-callback`
  render a new lightweight `SilentRenewApp`; otherwise render `<AppWrapper />` as before.
- New `components/silent-renew-app.tsx`: rebuilds the `UserManager` via
  `initializeAuthenticationProd(..., isSilentRenew=true, ...)` (same settings, same
  `localStorage` state store, same Azure authority hack as the parent) and completes
  the flow through commons-ui's `SilentRenewCallbackHandler`.
- `utils/rest-api.ts`: `fetchIdpSettings` now caches the result in `localStorage`;
  added `getCachedIdpSettings`, used only by the silent-renew path, which reads that
  cache (no network) and falls back to a real fetch if it is missing/corrupted.

### Result
Inside the silent-renew iframe, requests go from a full SPA boot
(`idpSettings.json` + `env.json` + WebSocket + app bundle execution) down to a
single `POST /token` — the actual token exchange. No more spurious WebSocket
reconnections triggered by the renewal iframe.

### Testing
- Open the app, let a silent renew happen (or wait for token expiry).
- In the Network tab, confirm the iframe no longer fetches `env.json`, no longer
  re-fetches `idpSettings.json`, and does not open a WebSocket — only `POST /token` remains.
- Confirm the user stays logged in across renewals and that the parent app's
  notification WebSocket is not reconnected on each cycle.


src/vite-env.d.ts
```diff
@@ -7,3 +7,5 @@
 
 /// <reference types="vite-plugin-svgr/client" />
 /// <reference types="vite/client" />
+declare module 'typeface-roboto';
```
etiennehomer: add a type file as it is done for module-file-saver.d.ts

src/services/utils.ts
```diff
@@ -49,8 +49,29 @@ function fetchEnv() {
...
+// Always hits the network: picks up config changes on each full app load
+// AND refreshes the cache read by the silent-renew iframe.
+export function fetchIdpSettings(): Promise<IdpSettings> {
+    return fetch('idpSettings.json')
+        .then((res) => res.json())
+        .then((settings: IdpSettings) => {
+            localStorage.setItem(IDP_SETTINGS_CACHE_KEY, JSON.stringify(settings));
+            return settings;
+        });
+}
```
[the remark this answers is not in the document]
ghazwarhili: fixed

src/services/utils.ts
```diff
@@ -49,8 +49,29 @@ function fetchEnv() {
...
+    const cached = localStorage.getItem(IDP_SETTINGS_CACHE_KEY);
+    if (cached) {
+        try {
+            return Promise.resolve(JSON.parse(cached) as IdpSettings);
+        } catch {
+            // corrupted cache -> fall back to a fresh fetch
+        }
+    }
+    return fetchIdpSettings();
 }
```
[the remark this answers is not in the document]
ghazwarhili: fixed

src/components/silent-renew-app.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { useCallback, useEffect, useState } from 'react';
+import type { UserManager } from 'oidc-client-ts';
+import {
+    handleSilentRenewCallback,
+    initializeAuthenticationProd,
+    SilentRenewCallbackHandler,
+} from '@gridsuite/commons-ui';
+import { getCachedIdpSettings } from 'services/utils';
+
+export default function SilentRenewApp() {
```
etiennehomer:
```suggestion
export default function SilentRenew() {
```
Also the file to rename

src/components/silent-renew.tsx
```diff
@@ -0,0 +1,42 @@
...
+    initializeAuthenticationProd,
+    SilentRenewCallbackHandler,
+} from '@gridsuite/commons-ui';
+import { getCachedIdpSettings } from 'services/utils';
+
+export default function SilentRenewApp() {
+    const [userManager, setUserManager] = useState<UserManager | null>(null);
+
+    useEffect(() => {
+        initializeAuthenticationProd(
```
etiennehomer, line 20: Why do we have initializeAuthenticationProd and initializeAuthenticationDev in our code ?
initializeAuthenticationDev() looks not used, except in griddyna

src/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+const container = document.getElementById('root');
+const root = createRoot(container!);
+
+async function renderApp() {
+    if (window.location.pathname.endsWith('/silent-renew-callback')) {
+        root.render(<SilentRenewApp />);
+        return;
+    }
+    await import('core-js/es/array/flat-map');
+    await import('typeface-roboto');
```
etiennehomer: We should check if it's still used.
But maybe another PR

src/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+import { createRoot } from 'react-dom/client';
+import SilentRenewApp from './components/silent-renew-app';
+
+const container = document.getElementById('root');
+const root = createRoot(container!);
+
+async function renderApp() {
+    if (window.location.pathname.endsWith('/silent-renew-callback')) {
```
etiennehomer: create a constant. This string is used in other files of the app

etiennehomer (changes requested):

src/components/silent-renew.tsx
ghazwarhili: initializeAuthenticationProd used to fetch the OIDC/IdP config from the file idpSettings.json => the prod mode
initializeAuthenticationDev uses a local config so griddyna still uses the dev config mode and run without the full setup => to remove initializeAuthenticationDev maybe we need first to migrate  griddyna to the prod mode

src/components/silent-renew-app.tsx
ghazwarhili: fixed

src/vite-env.d.ts
ghazwarhili: done and the same for declare module 'core-js/es/array/flat-map'; ?

src/index.tsx
ghazwarhili: done

PR #4016, opened 2026-06-10
Switch httpd base docker image bitnami -> dhi

jonenst:
This also brings the apache httpd upgrade from 2.4.55 to 2.4.68


antoinebhs (approved):
code ok

PR #4019, opened 2026-06-11
Default name when saving single composite

Meklo:
## PR Summary

Needs https://github.com/gridsuite/commons-ui/pull/1184

### DIFF
--- a/src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
+++ b/src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
@@ -1132,6 +1132,12 @@ const NetworkModificationNodeEditor = () => {
         [handleNameChange, isMonoRootStudy, rootNetworks]
     );
 
+    // If only one modification is selected and it is of type composite, saving it in gridexplore would make it take its name by default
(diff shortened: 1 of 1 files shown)


src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
```diff
@@ -1192,6 +1198,7 @@ const NetworkModificationNodeEditor = () => {
                     type={ElementType.MODIFICATION}
                     titleId="CreateCompositeModification"
                     prefixIdForGeneratedName="GeneratedModification"
+                    defaultName={defaultSaveModificationName}
```
souissimai, line 1201: Maybe originName would be more meaningful, just a suggestion

souissimai (approved):

src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
Meklo: It could fit but I'm afraid "origin" could be misleading here since it is generally associated with a movement

PR #4020, opened 2026-06-11
Reduce NAD re-renders and DOM work, and configure edges adaptive zoom thresholds

ayolab:
- Replace useCallback with useEffectEvent for NAD event handlers (hover, node click, equipment menu, node/text-node move) to avoid unnecessary reconstruction of NetworkAreaDiagramViewer on tree node change
- Remove redundant moveNodeToCoordinates calls on diagram load, the node postions are already stored in backend
- Optimize :has() selectors in colors.ts to only inspect direct nad-winding children, avoiding costly subtree scans on large
- diagrams.
- Configure the new NetworkAreaDiagramViewer edge infos adaptive zoom thresholds

### DIFF
--- a/src/components/grid-layout/cards/diagrams/networkAreaDiagram/network-area-diagram-content.tsx
+++ b/src/components/grid-layout/cards/diagrams/networkAreaDiagram/network-area-diagram-content.tsx
@@ -126,11 +126,9 @@ const NetworkAreaDiagramContent = memo(function NetworkAreaDiagramContent(props:
     const [isEditNadMode, setIsEditNadMode] = useState<boolean>(false);
     const workspaceId = useSelector(selectActiveWorkspaceId);
 
-    // refs to keep useLayoutEffect and callbacks stable
+    // ref to keep useLayoutEffect and callbacks stable
     const isEditNadModeRef = useRef(isEditNadMode);
-    const positionsRef = useRef(positions);
     isEditNadModeRef.current = isEditNadMode;
-    positionsRef.current = positions;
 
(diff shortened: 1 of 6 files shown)


src/components/grid-layout/cards/diagrams/networkAreaDiagram/network-area-diagram-content.tsx
```diff
@@ -514,32 +510,29 @@ const NetworkAreaDiagramContent = memo(function NetworkAreaDiagramContent(props:
...
+                    svgMetadata?.nodes.map((node) => [node.equipmentId, { x: node.x, y: node.y }]) ?? []
+                );
                 for (const position of positionsRef.current) {
                     if (position.xPosition !== undefined && position.yPosition !== undefined) {
-                        diagramViewer.moveNodeToCoordinates(
-                            position.voltageLevelId,
-                            position.xPosition,
-                            position.yPosition
-                        );
+                        const metaPos = svgMetaPositionMap.get(position.voltageLevelId);
```
sBouzols: Add comment here that it optimize perf if there is a large size of metadata

ayolab: I removed this code completely, the postions are already stored in backend there is no need to reposition nodes on load, I think this was necessary when positions were stored in redux

sBouzols (approved):
Code Review OK
Tests in GridSuite with large NAD OK
* Firefox
* Edge
* Multiple NADs
* with different zoom levels, edge Infos(LF ran) etc...
     
Console warning check OK

PR #4025, opened 2026-06-12
fixes; import composite from selection

souissimai:


src/components/dialogs/import-modification-dialog.tsx
```diff
@@ -197,6 +195,9 @@ const ImportModificationDialog = ({ open, onClose }: Readonly<ImportModification
     const action = watch(ACTION);
     const selectedModifications = watch(SELECTED_MODIFICATIONS);
     const isInsertMode = action === CompositeModificationAction.INSERT;
+    // Next is enabled as soon as at least one element is selected; derived from the
+    // form so it survives step changes and folder expansions.
+    const isNextDisabled = selectedModifications.length === 0;
```
Mathieu-Deharbe, line 198: I think this is clear enough without the comment.

src/components/dialogs/import-modification-dialog.tsx
```diff
@@ -218,12 +219,12 @@ const ImportModificationDialog = ({ open, onClose }: Readonly<ImportModification
         []
     );
 
-    // SPLIT mode — name is read-only, drag-and-drop only
+    // SPLIT mode — read-only, always shows the original composite name (never the user input)
```
Mathieu-Deharbe: I think the previous comment was better.

src/components/dialogs/import-modification-dialog.tsx
Mathieu-Deharbe: You could reuse : 
`const action = watch(ACTION);`

```suggestion
            second: action === CompositeModificationAction.SPLIT ? m.originalName : m.name,
        }));

        insertCompositeModifications(studyUuid, currentNode.id, modificationsToInsert, action).catch((error) =>
            snackWithFallback(snackError, error, { headerId: 'importComposites.error' })
        );

        handleClose();
    }, [studyUuid, currentNode, isValid, formMethods, snackError, handleClose, action]);
```

You could also completely remove : 

`const rows: SelectedComposite[] = formMethods.getValues(SELECTED_MODIFICATIONS);`

and use `selectedModifications` instead. After all it it shouldn't be needed to use watch and getValues for the same field right ?

Mathieu-Deharbe (commented):
Not a big deal but I think you should simplify `handleSave` by using the watched fields instead of `getValues`.

src/components/dialogs/import-modification-dialog.tsx
souissimai: ok

Mathieu-Deharbe (approved):

PR #4029, opened 2026-06-16
Add voltage-level band filtering to NAD

flomillot:
## Overview

Adds voltage-level band filtering to the Network Area Diagram (NAD), mirroring the geographic map's existing nominal-voltage filter. The user can show/hide voltage-level bands from a sidebar; bands are masked via CSS classes on the already-rendered SVG, **without refetching** the diagram.

## Changes

- **`useNadVoltageLevelFilter` hook** — derives the voltage-level bands present in the current diagram from the SVG metadata CSS classes (`nad-voltage-level-*`), and tracks the selected bands. Returns the unselected band names so the diagram can hide them.
- **CSS-based masking** — `network-area-diagram-content.tsx` turns the unselected bands into `display: none` rules targeting the `nad-*` classes, merged into the diagram container `sx`.
- **`NominalVoltageFilter`** — the two select-all / select-none buttons are replaced by a single master checkbox with an `indeterminate` state. Added per-element style overrides so the filter blends into the NAD sidebar instead of floating like the map overlay.

Need https://github.com/gridsuite/commons-ui/pull/1203


src/components/workspace/diagrams/nad/nad-navigation-sidebar.tsx
```diff
@@ -15,19 +17,29 @@ import {
...
+    root: { boxShadow: 'none', backgroundColor: 'transparent', borderRadius: 0 },
+    nominalVoltageZone: { maxHeight: 'none', minWidth: 0, width: '100%', py: 0 },
+};
 
 interface NadNavigationSidebarProps {
     readonly nadPanelId: UUID;
-    readonly onCollapseChange?: (collapsed: boolean) => void;
+    readonly allVoltages: number[];
+    readonly selectedVoltages: number[];
+    readonly onVoltagesChange: (checkedVoltages: number[]) => void;
```
Meklo: Why not keeping nominalVoltage naming convention instead of just voltage ?

src/components/workspace/diagrams/common/navigation-sidebar.tsx
```diff
@@ -68,94 +77,140 @@ const styles = {
...
 });
+
+interface HistorySectionContentProps {
+    navigationHistory: string[];
+    onNavigate: (voltageLevelId: string) => void;
+    isItemSelected?: (voltageLevelId: string) => boolean;
+    isDisabled: boolean;
+}
+
+export const HistorySectionContent = memo(function HistorySectionContent({
```
Meklo: It could have its down dedicated tsx file no ?

src/translations/en.json
```diff
@@ -1476,6 +1476,7 @@
     "windows": "windows",
     "reset": "Reset",
     "history": "History",
+    "nadVoltageFilter": "Voltage",
```
Meklo, line 1453: Nominal voltage ?

src/translations/fr.json
```diff
@@ -1470,6 +1470,7 @@
     "workspaceUpdateError": "Erreur lors de la mise à jour de l'espace de travail",
     "reset": "Réinitialiser",
     "history": "Historique",
+    "nadVoltageFilter": "Tension",
```
Meklo, line 1448: Niveau de tension ?

Meklo (commented):
When a new nominal voltage is added in the NAD (e.g. expanding voltage levels, adding a specific voltage level..) shouldn't its nominal voltage be preselected by default if All was ticked ? It would prevent confusion why the data doesn't show up

src/translations/fr.json
flomillot: I took the name on the figma mockup

<details>
image  
</details>

src/components/workspace/diagrams/nad/nad-navigation-sidebar.tsx
flomillot: Shorter and based on mockup name, but yeah maybe I should change it

src/translations/fr.json
flomillot: Validated by PO, I let like that

src/components/workspace/diagrams/common/navigation-sidebar.tsx
flomillot: I hesitated but I kept  like it was before, I usually preferred to keep the git history

flomillot:
> When a new nominal voltage is added in the NAD (e.g. expanding voltage levels, adding a specific voltage level..) shouldn't its nominal voltage be preselected by default if All was ticked ? It would prevent confusion why the data doesn't show up

I didn't know it was possible but indeed if it's possible you're right. I Changed it.

edit :  finally after discussion with the PO we dont change them

src/components/workspace/diagrams/nad/use-nad-voltage-level-filter.ts
```diff
@@ -0,0 +1,77 @@
...
+    const presentNominalVoltages = useMemo(() => presentBaseVoltages.map((bv) => bv.minValue), [presentBaseVoltages]);
+
+    // Representative voltages currently shown (checked). Initialized once to "all shown".
+    const [selectedNominalVoltages, setSelectedNominalVoltages] = useState<number[]>();
+    useEffect(() => {
+        if (presentNominalVoltages.length > 0) {
+            // only on initialization, need computed nominalVoltages, so can't do it at useState creation
+            setSelectedNominalVoltages((prev) => prev ?? presentNominalVoltages);
+        }
+    }, [presentNominalVoltages]);
```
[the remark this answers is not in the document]
flomillot: Every opened NAD has its own hook state in its own window.
After discussion with the PO, we never reset or update the selected voltages, except when we replace the NAD via the dedicated button.

src/components/workspace/diagrams/common/navigation-sidebar.tsx
Meklo: Fair enough

Meklo (approved):

src/components/workspace/diagrams/nad/use-nad-voltage-level-filter.ts
```diff
@@ -0,0 +1,94 @@
+/**
+ * Copyright (c) 2025, RTE (http://www.rte-france.com)
```
sBouzols: 2026

src/components/workspace/diagrams/common/navigation-sidebar.tsx
sBouzols: Nope one component one file is the rule. You keep history in either way, it's a bit more difficult to read but it's ok. Please move it in a neighbours separated file.

src/components/workspace/diagrams/common/navigation-sidebar.tsx
```diff
@@ -68,94 +77,140 @@ const styles = {
...
+        <Box sx={styles.sidebar(theme, !anyExpanded, isAbsolutePositioned)}>
+            {!anyExpanded
+                ? // Collapsed rail: one icon button per section.
+                  sections.map((section) => {
+                      const enabled = !section.isDisabled;
+                      return (
+                          <Box key={section.id} sx={mergeSx(styles.headerContainer, styles.headerCollapsed)}>
+                              <IconButton
+                                  onClick={enabled ? () => toggleExpand(section.id) : undefined}
+                                  disabled={!enabled}
```
sBouzols: keep `disabled` logic do not create `enable` var and keep it clear

src/components/workspace/diagrams/common/navigation-sidebar.tsx
```diff
@@ -68,94 +77,140 @@ const styles = {
...
-        cursor: hasHistory ? 'pointer' : 'default',
+        cursor: enabled ? 'pointer' : 'default',
         '&:hover': {
-            backgroundColor: hasHistory ? theme.palette.action.hover : 'transparent',
+            backgroundColor: enabled ? theme.palette.action.hover : 'transparent',
         },
     }),
-    icon: (theme: Theme, hasHistory: boolean) => ({
-        color: hasHistory ? theme.palette.text.primary : theme.palette.text.disabled,
+    icon: (theme: Theme, enabled: boolean) => ({
```
sBouzols: then here accept a `disabled` param instead of a `enabled` one

src/components/workspace/diagrams/common/navigation-sidebar.tsx
```diff
@@ -16,18 +16,27 @@ import {
...
+
+export interface SidebarSection {
+    /** Stable identifier of the section. */
+    id: string;
+    /** Icon element rendered in the header / collapsed rail (inherits color). */
+    icon: ReactNode;
+    /** Translation id of the section title. */
+    titleId: string;
+    /** When true the section cannot be expanded (header/icon greyed out). */
+    isDisabled?: boolean;
```
sBouzols: Actually, I prefer to keep the same naming logic as in MUI. Maybe it was used before in this component and others but I would prefer naming the prop `disabled` instead of `isDisabled` for homogeneity

src/components/workspace/diagrams/common/navigation-sidebar.tsx
```diff
@@ -68,94 +77,140 @@ const styles = {
...
         '&:hover': {
-            backgroundColor: hasHistory ? theme.palette.action.hover : 'transparent',
+            backgroundColor: enabled ? theme.palette.action.hover : 'transparent',
         },
     }),
-    icon: (theme: Theme, hasHistory: boolean) => ({
-        color: hasHistory ? theme.palette.text.primary : theme.palette.text.disabled,
+    icon: (theme: Theme, enabled: boolean) => ({
+        display: 'flex',
+        color: enabled ? theme.palette.text.primary : theme.palette.text.disabled,
```
sBouzols: strange, it should be manage by the IconButton not the icon itself...

src/components/workspace/diagrams/nad/nad-navigation-sidebar.tsx
```diff
@@ -53,24 +65,52 @@ export const NadNavigationSidebar = memo(function NadNavigationSidebar({
...
-            isDisabled={!isBuilt}
-            isItemSelected={(id) => associatedVoltageLevelIds.includes(id)}
-            onToggleCollapse={handleToggleExpand}
-            onNavigate={handleNavigationSidebarClick}
-        />
+    const sections = useMemo<SidebarSection[]>(
+        () => [
+            {
+                id: 'history',
+                icon: <HistoryIcon fontSize="medium" />,
```
sBouzols: In a `size=small` `IconButton` we often put a `fontSize="small"` icon then i would change it for homogeneity if Designer is ok.

src/components/workspace/diagrams/common/navigation-sidebar.tsx
```diff
@@ -68,94 +77,140 @@ const styles = {
...
+    return (
+        <List dense sx={{ py: 0 }}>
+            {navigationHistory.map((voltageLevelId, index) => (
+                <ListItemButton
+                    key={`${voltageLevelId}-${index}`}
+                    selected={isItemSelected ? isItemSelected(voltageLevelId) : false}
+                    onClick={() => onNavigate(voltageLevelId)}
+                    disabled={isDisabled}
+                >
+                    <ListItemIcon sx={{ minWidth: 32 }}>
```
sBouzols: I don't like this `minWidth` we should use flex, this should be removed

src/components/workspace/diagrams/common/navigation-sidebar.tsx
flomillot: Both were used, I decided to make it clearer by using `isDisabled` when its our components and letting `disabled` to MUI props. 
It seems to be what we do now ? (edit : both exists)

Image

But at first I also wanted to put `disabled` everywhere so I'm also OK with that if a TL confirm it's what we should do and then we can apply this rule everywhere.

@etiennehomer ?

src/components/workspace/diagrams/common/navigation-sidebar.tsx
flomillot: Well I kept the existing style behaviour

src/components/workspace/diagrams/common/navigation-sidebar.tsx
flomillot: Yes better to uniformize, I change it everywhere

src/components/workspace/diagrams/nad/nad-navigation-sidebar.tsx
flomillot: Alright it was already like that anyway

src/components/workspace/diagrams/common/navigation-sidebar.tsx
flomillot: I dont like either but sometimes CSS is tricky. I suppose it was here for a reason, and I dont want to introduce a visual regression by removing it, even if from my test, it seems to be useless.

sBouzols: No my bad it's ok, it's because of a static min width of the MUI component...

sBouzols (approved):
Code Review OK
Test OK
Console warning check OK

src/components/workspace/diagrams/common/navigation-sidebar.tsx
sBouzols: changed after TL accepts thanks

PR #4040, opened 2026-06-18
Parameters new design

khouadrired:
use new Parameters Layout 
refactor State Estimation Params


FranckLecuyer (approved):

src/components/results/common/glass-pane.tsx
```diff
@@ -31,19 +31,20 @@ interface GlassPaneProps {
     active: boolean;
     children: ReactNode;
     loadingMessageText?: string;
+    sx?: any;
```
thangqp:
```suggestion
    sx?: SxProps;
```

thangqp (approved):

src/components/dialogs/parameters/state-estimation/state-estimation-general-parameters.tsx
```diff
@@ -17,49 +17,53 @@ import { parametersStyles } from '../util/styles';
...
-        <Stack>
-            <Grid container alignItems="center" spacing={2}>
-                <Grid size={10} sx={parametersStyles.parameterName}>
-                    <FieldLabel label={'StateEstimationParametersPrincipalObservableZoneLabel'} />
-                </Grid>
-                <Grid size={2}>
-                    <SwitchInput name={`${TabValue.GENERAL}.${PRINCIPAL_OBSERVABLE_ZONE}`} />
-                </Grid>
-            </Grid>
+        <Grid container>
```
antoinebhs, line 20: why not keep stack here?

src/components/parameters-tabs.tsx
```diff
@@ -597,4 +675,23 @@ const ParametersTabs: FunctionComponent = () => {
     );
 };
 
+const useContainerXs = (threshold = 600) => {
```
antoinebhs, line 674: make sense to have it in a util or external hook maybe?

thangqp: He will consider do that in the next PR which does the adaptation for each inside fields..

antoinebhs: I'm not aware of that, do we have a ticket/issue for that?

thangqp: image
We must have a Jira ticket for this aspect.. containerXs in this work is only for switching the menu, but not yet used for the adaptation inside each form.

src/components/dialogs/parameters/state-estimation/state-estimation-general-parameters.tsx
thangqp: I want all fields take 8/12 columns. Stack as root has not size prop

PR #4044, opened 2026-06-19
Ihm shared modif

souissimai:


src/components/graph/menus/network-modifications/network-modification-table/createColumns.tsx
```diff
@@ -62,12 +63,24 @@ export const createBaseColumns = (
...
         minSize: 160,
     },
+    {
+        id: BASE_MODIFICATION_TABLE_COLUMNS.SHARE.id,
+        cell: ShareCellRenderer,
+        size: 48,
+        minSize: 48,
+        meta: {
+            cellStyle: {
+                paddingRight: '5px',
```
Mathieu-Deharbe: This should be checked with Stéphane but I think that the share icon is too far from description : 
Image

`DescriptionCellRenderer` has much smaller size than `ShareCellRenderer`.

When I set paddingRight to 0px I don't see any difference. Are you sure that this is useful ?

Mathieu-Deharbe (changes requested):
The tooltip over the share button is currently "copy the link". According to the jira specs it should be "shared/Partagé".

src/components/graph/menus/network-modifications/network-modification-table/createColumns.tsx
```diff
@@ -62,12 +63,24 @@ export const createBaseColumns = (
         },
         minSize: 160,
     },
+    {
+        id: BASE_MODIFICATION_TABLE_COLUMNS.SHARE.id,
+        cell: ShareCellRenderer,
+        size: 48,
+        minSize: 48,
```
Mathieu-Deharbe: I think this would be enough : 
```suggestion
        size: 40,
        minSize: 32,
```
(like DescriptionCellRenderer)
But Stéphane should decide.

src/components/graph/menus/network-modifications/network-modification-table/createColumns.tsx
souissimai: to be checked with UX designer

src/components/graph/menus/network-modifications/network-modification-table/createColumns.tsx
```diff
@@ -62,12 +63,24 @@ export const createBaseColumns = (
         },
         minSize: 160,
     },
+    {
+        id: BASE_MODIFICATION_TABLE_COLUMNS.SHARE.id,
+        cell: ShareCellRenderer,
```
Mathieu-Deharbe:
```suggestion
        id: BASE_MODIFICATION_TABLE_COLUMNS.SHARED.id,
        cell: ReferenceCellRenderer,
```

Mathieu-Deharbe (commented):
Given that shared modifications are only available in dev mode this is not a big deal, but maybe we should prevent any drag and drop into reference modifications for now. It causes this : 

Image

souissimai:
> copy the link

changed it

src/components/graph/menus/network-modifications/network-modification-table/createColumns.tsx
souissimai: yes

souissimai:
> Given that shared modifications are only available in dev mode this is not a big deal, but maybe we should prevent any drag and drop into reference modifications for now. It causes this :
> Image

managed it

src/translations/en.json
```diff
@@ -1508,5 +1508,8 @@
     "button.next": "Next",
     "button.previous": "Previous",
     "disconnectedLoadActivePower": "Load cut off (MW)",
-    "disconnectedGenerationActivePower": "Generation cut off (MW)"
+    "disconnectedGenerationActivePower": "Generation cut off (MW)",
+    "copyLink": "Copy link",
+    "linkCopied": "Link copied",
+    "linkCopyError": "Error while attempting to cooy link"
```
Mathieu-Deharbe: Those texts are used in commons-ui so I think they should be there. Especially because soon they will be used in gridexplore as well.

Mathieu-Deharbe (commented):
OK for everything except that the text should be put in commons-ui in my opinion.

src/translations/en.json
souissimai: yes done

Mathieu-Deharbe (approved):

src/components/dialogs/import-composite/import-modification-dialog.tsx
```diff
@@ -261,6 +262,7 @@ const ImportModificationDialog = ({ open, onClose }: Readonly<ImportModification
 
         const modificationsToInsert: CompositesToBeInserted[] = selectedModifications.map((m: SelectedComposite) => ({
             id: m.id,
+
```
flomillot: Can you remove the empty lines you added ?

flomillot (changes requested):

src/components/dialogs/import-composite/import-modification-dialog.tsx
Mathieu-Deharbe: Done here : https://github.com/gridsuite/gridstudy-app/pull/4044/changes/7d1d19f13ed2ad267ecdb972afe1994c12bdbf37

flomillot: I mean it's not the only place there is many new empty lines
image

flomillot (commented):
Image

flomillot (approved):

PR #4049, opened 2026-06-22
Allow recovery after invalid tabular CSV import

flomillot:
Tabular create/modify dialogs only validated cell types at CSV import time (via `setError` on the table field). Since the submit button is disabled while any error is present and nothing cleared that error on edit, an invalid import left the dialog permanently blocked — there was no way to correct a value and validate again.

Changes:
- Sanitize imported values: wrong-format cells are dropped (replaced by `null`, kept out of the table) instead of being injected raw.
- Mandatory boolean (checkbox) columns default to `false` when empty/invalid, since an empty checkbox cannot be distinguished from `false`.
- Invalid boolean values are no longer blocking errors: they are replaced by `false` and only surface a warning, displayed below the error alert, indicating the values were replaced automatically.

Relies on the companion change in commons-ui (clearing the table field error on cell edit) to re-enable the submit button once the user edits a cell.

`limit-sets` is intentionally left out (its cells are read-only).

Also bump commons-ui to 236

### DIFF
--- a/src/components/dialogs/network-modifications/tabular/tabular-form.tsx
+++ b/src/components/dialogs/network-modifications/tabular/tabular-form.tsx
@@ -46,13 +46,14 @@ import {
     generateCommentLines,
     isFieldTypeOk,
     PredefinedEquipmentProperties,
+    sanitizeRowValue,
     setFieldTypeError,
     TabularField,
     TabularModificationType,
     transformIfFrenchNumber,
 } from './tabular-common';
 import { ColDef } from 'ag-grid-community';
-import { ENUM, NUMBER } from '../../../network/constants';
+import { BOOLEAN, ENUM, NUMBER } from '../../../network/constants';
 import { TABULAR_CREATION_FIELDS } from './tabular-creation-utils';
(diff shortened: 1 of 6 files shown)


basseche (approved):

src/components/dialogs/network-modifications/tabular/tabular-form.tsx
```diff
@@ -225,6 +240,19 @@ export function TabularForm({ dataFetching, dialogMode }: Readonly<TabularFormPr
...
+                        if (!isFieldTypeOk(value, fieldDef)) {
+                            setFileWarningMessage(
+                                intl.formatMessage(
+                                    { id: 'WrongBooleanValueWarning' },
+                                    { field: intl.formatMessage({ id: key }) }
+                                )
+                            );
+                        }
+                        return false; // keep looking, a boolean never raises an error
+                    }
```
basseche: this code is exactly like the one above why not create a function

flomillot: I hesitated but you're right lets make an helper

PR #4058, opened 2026-06-24
Fix geographical Map : lines have incorrect operational statuses

sBouzols:
## PR Summary

fix(NetworkMapPanel): Do not overwrite `updatedLines` at each `reloadMapEquipments` but enrich the previous one. In fact Impacts from the last modification retrieve only the last substations impacted then we must enrich the state.


src/components/network/network-map-panel.tsx
```diff
@@ -85,6 +88,23 @@ const LABELS_ZOOM_THRESHOLD = 9;
 const ARROWS_ZOOM_THRESHOLD = 7;
 const EMPTY_ARRAY: any[] = [];
 
+function mergeByIdKeepOrder<T extends { id: string }>(array1: T[], array2: T[]): T[] {
+    const a2ById = new Map<string, T>();
```
EtienneLt:
```suggestion
    const array2ById = new Map<string, T>();
```

src/components/network/network-map-panel.tsx
```diff
@@ -85,6 +88,23 @@ const LABELS_ZOOM_THRESHOLD = 9;
 const ARROWS_ZOOM_THRESHOLD = 7;
 const EMPTY_ARRAY: any[] = [];
 
+function mergeByIdKeepOrder<T extends { id: string }>(array1: T[], array2: T[]): T[] {
```
EtienneLt: add that ?
```typescript
if (array1.length === 0) {
        return [...array2];
    }

    if (array2.length === 0) {
        return [...array1];
    }
```

sBouzols: Ok but we need to keep object reference to avoid useless state updates then I returned array1 ou array2 without copy.

EtienneLt: ok

EtienneLt (approved):
Test OK code OK

PR #4059, opened 2026-06-24
Prepare global filter migration

klesaulnier:
## PR Summary

This PR prepares the future migration of the `global-filter` directory to `commons-ui`.

The following subdirectories and files will be excluded from this migration and will remain in the current project:

* `hooks`
* `adapter`
* `src/components/results/common/global-filter/global-filter-selector.tsx`

As part of this preparation, the `GlobalFilter` UI component has been refactored to remove any direct dependency on Redux or data fetching. It is now responsible only for rendering and user interactions, making it reusable across projects.

This PR also reorganizes and isolates the code that will be migrated alongside the `GlobalFilter` UI component in the upcoming `commons-ui` migration PR.


src/components/results/common/global-filter/adapter/global-filter-provider.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+type GlobalFilterProviderProps = PropsWithChildren<{
+    filterCategories: FilterType[];
+    genericFiltersStrictMode: boolean;
+    filterableEquipmentTypes: string[];
+    tableType: TableType;
+    tableUuid: string;
+}>;
+
+export default function GlobalFilterProvider({
```
thangqp: @klesaulnier :
The following subdirectories will be excluded from this migration and will remain in the current project:
    hooks
    adapter
As part of this preparation, the GlobalFilter UI component has been refactored to remove any direct dependency on Redux or data fetching. It is now responsible only for rendering and user interactions, making it reusable across projects.

@thangqp: As I understand, only the JSX components (and context) will be moved to commons-ui, while the hooks and adapter directories will remain in gridstudy-app.

I noticed that these directories contain a lot of business logic that could be reused by other applications. If this logic is not moved to commons-ui, we'll end up duplicating them across other apps.

src/components/results/common/global-filter/types/limit-violation.type.ts
```diff
@@ -0,0 +1,12 @@
+/**
```
antoinebhs, line 1: typo in naming

src/components/results/common/global-filter/context/global-filter-context.type.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export type GlobalFilterContextValue = {
+    globalFilterOptions: GlobalFilter[];
+    selectedGlobalFilters: GlobalFilter[];
+    recentGlobalFilters: RecentGlobalFilter[];
+    substationPropertiesGlobalFilters?: Map<string, string[]>;
+    filterCategories: string[];
+    genericFiltersStrictMode: boolean;
+    filterableEquipmentTypes: string[];
+    translateCountryCode: (countryCode: string) => string;
```
antoinebhs: do we really need it in the context? util?

antoinebhs: same for these:
filterCategories                       
genericFiltersStrictMode               
filterableEquipmentTypes               
substationPropertiesGlobalFilters      
translateCountryCode                   

should it be in the context or passed to the component?

antoinebhs (commented):
fetchSubstationPropertiesGlobalFilters 
called twice?

src/components/results/common/global-filter/context/global-filter-context.type.ts
antoinebhs: we can discuss the separation provider/context/component maybe

antoinebhs (approved):
LGTM!

antoinebhs (approved):

thangqp (approved):

PR #4064, opened 2026-06-25
fix: display state estimation data when diverged

anass45:
Fix bug : display state estimation data when diverged in tabs Status, Criterion quality and Quality per region

### DIFF
--- a/src/components/results/stateestimation/state-estimation-quality-result.tsx
+++ b/src/components/results/stateestimation/state-estimation-quality-result.tsx
@@ -24,7 +24,6 @@ import { RenderTableAndExportCsv } from '../../utils/renderTable-ExportCsv';
 import { AgGridReact } from 'ag-grid-react';
 import { StateEstimationResultProps } from './state-estimation-result.type';
 import { TableType } from 'types/custom-aggrid-types';
-import { getRows } from './state-estimation-result-utils';
 
 export const StateEstimationQualityResult: FunctionComponent<StateEstimationResultProps> = ({
     result,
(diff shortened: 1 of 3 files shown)


src/components/results/stateestimation/state-estimation-quality-result.tsx
```diff
@@ -24,6 +24,7 @@ import { RenderTableAndExportCsv } from '../../utils/renderTable-ExportCsv';
 import { AgGridReact } from 'ag-grid-react';
 import { StateEstimationResultProps } from './state-estimation-result.type';
 import { TableType } from 'types/custom-aggrid-types';
+import { getRows } from './state-estimation-result-utils';
```
TheMaskedTurtle: Could you just remove this getRows usage ? It is quite useless because we already memoize result to be empty when different from success or diverged, so we don't need to filter it again here

src/components/results/stateestimation/state-estimation-result-utils.ts
```diff
@@ -64,3 +65,7 @@ export const stateEstimationQualityPerRegionColumnsDefinition = (intl: IntlShape
         }),
     ];
 };
+
+export function getRows(rows: any[] | undefined, status: string): any[] {
```
TheMaskedTurtle: No need

TheMaskedTurtle (changes requested):

src/components/results/stateestimation/state-estimation-quality-result.tsx
anass45: done

src/components/results/stateestimation/state-estimation-result-utils.ts
anass45: done

src/components/results/stateestimation/state-estimation-result-tab.tsx
```diff
@@ -82,7 +82,10 @@ export const StateEstimationResultTab: FunctionComponent<StateEstimationTabProps
     };
 
     const result = useMemo(() => {
-        if (stateEstimationStatus !== RunningStatus.SUCCEED || !stateEstimationResult) {
+        if (
+            !(stateEstimationStatus === RunningStatus.SUCCEED || stateEstimationStatus === RunningStatus.FAILED) ||
+            !stateEstimationResult
+        ) {
```
TheMaskedTurtle: Maybe we can reverse the condition now ? It's quite hard to understand

anass45: done

TheMaskedTurtle (approved):

PR #4069, opened 2026-06-29
Fix spreadsheet scroll to equipment

ayolab:
set `rowData` imperatively to ensure rows are loaded before `ensureNodeVisible` is triggered
a partial rollback on #3950 : passing `rowData` as a prop to `CustomAGGrid` creates a race condition where the scroll callback runs before AG Grid has processed the new row data


sBouzols (approved):
Code Review OK
Test OK
Console warning OK

PR #4072, opened 2026-06-30
Move the generate-template button above the tabular table

flomillot:
## Change

Follow-up to gridsuite/commons-ui#1219 (CSV buttons rework). On the tabular and limit-sets modification forms:

- Render the contained **"Generate template"** button above the table (it no longer comes from the bottom toolbar), next to "Define properties" (tabular) / the temporary-limits count (limit sets).
- Group the optional **prefilled-model** button next to "Generate template" (tabular), instead of the bottom toolbar.
- The imported file name truncates (single line + tooltip) so it no longer pushes the import row to a new line.
- Hide every field below the equipment type selector until a type is selected. Previously the buttons and the import picker were only disabled and just the table was hidden.
- Restore the row spacing lost when the root `Grid` container was replaced by a `Stack`: a nested `Grid` container only inherits the spacing CSS variables from a parent `Grid` container, so the row now declares `spacing` explicitly.

## Dependency

Requires gridsuite/commons-ui#1219 (exported `CsvDownloadButton`, `CsvProps` without `getTemplateData`) to be merged and released first.


src/components/dialogs/network-modifications/limit-sets/limit-sets-tabular-modification-form.tsx
```diff
@@ -277,20 +278,31 @@ export function LimitSetsTabularModificationForm({ dataFetching }: Readonly<Tabu
...
             <Grid container justifyContent="space-between" alignItems="center">
-                <Grid>
-                    <IntegerInput name={AMOUNT_TEMPORARY_LIMITS} label={'amountTemporaryLimits'} />
+                <Grid container alignItems="center">
+                    <Grid>
+                        <IntegerInput name={AMOUNT_TEMPORARY_LIMITS} label={'amountTemporaryLimits'} />
+                    </Grid>
+                    <Grid>
+                        <CsvDownloadButton
+                            data={getTemplateData}
```
ghazwarhili: I'm a little confused, you removed getTemplateData from the component props, 
so should data now use getTableData?

src/components/dialogs/network-modifications/tabular/tabular-form.tsx
```diff
@@ -544,18 +534,41 @@ export function TabularForm({ dataFetching, dialogMode }: Readonly<TabularFormPr
...
+                            onClick={() => {
+                                propertiesDialogOpen.setTrue();
+                            }}
+                        >
+                            <FormattedMessage id="DefinePropertiesButton" />
+                        </Button>
+                    </Grid>
+                    <Grid>
+                        <CsvDownloadButton
+                            data={getTemplateData}
```
ghazwarhili: the same here

src/components/dialogs/network-modifications/limit-sets/limit-sets-tabular-modification-form.tsx
flomillot: I moved the button somewhere else as requested by UX.
So I moved the associated callback.

basseche (approved):

ghazwarhili (approved):

PR #4078, opened 2026-07-03
Fix process is not defined crash from react-draggable following dependency upgrade

antoinebhs:
## PR Summary

Prevent the `process is not defined` crash in `react-draggable` by inlining `process.env.DRAGGABLE_DEBUG` to `false` in the Vite build.   
Crash happened when moving panels around in the gridstudy-app.
Related issue: https://github.com/react-grid-layout/react-draggable/issues/806 (still open).

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -27,6 +27,11 @@ const serverSettings: CommonServerOptions = {
 };
 
 export default defineConfig((_config) => ({
+    define: {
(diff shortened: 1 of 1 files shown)


benrejebmoh (approved):
test ok

FranckLecuyer (approved):
Test: OK

PR #4084, opened 2026-07-08
Loader while filtering with global filters in spreadsheet

flomillot:
## Issue

When loading a spreadsheet collection that has a preset global filter, the grid first displays all unfiltered rows (with the global total), then reloads a second later with the filter applied.

## Root cause

The global filter is applied client-side as an AG Grid external filter, whose matching equipment ids are computed by `useGlobalFilterResults` → `evaluateGlobalFilter`, a backend call.

**Nothing tells the grid that the evaluation is still in flight.** `filteredIds` starts `undefined`, and `doesExternalFilterPass` then lets every row pass. So between the moment a filter is set and the arrival of its ids, the grid displays unfiltered data — this is the flash, and the 700ms debounce on the evaluation only makes the window longer.

A second, more specific issue could makes the collection case worse: `INIT_TABLE_DEFINITIONS` receives the full `GlobalFilter` objects from the backend but only registers the ones matching `isCriteriaFilter` in `globalFilterOptions`. `COUNTRY` / `VOLTAGE_LEVEL` / `SUBSTATION_PROPERTY` options are therefore expected from `useGlobalFilterOptions()`, which fetches them asynchronously, racing the collection load. Until they arrive, `useSelectedGlobalFilters` resolves nothing and the tab behaves as if no filter was set at all.

## Fix

- `useGlobalFilterResults` now returns `{ filteredIds, isPending }`. While pending, no row passes the external filter and the grid keeps its loading overlay, so unfiltered rows are never displayed and the row counter never shows the global total. The previous result is kept during a re-evaluation, so changing a filter does not blank the grid. `SpreadsheetContent` folds `isPending` into the `isFetching` it already passes to AG Grid, so this shows up as a single loading phase — the equipment fetch is untouched and still runs in parallel with the filter evaluation.
- Register **all** filters of the collection in `globalFilterOptions`. Known options are left untouched, and `ADD_TO_GLOBAL_FILTER_OPTIONS` upserts by id, so the options fetched from the network always win over the ones restored from a stale collection.
- The row counter is kept on a single loading phase for the whole evaluation. Rejecting every row while pending takes the grid through two filtering passes, and the counter's own 600ms debounce is shorter than the 700ms evaluation debounce plus the backend call: it used to display the intermediate empty count and spin twice. It now skips the updates triggered while an evaluation is pending and replays them once it has resolved. `useSpreadsheetGlobalFilter` therefore moves up from `SpreadsheetContent` to `Spreadsheet`, so the grid and the counter share the same pending state.

## Also fixed along the way

Removing the **last** global filter left the grid filtered: the effect calling `onFilterChanged()` only depended on the filtered ids and on the pending state, and neither changes in that case. It now depends on `isExternalFilterPresent` as well.


src/redux/reducer.ts
```diff
@@ -1586,6 +1588,8 @@ export const reducer = createReducer(initialState, (builder) => {
             state.tableFilters.globalFilters[action.tabUuid] ??= { selected: [], recents: [] };
             if (!areSelectedEqual) {
                 state.tableFilters.globalFilters[action.tabUuid].selected = newSelectedIds;
+                // programmatic change (init/sync): evaluate the filter immediately, without debounce
+                state.tableFilters.globalFilters[action.tabUuid].lastChangeFromUser = false;
```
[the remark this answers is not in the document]
flomillot: Not accurate anymore

src/components/results/common/global-filter/use-global-filter-results.ts
```diff
@@ -18,53 +18,55 @@ import { buildValidGlobalFilters } from './build-valid-global-filters';
...
+                // nothing to evaluate: don't leave the caller waiting for ids that will never come
+                setIsPending(false);
+                return;
             }
+            evaluateGlobalFilter(studyUuid, currentNode.id, currentRootNetworkUuid, equipmentTypesParam, globalFilters)
+                .then(setFilteredIds)
+                .catch((error) => {
+                    snackWithFallback(snackError, error, { headerId: 'FilterEvaluationError' });
+                })
+                .finally(() => setIsPending(false));
```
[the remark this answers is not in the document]
flomillot: very unlikely with a 700ms debounce

src/components/results/common/global-filter/hooks/use-global-filter-results.ts
```diff
@@ -25,46 +25,48 @@ export function useGlobalFilterResults(filters: GlobalFilter[], equipmentTypes:
...
-                        currentRootNetworkUuid,
-                        equipmentTypesParam,
-                        globalFilters
-                    )
-                        .then(setFilteredIds)
-                        .catch((error) => {
-                            snackWithFallback(snackError, error, { headerId: 'FilterEvaluationError' });
-                        });
+        (globalFiltersParam: GlobalFilters, equipmentTypesParam: NonEmptyTuple<FilterEquipmentType>) => {
+            if (!studyUuid || !nodeUuid || !currentRootNetworkUuid) {
```
Meklo, line 44: There's cases where canEvaluate be true but not this expression ?

src/components/spreadsheet-view/spreadsheet/spreadsheet-content/spreadsheet-content.tsx
```diff
@@ -248,7 +248,7 @@ export const SpreadsheetContent = memo(
                             rowData={transformedRowData}
                             currentNode={currentNode}
                             columnData={columns}
-                            isFetching={equipments.isFetching}
+                            isFetching={equipments.isFetching || isGlobalFilterPending}
```
Meklo, line 250: It prevents displaying stale data on initial startup but now also displays loading screen for all global filters changes, are POs okay with that change ?

src/redux/reducer.ts
```diff
@@ -814,13 +830,7 @@ export const reducer = createReducer(initialState, (builder) => {
...
                 recents: buildSortedRecents(recentFilters),
             };
-            // Store full objects in globalFilterOptions only if not already present
-            filters.filter(isCriteriaFilter).forEach((filter) => {
-                const alreadyExists = state.globalFilterOptions.some((opt) => opt.uuid === filter.uuid);
-                if (!alreadyExists) {
-                    state.globalFilterOptions.push(addGlobalFilterId(filter));
-                }
-            });
+            registerGlobalFilterOptions(state.globalFilterOptions, filters);
```
Meklo, line 844: Wasn't there any specific reason to filter items using `isCriteriaFilter` before? It's a little strange that this condition was added in the first place.

Meklo (commented):
Reported bug thoroughly fixed, just few remarks

src/components/results/common/global-filter/hooks/use-global-filter-results.ts
flomillot: It's just for typescript

src/components/spreadsheet-view/spreadsheet/spreadsheet-content/spreadsheet-content.tsx
flomillot: Fair point I didn't think about, I'll ask them

src/redux/reducer.ts
flomillot: It's because they were already included via the hook `useGlobalFilterOptions`.
Honestly it doesn't change much.

src/components/spreadsheet-view/spreadsheet/spreadsheet-content/spreadsheet-content.tsx
flomillot: After asking to the PO, we can let it like that and we will change later if necessary

Meklo (approved):

PR #4085, opened 2026-07-08
Work around for React useEffectEvent + memo() stale closure bug in NAD content

ayolab:
Workaround for https://github.com/react/react/issues/35187 and https://github.com/react/react/issues/35034:
useEffectEvent retains the first render value when is used inside a component wrapped in memo()
This workaround should be removed once the issue is fixed


sBouzols (approved):
Code Review OK
Tests OK
Console warning check OK

PR #4096, opened 2026-07-15
fix icon regression on line atrach to voltage level

JacquesBorsenberger:
## PR Summary
fix icon regression on line atrach to voltage level


anass45:
Tests OK

anass45 (approved):

PR #4107, opened 2026-07-21
use only authorized server to make rest calls

klesaulnier:
## PR Summary

Migrate all API calls in order to use only authorized servers


src/services/directory.ts
```diff
@@ -8,8 +8,9 @@
 import { getRequestParamFromList } from './utils';
 import { backendFetchJson, ElementAttributes } from '@gridsuite/commons-ui';
 import type { UUID } from 'node:crypto';
+import { PREFIX_STUDY_QUERIES } from './study';
 
-const PREFIX_DIRECTORY_SERVER_QUERIES = import.meta.env.VITE_API_GATEWAY + '/directory';
+const PREFIX_DIRECTORY_QUERIES = PREFIX_STUDY_QUERIES + '/v1/directory';
```
antoinebhs, line 13: this endpoint exist already in explore? i think we should not duplicate if it already exsit?

src/services/dynamic-simulation.ts
```diff
@@ -7,6 +7,7 @@
 
 import { backendFetch } from '@gridsuite/commons-ui';
 import type { UUID } from 'node:crypto';
+import { PREFIX_STUDY_QUERIES } from './study';
 
 const PREFIX_DYNAMIC_SIMULATION_SERVER_QUERIES = import.meta.env.VITE_API_GATEWAY + '/dynamic-simulation';
```
antoinebhs: This one needs to go by study

klesaulnier: fixed

antoinebhs (approved):

PR #4121, opened 2026-07-31
fix(TabularModification): preserve table rows when property selection changes

flomillot:
## Summary

- Adding or removing a property in the "Define properties" dialog of the tabular creation/modification forms used to wipe the whole table.
- Existing rows are now preserved: property columns removed from the selection are dropped from each row, and newly selected ones are added with an empty value.
- The CSV filename is no longer reset either, since the loaded data is kept.


src/components/dialogs/network-modifications/tabular/tabular-form.tsx
```diff
@@ -511,10 +511,26 @@ export function TabularForm({ dataFetching, dialogMode }: Readonly<TabularFormPr
...
+            const addedPropertyColumns = newSelectedProperties
+                .filter((name) => !selectedProperties.includes(name))
+                .map((name) => PROPERTY_CSV_COLUMN_PREFIX + name);
+            const currentRows = (getValues(MODIFICATIONS_TABLE) ?? []) as Record<string, unknown>[];
+            const updatedRows = currentRows.map((row) => {
+                const newRow = { ...row };
+                removedPropertyColumns.forEach((col) => {
+                    delete newRow[col];
+                });
+                addedPropertyColumns.forEach((col) => {
```
etiennehomer, line 528: No needed ?

etiennehomer (approved):

PR #4125, opened 2026-08-03
move cspr form to commons-ui

anass45:
Moving the CSPR creation form, from gridstudy to commons-ui


KoloMenek (approved):
Looks good to me, tested and working properly

PR #4137, opened 2026-08-07
Handle the modification applicability per root network tag

flomillot:
Depends on gridsuite/commons-ui#1281.

study-server now exposes the applicability of each modification per root network tag instead of a list of exclusions per root network.

- `fetchExcludedNetworkModifications` is removed and nothing replaces it: the applicability comes with the modifications the panel already fetches.
- The table resolves each chip on the uuid of its root network, the tag it is labelled with being renamable.
- The modifications are fetched again on `rootNetworksUpdated`: they carry their applicability keyed by tag, so the ones in hand still name a renamed root network by its previous one.

TODO : manage modification of reference modification once the rights will be merged


Mathieu-Deharbe (approved):

Mathieu-Deharbe (changes requested):
Sorry I changed my mind while reading study-server. I think most of the code surrounding `fetchNetworkModificationApplicabilities` should be changed because the applicability data should be fetched along with the network modification data, just like the "activated" boolean value in `NetworkModificationMetadata`.

cf : https://github.com/gridsuite/study-server/pull/1053#discussion_r3821287171

Meklo (approved):

Mathieu-Deharbe (approved):

PR #4148, opened 2026-08-13
actions , ref modif

souissimai:


src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
```diff
@@ -179,14 +180,6 @@ const NetworkModificationNodeEditor = () => {
...
     );
 
-    // TODO : this is temporary, until copy/paste/save is done for the shared modifications in GRD-4785 :
-    const selectionContainsShared: boolean = useMemo(() => {
-        return selectedNetworkModifications.some(
-            (modification: ComposedModificationMetadata) =>
-                modification.type === ModificationType.MODIFICATION_REFERENCE
-        );
-    }, [selectedNetworkModifications]);
-
```
Mathieu-Deharbe, line 189: Not your fault but I think we should keep this and use it in order to disable the composite creation, which doesn't work yet for reference modifications. Ie : 

`const disabledCompositeCreation: boolean = useMemo(() => {`

And Maybe create a ticket, I will see with Thibault.

src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
```diff
@@ -970,10 +963,21 @@ const NetworkModificationNodeEditor = () => {
...
+        Promise.all(
+            selectedNetworkModifications.map((item) =>
+                item.type === MODIFICATION_TYPES.MODIFICATION_REFERENCE.type
+                    ? fetchNetworkModification(item.uuid as UUID)
+                          .then((res) => res.json())
+                          .then((detail: ReferenceModificationInfos) => detail.referenceId ?? null)
+                    : Promise.resolve(item.uuid)
+            )
+        )
+            .then((uuids) => uuids.filter((uuid): uuid is UUID => uuid !== null))
```
Mathieu-Deharbe, line 977: I think we shouldn't go ahead and save a composite when `detail.referenceId` is null, the operation should be interrupted with a : 
`snackWithFallback(snackError, error, { headerId: 'errCreateModificationsMsg' });`

src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
```diff
@@ -970,10 +963,21 @@ const NetworkModificationNodeEditor = () => {
...
+            .then((selectedModificationsUuid) =>
+                createCompositeModifications(name, description, folderId, selectedModificationsUuid)
+            )
```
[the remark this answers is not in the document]
Mathieu-Deharbe: Technically true but I think this would be better to completely interrupt the operation if this ever happen : https://github.com/gridsuite/gridstudy-app/pull/4148#discussion_r3805083808

src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
souissimai: would you mean the merge into composite?

Mathieu-Deharbe: Yes. For now it seems to work but it doesn't update the references, so it doesn't really work.

souissimai: I will block it right

Mathieu-Deharbe (approved):

src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
```diff
@@ -179,7 +180,7 @@ const NetworkModificationNodeEditor = () => {
     const [selectedNetworkModifications, setSelectedNetworkModifications] = useState<ComposedModificationMetadata[]>(
         []
     );
-
+    console.log('======== ', selectedNetworkModifications);
```
Mathieu-Deharbe: Should be removed.

src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
```diff
@@ -1023,39 +1040,43 @@ const NetworkModificationNodeEditor = () => {
...
             return;
         }
+        // no source hint: study-server now looks up each modification's real container itself
+        // (network-modification-server owns that data), instead of this having to guess it from
+        // whatever the table's selection happens to expose
+        const modificationsToMoveOrCopy: ModificationMoveOrCopyInfos[] = networkModificationsToCopy.map(
+            (modification) => ({
+                modificationUuid: modification.uuid,
+            })
+        );
```
Mathieu-Deharbe, line 992: Ok but in this case what is the point of `ModificationMoveOrCopyInfos` and most of the changed you made in this commits ? Aren't uuids enough ?

Mathieu-Deharbe (commented):
**Edit :** Ok I get it : those data are used when drag and dropping but not when cutting and pasting.

I feel like I am missing something here. When I am cut/pasting (= moving) from a node to another and **the source ref modification was inside a composite** I have this request content : 

Image

No container data at all even if it is precisely where this data would be useful. What am I missing ? What is the point of this structure ?

Mathieu-Deharbe (changes requested):
OK but there is a log that should be removed.

Mathieu-Deharbe (approved):

PR #4150, opened 2026-08-13
Moving VoltageLevelSectionCreationForm from gridstudy to commons-ui

benrejebmoh:


src/components/network/constants.ts
```diff
@@ -99,11 +99,6 @@ export const SWITCH_TYPE = {
     DISCONNECTOR: { id: 'DISCONNECTOR', label: 'Disconnector' },
 } as const;
 
-export const POSITION_NEW_SECTION_SIDE = {
```
KoloMenek, line 101: In a recent PR from Anass, he created a constants.ts file in commons-ui. You moved it to utils. I think we should decide on either one of the two methods and be consistent. What do you think ?

KoloMenek (changes requested):
Works locally, few minor comments

src/components/network/constants.ts
benrejebmoh: i m not sure it is necessary here, it could be useful if you have for example a constant file shared between creation and modification forms, but it is not the case here

KoloMenek (approved):

PR #4154, opened 2026-08-17
Remove the any casts around the voltage level creation dialog

flomillot:
Depends on gridsuite/commons-ui#1299.

`VoltageLevelCreationInfo` mixed the modification-server DTO with the call context of `createVoltageLevel` (`studyUuid`, `nodeUuid`, `isUpdate`, `modificationUuid`). `LineSplitWithVoltageLevelDialog` and `LineAttachToVoltageLevelDialog` keep the pending voltage level in their own form and hand it back as `editData`, which expects the DTO alone — hence the `as any` casts.

It is now `VoltageLevelCreationDto` plus that context, the same split `modifyVoltageLevel` already uses, and it no longer reaches the components: `VoltageLevelCreationDialog` used to default its `onCreateVoltageLevel` prop to the `createVoltageLevel` service, which is what forced the context into the callback. The prop now takes the DTO alone and the dialog adds the context itself when no handler is given — the only case that needs it — so the embedding dialogs receive exactly the voltage level they keep in their own state. Both form mappers now declare `VoltageLevelCreationFormData` as their return type. Duplicates of commons-ui types are dropped: `VoltageLeveInfo`, `ExtendedVoltageLevelCreationInfo`, `ExtendedVoltageLevelFormInfos`, `AttachedSubstationCreationInfo`, `CouplingDeviceInfos`, the local `SwitchKind` enum and `voltage-level.type.ts`.

## Notes

Two behaviour changes: an attachment point submitted without opening its dialog now sends `busbarCount` and `sectionCount` at 1 instead of 0, and `hasSubstationCreation` tests the substation creation value rather than the presence of the key, which the full DTO now always has.


src/components/dialogs/network-modifications/line-attach-to-voltage-level/line-attach-to-voltage-level-dialog.tsx
```diff
@@ -185,11 +193,11 @@ const LineAttachToVoltageLevelDialog = ({
                 setNewVoltageLevel(newVoltageLevelInfos);
                 const formattedVoltageLevel = {
                     id: newVoltageLevelInfos.equipmentId,
-                    name: newVoltageLevelInfos.equipmentName,
+                    name: newVoltageLevelInfos.equipmentName ?? undefined,
                     exist: false,
                     busbarCount: newVoltageLevelInfos.busbarCount,
                     sectionCount: newVoltageLevelInfos.sectionCount,
-                    switchKinds: newVoltageLevelInfos.switchKinds ?? [],
+                    switchKinds: newVoltageLevelInfos.switchKinds,
```
ghazwarhili: switchKinds can be undefined when reading backend data 
revert switchKinds ?? []

src/components/dialogs/network-modifications/line-attach-to-voltage-level/line-attach-to-voltage-level-dialog.tsx
```diff
@@ -185,11 +193,11 @@ const LineAttachToVoltageLevelDialog = ({
                 setNewVoltageLevel(newVoltageLevelInfos);
                 const formattedVoltageLevel = {
                     id: newVoltageLevelInfos.equipmentId,
-                    name: newVoltageLevelInfos.equipmentName,
+                    name: newVoltageLevelInfos.equipmentName ?? undefined,
```
ghazwarhili: This change could breaks the functionality
Please revert it

src/components/dialogs/network-modifications/voltage-level/creation/voltage-level-creation.utils.ts
```diff
@@ -0,0 +1,20 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import { VoltageLevelCreationDto } from '@gridsuite/commons-ui';
+import { VoltageLevelCreationInfo } from '../../../../../services/network-modification-types';
+
+/**
+ * Drops the call context, keeping only the modification-server DTO.
+ */
+export const toVoltageLevelCreationDto = ({
```
ghazwarhili: duplicates existing DTO logic to be removed

src/components/dialogs/network-modifications/line-attach-to-voltage-level/line-attach-to-voltage-level-dialog.tsx
```diff
@@ -367,55 +340,31 @@ const LineAttachToVoltageLevelDialog = ({
...
+                        [VOLTAGE_LEVEL]: { [ID]: preparedVoltageLevel.equipmentId },
+                        [BUS_OR_BUSBAR_SECTION]: null,
                     },
                     {
-                        shouldValidate: true,
                         shouldDirty: true,
                     }
                 );
-                setValue(`${CONNECTIVITY}.${BUS_OR_BUSBAR_SECTION}` as any, null);
+                trigger(`${CONNECTIVITY}.${VOLTAGE_LEVEL}`);
```
ghazwarhili, line 356: The change from setValue(..., { shouldValidate: true }) to setValue(..., { shouldDirty: true }) followed by trigger() ... why ?
for me this change is not related to typing !

src/components/dialogs/network-modifications/line-split-with-voltage-level/line-split-with-voltage-level-dialog.tsx
```diff
@@ -188,9 +188,9 @@ const LineSplitWithVoltageLevelDialog = ({
                     id: newVoltageLevel.equipmentId,
                     name: newVoltageLevel.equipmentName ?? '',
                     exist: false,
-                    busbarCount: newVoltageLevel.busbarCount!,
-                    sectionCount: newVoltageLevel.sectionCount!,
-                    switchKinds: newVoltageLevel.switchKinds ?? [],
+                    busbarCount: newVoltageLevel.busbarCount,
+                    sectionCount: newVoltageLevel.sectionCount,
+                    switchKinds: newVoltageLevel.switchKinds,
```
ghazwarhili: the same here

src/components/dialogs/network-modifications/line-split-with-voltage-level/line-split-with-voltage-level-dialog.tsx
```diff
@@ -250,52 +250,19 @@ const LineSplitWithVoltageLevelDialog = ({
...
                 const formattedVoltageLevel = {
                     id: preparedVoltageLevel.equipmentId,
                     name: preparedVoltageLevel.equipmentName ?? '',
                     exist: false,
-                    busbarCount: busbarCount!,
-                    sectionCount: sectionCount!,
-                    switchKinds: switchKinds ?? [],
+                    busbarCount: preparedVoltageLevel.busbarCount,
+                    sectionCount: preparedVoltageLevel.sectionCount,
+                    switchKinds: preparedVoltageLevel.switchKinds,
```
ghazwarhili: the same here

ghazwarhili (changes requested):

ghazwarhili:
busbarCount and sectionCount at 1 instead of 0
PO agrees with this change ?

src/components/dialogs/network-modifications/line-attach-to-voltage-level/line-attach-to-voltage-level-dialog.tsx
flomillot: I don't think it can. 
https://github.com/gridsuite/commons-ui/blob/dbdac341ab0ef7eee97d2c4c1b2464b1eddde0d2/src/features/network-modifications/common/connectivity/VoltageLevelConnectivityForm.tsx#L57-L56
null or undefined give the same empty string at the end
I can uniformise with `?? ''` if you want.

src/components/dialogs/network-modifications/line-attach-to-voltage-level/line-attach-to-voltage-level-dialog.tsx
flomillot: Can you tell me in which case it can be undefined ?

flomillot: There is however a problem with the DTO. 
Because it's used after removeNullFields.
But it's another pre existing problem.

src/components/dialogs/network-modifications/line-attach-to-voltage-level/line-attach-to-voltage-level-dialog.tsx
flomillot: Because of TypeScript, it avoid casting, and is actually more correct (BBS was silently cleaned) than before, except if you have a case in mind ?

src/components/dialogs/network-modifications/voltage-level/creation/voltage-level-creation.utils.ts
flomillot: Duplicates ?

flomillot: I did some cleaning on pre-existing code to avoid this file

flomillot:
> busbarCount and sectionCount at 1 instead of 0 PO agrees with this change ?

It is actually a fix (avoid NPE) but I mentionned it in the US

src/components/dialogs/network-modifications/line-attach-to-voltage-level/line-attach-to-voltage-level-dialog.tsx
ghazwarhili: VoltageLevelFormInfos {
    ....
    @JsonInclude(JsonInclude.Include.NON_NULL)
    private List<SwitchKind> switchKinds;
    ...
    }

src/components/dialogs/network-modifications/line-attach-to-voltage-level/line-attach-to-voltage-level-dialog.tsx
ghazwarhili: newVoltageLevelInfos.equipmentName ?? ' ' it is better than undefined

src/components/dialogs/network-modifications/line-attach-to-voltage-level/line-attach-to-voltage-level-dialog.tsx
flomillot: This DTO is not used here. 
Check VoltageLevelCreationInfos, it's actually not possible.

flomillot: But okay, whatever I can put it back, it's nothing.

src/components/dialogs/network-modifications/line-attach-to-voltage-level/line-attach-to-voltage-level-dialog.tsx
flomillot: Yes that's what I already did.

ghazwarhili (approved):

PR #4173, opened 2026-08-27
Autosave network area diagrams

ayolab:


src/components/workspace/diagrams/nad/use-nad-diagram.ts
```diff
@@ -189,181 +198,133 @@ export const useNadDiagram = ({ panelId, studyUuid, currentNodeId, currentRootNe
...
+        (voltageLevelId: string, position: Partial<DiagramConfigPosition>) => {
+            updateDiagram({
+                positions: diagramRef.current.positions.map((p) =>
+                    p.voltageLevelId === voltageLevelId ? { ...p, ...position } : p
+                ),
             });
+            debounceSaveNad();
+        },
+        [updateDiagram, debounceSaveNad]
+    );
```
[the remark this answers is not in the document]
ayolab: The main way to hit the 700 ms window is dragging a node then closing the panel which deletes the panel's NAD config anyway, so nothing is lost. 
The only real case is a drag immediately followed by a workspace switch, costing one node position. It's also not a regression: previously the config was only saved on leaving edit mode, so closing mid-edit lost more.

Doing it safely needs a "save queued" flag kept in step with the debounce across three call sites; an unconditional saveNad() on unmount would POST for every panel on every workspace switch and race panel deletion. Not worth it for 700 ms. so it is better to leave it as it is

src/components/grid-layout/cards/diagrams/networkAreaDiagram/network-area-diagram-content.tsx
```diff
@@ -163,15 +164,9 @@ const NetworkAreaDiagramContent = memo(function NetworkAreaDiagramContent(props:
...
-    // save nad when exiting edit mode
-    const handleSetIsEditNadMode = useCallback(
-        (newMode: boolean) => {
-            if (isEditNadMode && !newMode) {
-                onSaveNad?.();
-            }
-            setIsEditNadMode(newMode);
-        },
-        [isEditNadMode, onSaveNad]
+    const handleToggleEditNadMode = useCallback(
```
sBouzols, line 167: then it's not toggle if you set it with a param value. I will do another PR for this minor change request.

src/components/workspace/diagrams/nad/use-nad-diagram.ts
```diff
@@ -189,181 +198,133 @@ export const useNadDiagram = ({ panelId, studyUuid, currentNodeId, currentRootNe
...
-                id: diagram.currentNadConfigUuid || null,
-                scalingFactor,
-                voltageLevelIds: diagram.voltageLevelIds,
-                positions: diagram.positions,
+    const movePosition = useCallback(
+        (voltageLevelId: string, position: Partial<DiagramConfigPosition>) => {
+            updateDiagram({
+                positions: diagramRef.current.positions.map((p) =>
+                    p.voltageLevelId === voltageLevelId ? { ...p, ...position } : p
+                ),
```
sBouzols, line 250: we make a loop over all the positions in the diagram, just to change one ? and a copy of all the positions ? Does it has effect over performance ?

sBouzols: it was already the case before ok ...

sBouzols (approved):
Code review OK but there are some refactoring and feature dev, it should be in different PRs in the future.
Tests OK (as much complete as possible)
Console warning check ok

PR #4180, opened 2026-09-02
Spreadsheet optimizations

Meklo:
## PR Summary
Combines two optimizations which dramatically improves spreadsheet performances at scale :
- Return null instead of undefined for empty values. it enables AG grid to store the result in its value cache
- Add a cache which stores the result of each formula compilation which can then be reevaluated when rerenders without parsing and compiling in each pass. 

Especially visible when spreadsheet contains complex formulas on a large dataset


src/components/spreadsheet-view/spreadsheet/spreadsheet-content/equipment-table.tsx
```diff
@@ -134,7 +135,13 @@ export const EquipmentTable: FunctionComponent<EquipmentTableProps> = ({
         [currentNode?.type, theme, isDataEditable]
     );
 
-    const gridContext = useMemo(() => ({ theme, currentNode, studyUuid }), [currentNode, studyUuid, theme]);
+    // The Map lives in a ref, NOT in the memo below: gridContext is rebuilt on every currentNode
+    // change, and creating the cache there would silently flush all compiled formulas per rebuild.
+    const compiledFormulaCache = useRef(createCompiledFormulaCache()).current;
```
flomillot: I prefer when a ref ends with `...Ref` and when we use `.current` to access it, easier to see its a ref, and it's what we do everywhere I think

src/components/spreadsheet-view/columns/utils/math.ts
```diff
@@ -24,16 +24,38 @@ function transformExpression(expr: string): string {
     return expr.replaceAll(regex, `$1.steps[string($2)]`);
 }
 
-const originalEvaluate = instance.evaluate;
+const originalParse = instance.parse;
```
flomillot: Why not using `instance.compile` instead ?
It does both parsing and compiling in the same time.
https://mathjs.org/docs/expressions/parsing.html#compile

src/components/spreadsheet-view/columns/utils/column-mapper.ts
```diff
@@ -24,24 +24,26 @@ export const SPREADSHEET_INVALID_CELL_CLASS = 'spreadsheet-invalid-cell';
 
 const createValueGetter =
     (colDef: ColumnDefinition) =>
-    (params: ValueGetterParams): CustomAggridValue | undefined => {
+    (params: ValueGetterParams): CustomAggridValue | null => {
         try {
             // Skip formula processing for pinned rows and use raw value
             if (isCalculationRow(params.node?.data?.rowType)) {
-                return params.data[colDef.id];
+                return params.data[colDef.id] ?? null;
```
flomillot, line 31: `0` is falsy so return `null`, is it what we want here ?

src/components/spreadsheet-view/columns/utils/column-mapper.ts
```diff
@@ -24,24 +24,26 @@ export const SPREADSHEET_INVALID_CELL_CLASS = 'spreadsheet-invalid-cell';
...
             }
             const scope = { ...params.data };
             const colDependencies = colDef.dependencies ?? [];
+
+            //Empty values are assumed to be equal to "undefined" by users
             colDependencies.forEach((dep) => {
-                scope[dep] = params.getValue(dep);
+                scope[dep] = params.getValue(dep) ?? undefined;
```
flomillot, line 41: Same

flomillot: And why `undefined` here btw ?

src/components/spreadsheet-view/spreadsheet/spreadsheet-content/equipment-table.tsx
flomillot: BTW, `useMemo(createCompiledFormulaCache, [])` would be better in term of performance.
Here you initialize a new Map which it throw away at every render (the argument is evaluated before calling `useRef`)
`useRef` should always be used in last option IMO.

flomillot:
To keep ISO behaviour, I think you need to change this also
https://github.com/gridsuite/commons-ui/blob/dbdac341ab0ef7eee97d2c4c1b2464b1eddde0d2/src/components/composite/customAGGrid/cell-renderers.tsx#L50

src/components/spreadsheet-view/columns/utils/column-mapper.ts
Meklo: This undefined is crucial to keep existing formulas having this type of syntax "typeOf(IT10_Or) == 'undefined' ? IT10_Ex :  typeOf(IT10_Ex) == 'undefined' ? IT10_Or : max(IT10_Or, IT10_Ex)" which are largely used in production.
As of today empty values were treated as undefined. In order to keep the system working while treating empty value as null, we now do the inverse and inject undefined in the scope when the value is empty

~~But if it's equal to 0 it should stay 0 you're right~~
Edit: same as above for the nullish coalescing operator

src/components/spreadsheet-view/columns/utils/column-mapper.ts
Meklo: ~~I don't think so, I'll fix this~~
Edit: in fact the nullish coalescing operator is only applied if the value is equal to undefined or null, not falsy so it should work as expected

src/components/spreadsheet-view/columns/utils/math.ts
Meklo: True that, we can directly invoke compile

src/components/spreadsheet-view/spreadsheet/spreadsheet-content/equipment-table.tsx
Meklo: Fine by me

src/components/spreadsheet-view/columns/utils/column-mapper.ts
```diff
@@ -24,24 +24,26 @@ export const SPREADSHEET_INVALID_CELL_CLASS = 'spreadsheet-invalid-cell';
...
             });
-            const result = limitedEvaluate(colDef.formula, scope);
-            return result == null ? undefined : validateFormulaResult(result, colDef.type);
+            const result = limitedEvaluate(colDef.formula, scope, params.context?.compiledFormulaCache);
```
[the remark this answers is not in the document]
Meklo: I'd rather update the renderers check

src/components/spreadsheet-view/columns/utils/column-mapper.ts
flomillot: Ah ok my bad I fogot that

src/components/spreadsheet-view/columns/utils/column-mapper.ts
flomillot: Okay thanks, maybe update the comment to make it more explicit ?

flomillot (approved):
Didnt test but I checked the videos of the result. 👍 
Just need the commons-ui changes also.

src/components/spreadsheet-view/spreadsheet/spreadsheet-content/equipment-table.tsx
```diff
@@ -134,7 +135,13 @@ export const EquipmentTable: FunctionComponent<EquipmentTableProps> = ({
         [currentNode?.type, theme, isDataEditable]
     );
 
-    const gridContext = useMemo(() => ({ theme, currentNode, studyUuid }), [currentNode, studyUuid, theme]);
+    // The Map lives in a distinct memo, NOT in the memo below: gridContext is rebuilt on every currentNode
+    // change, and creating the cache there would silently flush all compiled formulas per rebuild.
+    const compiledFormulaCache = useMemo(() => createCompiledFormulaCache(), []);
```
dbraquart, line 140:
```suggestion
    const compiledFormulaCacheRef = useRef<CompiledFormulaCache>(undefined);
    if (!compiledFormulaCacheRef.current) {
        compiledFormulaCacheRef.current = createCompiledFormulaCache();
    }
    const compiledFormulaCache = compiledFormulaCacheRef.current;
```

using useRef to guarantee a unique creation of the cache for the life of the component ?

useMemo() should do the same in practice, but this is not guaranteed by React. 

This may be worth it because the cache is the whole point of your PR..

dbraquart (approved):
code : just a suggestion

tests (on quads tab, 2 aliases, tons of formulas): **OK**
- column sort is very much faster (no cpu peak)
- column filtering is very much faster (still some cpu little peak)
- adding computed lines very much faster too
- csv copy/export too
- spreadsheet more stable (no frozzing periods)
- same cells content
- error still display empty cell

dbraquart:
ahah you started with Ref, reviewer1 requested a Memo, reviewer2 suggests a Ref (I didnt check the first review before).

well, this was just a (theoretical) suggestion from me.

PR #4181, opened 2026-09-02
save description for created new composite

souissimai:


src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
```diff
@@ -883,31 +887,52 @@ const NetworkModificationNodeEditor = () => {
...
+                                  };
+                              })
+                        : Promise.resolve<CompositeModificationContent>({
+                              modificationUuid: item.uuid,
+                              description: item.description || undefined,
+                          })
+                )
+            ),
+        [selectedNetworkModifications]
+    );
```
Meklo: Looks weird to me that the reference resolution is handled in the frontend, shouldn't it be done in the backend ?

src/services/explore.ts
```diff
@@ -63,11 +63,16 @@ export function elementExists(directoryUuid: UUID, elementName: string, type: El
     });
 }
 
+export interface CompositeModificationContent {
+    modificationUuid: UUID;
+    description?: string;
+}
```
Meklo: Isn't it redundant ? We already pass the description as a parameter in the create and update functions

Meklo (approved):

PR #4187, opened 2026-09-08
Add warning when copying modification reference

Meklo:


src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
```diff
@@ -213,6 +217,35 @@ const NetworkModificationNodeEditor = () => {
         });
     }, [cleanOtherTabsClipboard]);
 
+    useEffect(() => {
```
souissimai: const openCreateCompositeModificationDialog = useCallback(async () => {
        // nested references are lazily loaded by the table, so the selection alone can't tell :
        // ask the backend for each selected composite before opening the save dialog
        setSelectionHasSharedContent(false);
        if (selectionContainsShared) {
            setSelectionHasSharedContent(true);
        } else {
            const compositeUuids = selectedNetworkModifications
                .filter((m) => m.type === ModificationType.COMPOSITE_MODIFICATION)
                .map((m) => m.uuid);
            if (compositeUuids.length > 0) {
                try {
                    setSelectionHasSharedContent(await hasModificationReferences(compositeUuids));
                } catch (error) {
                    snackWithFallback(snackError, error);
                }
            }
        }
        setCreateCompositeModificationDialogOpen(true);
    }, [selectionContainsShared, selectedNetworkModifications, snackError]);
 why useeffect and not when opening the composite modification dialog

souissimai (changes requested):

src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
Meklo: No valid reason to dedicate it a useEffect you're right

souissimai (approved):

PR #4191, opened 2026-09-09
add filter for voltage on vl in nad

EtienneLt:


src/components/workspace/diagrams/nad/nad-info-filter.tsx
```diff
@@ -17,7 +17,8 @@ const INFO_ROWS: { key: NadSelectedInfoKey; labelId: string }[] = [
     { key: 'permanentLimitPercentage', labelId: 'nadInfoPermanentLimitPercentage' },
     { key: 'activePowerArrow', labelId: 'nadInfoActivePowerArrow' },
     { key: 'reactivePowerArrow', labelId: 'nadInfoReactivePowerArrow' },
-    { key: 'voltageLevelName', labelId: 'nadInfoVoltageLevelName' },
+    { key: 'voltageLevelInfo', labelId: 'nadInfoVoltageLevelInfo' },
```
Mathieu-Deharbe: Why did you change from `voltageLevelName` to `voltageLevelInfo` ? It looks like name was fine and more precise.

src/components/workspace/diagrams/nad/use-nad-info-filter.ts
```diff
@@ -14,7 +14,8 @@ export type NadSelectedInfoKey =
     | 'permanentLimitPercentage'
     | 'activePowerArrow'
     | 'reactivePowerArrow'
-    | 'voltageLevelName';
+    | 'voltageLevelInfo'
```
Mathieu-Deharbe: Same : why not voltageLevelName ?

Mathieu-Deharbe (changes requested):
I didn't manage to join Michaël to ask him but I don't think that you did what he wanted : here "voltage level info" hide both the name and the voltage. I think there should be a "Voltage level name" which only hides the name and the "voltage level voltages" that you already added.

I also think that "info" doesn't sound very professional.

Nevermind if Michael already validated this.

Mathieu-Deharbe (approved):

PR #4157, opened 2026-08-19
Allow to share a composite modification from a study

flomillot:
In the dialog saving modifications into GridExplore, a third option "New (shared)" stores the selected composite modification as a shared one: the composite itself leaves the study and is replaced in the node by a reference to it.

The option is only available when a single composite modification is selected — it is greyed out for any other selection, an already shared modification not being shareable again.

The labels of the dialog are shortened accordingly: "Create a composite modification", with "New", "New (shared)" and "Replace an existing one".

Requires gridsuite/commons-ui#1302 and gridsuite/study-server#1070.

PR #4192, opened 2026-09-10
Share a composite modification nested in another composite

flomillot:
Follow-up of #4157.

The "New (shared)" option of the dialog saving modifications into GridExplore is now only enabled in developer mode — it stays visible but greyed out otherwise.

A composite modification nested in another composite of the node can now be shared, as long as it is not contained by an already shared composite. The option is also greyed out when the selected composite contains a shared modification, at any depth: since the table only loads the content of a composite as it is unfolded, the server is asked, once the dialog is open.

The saving logic moves out of the node editor into a dedicated `SaveNetworkModificationsDialog` wrapping `ElementSaveDialog`. A single server call there both warns that shared modifications are saved as a copy and greys out the sharing.

Requires gridsuite/study-server#1089 and gridsuite/commons-ui#1336.

### DIFF
--- a/src/components/graph/menus/network-modifications/save-network-modifications-dialog.tsx
+++ b/src/components/graph/menus/network-modifications/save-network-modifications-dialog.tsx
@@ -1,0 +1,104 @@
+/**
+ * Copyright (c) 2026, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import {
+    ComposedModificationMetadata,
+    ElementSaveDialog,
+    ElementType,
+    type IElementCreationDialog,
+    type IElementUpdateDialog,
+    ModificationType,
+    PARAM_DEVELOPER_MODE,
+} from '@gridsuite/commons-ui';
+import type { UUID } from 'node:crypto';
+import { useEffect, useState } from 'react';
+import { useParameterState } from 'components/dialogs/parameters/use-parameters-state';
+import { containsSharedModification } from '../../../../services/study/network-modifications';
+
+export interface SaveNetworkModificationsDialogProps {
+    open: boolean;
+    onClose: () => void;
+    studyUuid: UUID;
+    selectedModifications: ComposedModificationMetadata[];
+    defaultName: string | null;
+    onSave: (data: IElementCreationDialog) => void;
+    onSaveShared: (data: IElementCreationDialog) => void;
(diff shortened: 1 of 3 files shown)


src/components/graph/menus/network-modifications/save-network-modifications-dialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+        }
+        let active = true; // to manage race condition
+        containsSharedModification(selectedComposite.uuid)
+            .then((containsShared) => {
+                if (active) {
+                    setSelectedCompositeContainsShared(containsShared);
+                }
+            })
+            .catch((error) => {
+                console.error(
```
ghazwarhili, line 72: snackWithFallback(snackError, error, { headerId: 'errContainsSharedModificationMsg' });

src/services/study/network-modifications.ts
```diff
@@ -1591,6 +1591,14 @@ export function shareCompositeModification(
     return backendFetch(url, { method: 'POST' });
 }
 
+export function containsSharedModification(compositeModificationUuid: UUID): Promise<boolean> {
+    const url = `${PREFIX_STUDY_QUERIES}/v1/network-composite-modifications/${safeEncodeURIComponent(
+        compositeModificationUuid
+    )}/contains-shared-modification`;
+    console.debug(url);
```
ghazwarhili, line 1598: console.debug(url); to be cleaned up maybe

src/components/graph/menus/network-modifications/save-network-modifications-dialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const isSharingAvailable =
+        isDeveloperMode && isSelectedCompositeShareable && selectedCompositeContainsShared === false;
+
+    return (
+        <ElementSaveDialog
+            open={open}
+            onSave={onSave}
+            onSaveShared={onSaveShared}
+            createSharedDisabled={!isSharingAvailable}
+            OnUpdate={onUpdate}
```
ghazwarhili, line 91: It would be better to use onUpdate instead of OnUpdate

ghazwarhili (approved):

src/components/graph/menus/network-modifications/save-network-modifications-dialog.tsx
flomillot: Done

src/services/study/network-modifications.ts
flomillot: Its just how its done everywhere

src/components/graph/menus/network-modifications/save-network-modifications-dialog.tsx
flomillot: Yup but not my code

PR #4195, opened 2026-09-11
add description when creating reference modification

EtienneLt:


src/components/graph/menus/network-modifications/network-modification-node-editor.tsx
Mathieu-Deharbe:
```suggestion
    // If only one modification is selected and it is of type composite, saving it in gridexplore would make it take its name and description by default
```

Mathieu-Deharbe (approved):

PR #4209, opened 2026-09-18
Warn about shared modifications when renaming a root network tag

flomillot:
When the tag of a root network is being changed in a study containing shared modifications, a warning next to the field tells that the applicabilities these modifications may already have for the new tag will be overwritten.

Goes with gridsuite/network-modification-server#901 and gridsuite/study-server#1104.


src/components/dialogs/root-network/root-network-dialog.tsx
```diff
@@ -109,12 +113,25 @@ const RootNetworkDialog: React.FC<RootNetworkDialogProps> = ({
...
         reset,
         setValue,
         setError,
         formState: { errors },
     } = formMethods;
 
+    const [isStudyContainingSharedModifications, setIsStudyContainingSharedModifications] = useState(false);
+    useEffect(() => {
+        if (open && isModification && studyUuid) {
+            hasSharedModifications(studyUuid)
```
ghazwarhili, line 126: fetchHasSharedModifications(studyUuid)

src/components/dialogs/root-network/root-network-dialog.tsx
```diff
@@ -109,12 +113,25 @@ const RootNetworkDialog: React.FC<RootNetworkDialogProps> = ({
...
 
     const {
+        control,
         reset,
         setValue,
         setError,
         formState: { errors },
     } = formMethods;
 
+    const [isStudyContainingSharedModifications, setIsStudyContainingSharedModifications] = useState(false);
```
ghazwarhili, line 123: [hasSharedModifications, setHasSharedModifications ] = useState(flase)

src/components/dialogs/root-network/root-network-dialog.tsx
```diff
@@ -109,12 +113,25 @@ const RootNetworkDialog: React.FC<RootNetworkDialogProps> = ({
...
+    useEffect(() => {
+        if (open && isModification && studyUuid) {
+            hasSharedModifications(studyUuid)
+                .then(setIsStudyContainingSharedModifications)
+                .catch(() => setIsStudyContainingSharedModifications(false));
+        }
+    }, [isModification, open, studyUuid]);
+
+    const tag = useWatch({ control, name: TAG })?.trim();
```
ghazwarhili, line 132: rename to watchedTag or the currentTag

src/components/dialogs/root-network/root-network-dialog.tsx
flomillot: Why ? We don't want to fetch anything, we just want a boolean.

flomillot: If I really want to follow what exists, I can add check at the beginning. But I don't think it's necessary.

src/components/dialogs/root-network/root-network-dialog.tsx
flomillot: We are inside a component which is called RootNetworkDialog. A root network doesn't hold modifications itself. I prefer a longer but clearer name.

src/components/dialogs/root-network/root-network-dialog.tsx
flomillot: It's not the convention we use actually.

src/components/dialogs/root-network/root-network-dialog.tsx
```diff
@@ -109,12 +110,28 @@ const RootNetworkDialog: React.FC<RootNetworkDialogProps> = ({
...
     } = formMethods;
 
+    const [isStudyContainingSharedModifications, setIsStudyContainingSharedModifications] = useState(false);
+    useEffect(() => {
+        if (open && isModification && studyUuid) {
+            hasSharedModifications(studyUuid)
+                .then(setIsStudyContainingSharedModifications)
+                .catch((error) => {
+                    console.error('Failed to fetch whether the study contains shared modifications', error);
+                    setIsStudyContainingSharedModifications(false);
```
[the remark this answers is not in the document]
flomillot: It's not a quick win, The back end already protects the renaming, And it's already the case here : FRONT-END/commons-ui/src/services/directory.ts:68

